PR #44, opened 2014-11-26
Cover Art Archive support for summary.html

Freso:
Some talk on IRC for more background: http://chatlogs.musicbrainz.org/musicbrainz-devel/2014/2014-11/2014-11-26.html#T10-15-07-807376

Note that the complicated CAA querying will be useful for making a thickbox (or whatever) CAA gallery, per my comment on IRC.

Also, it would be good if nginx or something could be set up to cache CAA responses, so we won't have to send off a query whenever someone reloads a page or loads another recording MBID from the same release they were just at. :)

Screenshot!

### DIFF
--- a/acousticbrainz/server.py
+++ b/acousticbrainz/server.py
@@ -8,6 +8,7 @@ import uuid
 import datetime
 import time
 import os
+import urllib2
 from operator import itemgetter
 from logging.handlers import RotatingFileHandler
 from flask import Flask, request, Response, jsonify, render_template, redirect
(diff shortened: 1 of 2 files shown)


tsukanov:
Any plans to update this?

mayhem:
I think Freso said he wasn't planning updating it. I'll leave it to you to decide if:

1. Reject the PR and close it.
2. Re-work the PR to load the cover art as CB does.

mayhem:
Generally, I suppose showing CAA coverart and would like to have this!

Freso:
Yeah. I'm not good with client side stuff, so if the image should not be "discovered" server-side, I don't really have anything more to add. Granted, now that we're interfacing with the database, it should be possible to look up the CAA CA there, no? So no need for "expensive" HTTP round-trips to caa.org to see if there's CA or not. Anyway, it's all up to you @tsukanov. :)

alastair:
We're not actually connecting directly to the database, it's still a ws
call. I thought the ws doesn't expose this directly...

PR #55, opened 2014-12-12
Data exporting and importing

tsukanov:
*http://tickets.musicbrainz.org/browse/AB-22 and other stuff...*


tsukanov:
Can you please clarify parts about *streaming mode* (1) and compressing tar file on the fly (3)?
Also, I don't think we write anything during exporting.

mayhem:
Streaming mode means that we're compressing the file on the fly as we're generating it. Compare to the approach where files get written to disk and are then tar'red and compressed.

alastair:
From our discussion in IRC:
We should do an import the same way as MusicBrainz - where we first create the table structure, then do the import, and finally create PKs, FKs, and indexes. This is a lot faster than importing all the data with existing indexes, and is what pg_dump and pg_restore do.
http://www.postgresql.org/docs/current/static/populate.html

lalinsky:
Regarding the streaming mode, you can open `pxz` or `xz` with `subprocess.Popen` with `stdin` or `stdout` as a pipe (but only one, the other one should be a real file, otherwise you get into synchronization problems) and use the pipe as `tarfile` input.

manage.py
```diff
@@ -0,0 +1,197 @@
...
+    create_path(temp_dir)
+
+    # Preparing meta files
+    with open('%s/TIMESTAMP' % temp_dir, 'w') as f:
+        f.write(time_now.isoformat(' '))
+    with open('%s/SCHEMA_SEQUENCE' % temp_dir, 'w') as f:
+        f.write(str(acousticbrainz.__version__))
+
+    # Creating the archive
+    with tarfile.open("%s/abdump.tar" % dump_dir, "w") as tar:
```
alastair: I think the output file should have the date in it

manage.py
```diff
@@ -0,0 +1,178 @@
...
+    print("Done!")
+
+
+def _run_sql_script(sql_file):
+    cursor = db_connection.cursor()
+    with open(sql_file) as sql:
+        cursor.execute(sql.read())
+
+
+def _import_data(archive):
```
tsukanov: Ah, interesting. Seems to be working now. Thanks.

admin/utils.py
```diff
@@ -0,0 +1,51 @@
...
+import sys
+
+
+def create_path(path):
+    """Creates a directory structure if it doesn't exist yet."""
+    try:
+        os.makedirs(path)
+    except OSError as exception:
+        if exception.errno != errno.EEXIST:
+            sys.exit("Failed to create directory structure %s. Error: %s" % (path, exception))
```
alastair: you could just `raise` here instead wrapping it in exit

admin/sql/create_tables.sql
```diff
@@ -0,0 +1,33 @@
...
+  id          SERIAL,
+  mbid        UUID NOT NULL,
+  build_sha1  TEXT NOT NULL,
+  lossless    BOOLEAN                  DEFAULT 'n',
+  data        JSON NOT NULL,
+  submitted   TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  data_sha256 TEXT NOT NULL
+);
+
+CREATE TABLE highlevel (
```
alastair: I'm confused - can someone explain to me how create_tables should work in conjunction with the other update sql scripts? Should create_tables always make the most up-to-date version, and people with older versions run the update script to get up-to-date?

mayhem: The update tables only get used for schema changes.

manage.py
```diff
@@ -0,0 +1,213 @@
...
+        export_highlevel_json()
+
+    if not no_lowlevel:
+        export_lowlevel_json()
+
+
+@manager.command
+def export_lowlevel_json(location=os.path.join(os.getcwd(), 'export'), rotate=False):
+    print("Dumping lowlevel data (JSON)...")
+    print("Done! Created:", dump_lowlevel_json(location))
```
alastair: Please don't put side effects in a print message!

manage.py
```diff
@@ -0,0 +1,213 @@
...
+    pxz_command = ['pxz', '--decompress', '--stdout', archive]
+    pxz = subprocess.Popen(pxz_command, stdout=subprocess.PIPE)
+
+    table_names = _tables.keys()
+    cursor = db_connection.cursor()
+
+    with tarfile.open(fileobj=pxz.stdout, mode='r|') as tar:
+        for member in tar:
+
+            if member.name == 'SCHEMA_SEQUENCE':
```
alastair: If schema_sequence comes after the tables in the tar file you're going to exit /after/ you do the import. Can you confirm that you've made the tar file in order? (just telling me yes/no is enough)

manage.py
```diff
@@ -0,0 +1,213 @@
...
+            pxz_command.append('-T %s' % threads)
+        pxz = subprocess.Popen(pxz_command, stdin=subprocess.PIPE, stdout=archive)
+
+        # Creating the archive
+        with tarfile.open(fileobj=pxz.stdin, mode='w|') as tar:
+            # TODO(roman): Get rid of temporary directories and write directly to tar file
+            temp_dir = tempfile.mkdtemp()
+
+            # Adding metadata
+            with open('%s/SCHEMA_SEQUENCE' % temp_dir, 'w') as f:
```
alastair: os.path.join

manage.py
tsukanov: *SCHEMA_SEQUENCE* is added to tar file first during creation, so it should be a first member there. We can check if it is.

manage.py
tsukanov: Should I do this everywhere paths are constructed? For `archive_path` variable above and stuff below, for example.

tsukanov:
I also split `acousticbrainz.data` module into multiple ones (as a continuation of improvements made in #51) and moved dump creation stuff there. Not sure if that's the right way to do that, but it seems like a good idea to keep related functionality in one place. Let me know what you think.

mayhem:
I just checked this out and played with the database dumping / importing and it worked great. Keep it up!

tsukanov:
Don't know if we should combine commands in *manage.py* or keep them separate (like `export_lowlevel_json` and `export_lowlevel_json_incremental`). Also not sure about their names.

manage.py
```diff
@@ -0,0 +1,135 @@
...
+
+@manager.command
+def export_lowlevel_json(location=os.path.join(os.getcwd(), 'export'), rotate=False):
+    print("Creating low level JSON data dump...")
+    path = dump_lowlevel_json(location)
+    print("Done! Created:", path)
+
+    if rotate:
+        print("Removing old dumps (except two latest)...")
+        remove_old_archives(location, "acousticbrainz-lowlevel-json-[0-9]+-json.tar.bz2",
```
alastair: has anyone noticed that this file says json twice? Do the new dumps do this? will this rm work now?

tsukanov: Ah, I updated names in dump creator but not there. Should be fixed now.

tsukanov:
Should be complete now.

mayhem:
When no incremental data is available, it would be nice to just print an error message and not to throw an error: Exception: No new data since the last incremental dump!

Stack traces are scary!

mayhem:
\ø/ Lets merge this! :+1:

PR #64, opened 2015-01-17
Template cleanup

tsukanov:
Cleaned up templates without changing the content.

* Removed unnecessary blocks, class/id definitions, etc.
* Fixed indentation
* Replaced internal links with `url_for()` calls.  
\+ some other small improvements.


mayhem:
On monday I'll review the dumping patch and see if I can get a server update out.

PR #65, opened 2015-01-20
AB-17: Integration with MusicBrainz

tsukanov:


templates/data/summary.html
```diff
@@ -24,25 +31,38 @@
...
-              {% if lowlevel.metadata.tags.musicbrainz_artistid %}
-                <a href="https://musicbrainz.org/artist/{{ lowlevel.metadata.tags.musicbrainz_artistid[0] }}">
-                  {{ lowlevel.metadata.tags.artist[0] }}
-                </a>
-              {% else %}
-                {{ lowlevel.metadata.tags.artist[0] }}
-              {% endif %}
+              {%- if metadata -%}
+                {# TODO(roman): Handle multiple artists properly. #}
+                {%- set artist_id = metadata['artist-credit'][0]['artist']['id'] -%}
```
alastair: I'd prefer to have this object navigation in the view instead of the template (or even better, a separate helper function)

templates/data/summary.html
```diff
@@ -24,25 +31,38 @@
...
+                <a href="https://musicbrainz.org/artist/{{ artist_id }}">{{ metadata['artist-credit-phrase'] }}</a>
+              {%- else -%}
+                {%- if lowlevel.metadata.tags.musicbrainz_artistid -%}
```
alastair: Is it worth verifying if the tag information (if present) is the same or different from musicbrainz information? Maybe it could be an interesting report to see "badly tagged" file? On the other hand, maybe it would get noisy as stuff on MB is edited for normal reasons

acousticbrainz/views/data.py
```diff
@@ -58,6 +60,13 @@ def get_high_level(mbid):
...
 
+    # Getting good metadata from MusicBrainz
+    try:
+        metadata = musicbrainzngs.get_recording_by_id(
+            mbid, includes=['artists', 'releases', 'media'])['recording']
+    except ResponseError:
+        metadata = None
+
     # Tomahawk player stuff
     if not ('artist' in lowlevel['metadata']['tags'] and 'title' in lowlevel['metadata']['tags']):
```
alastair: Should use metadata from musicbrainz for tomahawk

alastair: and on the front page?

templates/data/summary.html
tsukanov: Some kind of indication if there is a difference between submitted info and one that we get from MusicBrainz?

templates/data/summary.html
```diff
@@ -11,7 +11,14 @@
 
 {%- block content -%}
   <h3>Details for track {{ mbid }}</h3>
-  {% if mbid %}
+
+  {%- if not metadata -%}
+    <div class="alert alert-warning" role="alert">
```
alastair: This isn't really a warning. It's not actionable by the person viewing the data.
Maybe instead we can have a "Source" label in the metadata table (where it currently says 'value')

acousticbrainz/views/data.py
tsukanov: On the front page for the "5 last submitted tracks" list?

templates/data/summary.html
Freso: Perhaps another column in the table so both could be presented?

alastair: I don't think doubling information is worthwhile, it just makes it messy.

acousticbrainz/views/data.py
alastair: yes

templates/data/summary.html
tsukanov: Done.

acousticbrainz/views/data.py
tsukanov: That would be 5 requests for content that changes all the time. Do we really want to do it there?

mayhem:
As per our IRC conversation, this should also be merged.

PR #67, opened 2015-01-26
Added high level extractor setup into Vagrant VM

tsukanov:


mayhem:
:+1: Not tested, but looks ok.

PR #70, opened 2015-02-01
Separate data importing command in manage.py

tsukanov:
It makes possible importing incremental dumps after database is initialized with `init_db`.


mayhem:
:+1:

PR #74, opened 2015-02-05
MusicBrainz Login

tsukanov:


acousticbrainz/__init__.py
```diff
@@ -4,7 +4,7 @@
 from logging.handlers import RotatingFileHandler
 
 # This value must be incremented after any schema changes!
-__version__ = 2
+__version__ = 3
```
alastair: only a schema change on replicated tables. My guess is that we're not going to replicate the user table

acousticbrainz/data/user.py
```diff
@@ -0,0 +1,65 @@
+import psycopg2
+from flask import current_app
+from flask_login import UserMixin
+
+
+def create(musicbrainz_id):
+    connection = psycopg2.connect(current_app.config['PG_CONNECT'])
+    cursor = connection.cursor()
+    cursor.execute('INSERT INTO "user" (musicbrainz_id) VALUES (%s)',
```
mayhem: Should you be catching DB exceptions here (and below)?

acousticbrainz/testing.py
```diff
@@ -21,16 +22,17 @@ def tearDown(self):
...
-        cur = conn.cursor()
-        cur.execute("TRUNCATE highlevel_json CASCADE;")
-        cur.execute("TRUNCATE highlevel CASCADE;")
-        cur.execute("TRUNCATE lowlevel CASCADE;")
-        cur.execute("TRUNCATE statistics CASCADE;")
-        cur.execute("TRUNCATE incremental_dumps CASCADE;")
-        conn.commit()
-        cur.close()
-        conn.close()
+        connection = psycopg2.connect(current_app.config['PG_CONNECT'])
```
mayhem: Ugh. I think I would prefer if it connected to PG_CONNECT_TEST, rather than rebinding PG_CONNECT. If for some reason that rebind doesn't happen, a non-test (production!) DB could get nuked. :(

mayhem:
In general this looks ok to me. However, we should think about how to merge this -- just being able to log in isn't very useful, but it would be nice to merge this PR once it's done. Maybe we can just hide the login link for the time being? That way would could release other features to the server after this is merged and not confuse users...

tsukanov:
It's hidden right now.

mayhem:
Perfect. Then I would ask alastair to give this a thumbs up as well and then we can merge. This one is important for both of us to review.

acousticbrainz/testing.py
tsukanov: I don't think these tests are meant to be run on production DB. I can explicitly reference `PG_CONNECT_TEST` there in cleanup code, but in other places it would make rebinding pointless. We'll have to check if tests are being run and switch connection strings assuming that `TESTING` config variable is set properly.

mayhem: That's it -- not meant to. But if it were to happen accidentally (or habitually) then we lose our production DB. :(

acousticbrainz/data/user.py
tsukanov: Yeah, I probably should.

alastair:
(this is still on my list, sorry for the delay)

acousticbrainz/__init__.py
```diff
@@ -34,6 +34,12 @@ def create_app():
     if app.config['MUSICBRAINZ_HOSTNAME']:
         musicbrainzngs.set_hostname(app.config['MUSICBRAINZ_HOSTNAME'])
 
+    # OAuth
+    from acousticbrainz.login import login_manager, provider
+    login_manager.init_app(app)
+    provider.init(app.config['MUSICBRAINZ_CLIENT_ID'],
```
alastair: Do you want to give a warning if this is unset, or should it just fail?

acousticbrainz/data/user.py
```diff
@@ -0,0 +1,81 @@
+import psycopg2
+from flask import current_app
+from flask_login import UserMixin
+from werkzeug.exceptions import BadRequest, ServiceUnavailable
+
+
+def create(musicbrainz_id):
+    try:
+        connection = psycopg2.connect(current_app.config['PG_CONNECT'])
```
alastair: how many places do you use this `connect` command? It feels to me like it should be an application-wide method than you can call. This way if we ever need to change arguments we only have to do it in one place

acousticbrainz/data/user.py
```diff
@@ -0,0 +1,81 @@
...
+        cursor = connection.cursor()
+        cursor.execute('INSERT INTO "user" (musicbrainz_id) VALUES (%s) RETURNING id',
+                       (str(musicbrainz_id),))
```
alastair: can musicbrainz usernames be unicode? what will this do?

acousticbrainz/data/user.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return new_id
+
+
+def get(id):
+    try:
+        connection = psycopg2.connect(current_app.config['PG_CONNECT'])
+        cursor = connection.cursor()
+        cursor.execute('SELECT id, created, musicbrainz_id FROM "user" WHERE id = %s',
+                       (str(id),))
```
alastair: can you clarify if arguments to `execute` should always be string, or if this should be an int since the id column is a number

acousticbrainz/__init__.py
tsukanov: It will fail.

admin/updates/update-3.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+CREATE TABLE "user" (
```
alastair: `user` is a reserved word in postgres. You get around it by using quotes everywhere, but I can imagine this causing someone lots of pain if they don't realise this (I've been there before)

acousticbrainz/data/user.py
tsukanov: Ah, yeah. Definitely can: https://musicbrainz.org/user/Пользователь
Should fix this.

acousticbrainz/data/user_test.py
```diff
@@ -0,0 +1,29 @@
...
+        self.assertIsNone(user_data.get_by_mb_id("fuzzy"))
+
+        musicbrainz_id = "fuzzy_dunlop"
+        user_id = user_data.create(musicbrainz_id)
+        user = user_data.get_by_mb_id(musicbrainz_id)
+        self.assertIsNotNone(user)
+        self.assertEqual(user.id, user_id)
+
+    def test_get_or_create(self):
+        self.assertIsNotNone(user_data.get_or_create("fuzzy_dunlop"))
```
alastair: What exactly are you testing here? I'd like to see an additional test that creates then uses get_or_create and checks they have the same id

acousticbrainz/data/user.py
```diff
@@ -0,0 +1,81 @@
...
+    except psycopg2.IntegrityError, e:
+        raise BadRequest(str(e))
+    except psycopg2.OperationalError, e:
+        raise ServiceUnavailable(str(e))
+
+    if cursor.rowcount > 0:
```
alastair: why might there be two or more rows as a result of this query? Should it silently ignore this fact, or cause an error?

acousticbrainz/login/provider.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def get_user():
+    """Function should fetch user data from database, or, if necessary, create it, and return it."""
+    s = _musicbrainz.get_auth_session(data={
+        'code': _fetch_data('code'),
+        'grant_type': 'authorization_code',
+        'redirect_uri': url_for('login.musicbrainz_post', _external=True)
+    }, decoder=json.loads)
+    data = s.get('oauth2/userinfo').json()
```
alastair: what if musicbrainz doesn't return a `sub` key in the data?

acousticbrainz/data/user.py
tsukanov: Yeah, I think that's something we should do. I'll look into that.

acousticbrainz/login/provider.py
```diff
@@ -0,0 +1,76 @@
...
+    code = request.args.get('code')
+    if not code:
+        return False
+    _persist_data(code=code)
+    return True
+
+
+def _persist_data(**kwargs):
+    """Save data in session."""
+    if _session_key not in session:
```
alastair: a `collections.defaultdict(dict)` will stop you having to have this guard here

alastair: ah, sorry - I didn't notice that this was provided by flask.

Is this session persisted over server restarts?

acousticbrainz/testing.py
```diff
@@ -21,16 +22,17 @@ def tearDown(self):
...
-        cur.execute("TRUNCATE highlevel CASCADE;")
-        cur.execute("TRUNCATE lowlevel CASCADE;")
-        cur.execute("TRUNCATE statistics CASCADE;")
-        cur.execute("TRUNCATE incremental_dumps CASCADE;")
-        conn.commit()
-        cur.close()
-        conn.close()
+        connection = psycopg2.connect(current_app.config['PG_CONNECT_TEST'])
+        cursor = connection.cursor()
+        cursor.execute('TRUNCATE highlevel_json CASCADE;')
```
alastair: maybe consider using the `RESTART IDENTITY` parameter here too
http://www.postgresql.org/docs/9.1/static/sql-truncate.html

templates/user/profile.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'base.html' %}
+
+{% block title %}Profile - AcousticBrainz{% endblock %}
+
+{% block content %}
+  <h2>Your profile</h2>
+  <dl class="dl-horizontal">
+    <dt>ID</dt><dd>{{ user.id }}</dd>
```
alastair: it's not important to show `id` and `created` on this page, but I guess you need at least something here until we put real content up :)

acousticbrainz/login/provider.py
tsukanov: I think sessions variables are stored client side.

acousticbrainz/data/user.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def get_by_mb_id(musicbrainz_id):
+    try:
+        connection = psycopg2.connect(current_app.config['PG_CONNECT'])
+        cursor = connection.cursor()
+        cursor.execute('SELECT id, created FROM "user" WHERE musicbrainz_id = %s',
+                       (str(musicbrainz_id),))
+    except psycopg2.IntegrityError, e:
+        raise BadRequest(str(e))
```
alastair: similar to usernames - will any of these exceptions ever print stuff that's unicode? Maybe some postgres errors print the query or query args that you gave.

acousticbrainz/data/user_test.py
tsukanov: Testing that new user is actually created. But yeah, I should add more assertions there.

templates/user/profile.html
tsukanov: Yeah, that's just placeholder stuff.

acousticbrainz/data/user_test.py
alastair: but you're not testing first that that user doesn't already exist. Maybe the user was created in a previous test and not removed, and `get_or_create` actually has broken create code.

acousticbrainz/login/provider.py
alastair: oh, I understand what you're using the session data for now. fine

acousticbrainz/data/user.py
tsukanov: There shouldn't be more than one row. See `user_musicbrainz_id_key` constraint.
Here we just check if row has been found. I'll try to rewrite this part to make it simpler.

acousticbrainz/data/user.py
tsukanov: No they shouldn't. http://initd.org/psycopg/docs/usage.html#adaptation-of-python-values-to-sql-types
Will fix.

alastair:
cool, looks good

PR #75, opened 2015-02-05
AB-11: Duplicate data fetching

tsukanov:
Since we don't have a way to order data by it's "goodness", there's no selection of the best dataset. First one is returned if there's no offset.

*Fixes http://tickets.musicbrainz.org/browse/AB-11.*

### DIFF
--- a/acousticbrainz/data/__init__.py
+++ b/acousticbrainz/data/__init__.py
@@ -76,12 +76,13 @@ def submit_low_level_data(mbid, data):
         raise ServiceUnavailable(str(e))
 
 
-def load_low_level(mbid):
+def load_low_level(mbid, offset=0):
     """Load low level data for a given MBID."""
     conn = psycopg2.connect(current_app.config['PG_CONNECT'])
     cur = conn.cursor()
     try:
-        cur.execute("SELECT data::text FROM lowlevel WHERE mbid = %s", (str(mbid), ))
+        cur.execute("SELECT data::text FROM lowlevel WHERE mbid = %s OFFSET %s",
+                    (str(mbid), offset))
         if not cur.rowcount:
             raise NotFound
 
--- a/acousticbrainz/data/__init__.py
+++ b/acousticbrainz/data/__init__.py
@@ -121,6 +121,23 @@ def load_high_level(mbid):
     return InternalServerError("Bummer, dude.")
 
 
+def count_lowlevel(mbid):
+    """Count number of stored datasets for a specified MBID."""
+    conn = psycopg2.connect(current_app.config['PG_CONNECT'])
+    cur = conn.cursor()
+    try:
+        cur.execute("SELECT count(*) FROM lowlevel WHERE mbid = %s",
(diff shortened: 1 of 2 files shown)


tsukanov:
I still need to implement endpoint that would return info about how many duplicates we have. Documentation update would be nice as well.

acousticbrainz/data/__init__.py
```diff
@@ -120,6 +121,23 @@ def load_high_level(mbid):
...
+        cur.execute("SELECT count(*) FROM lowlevel WHERE mbid = %s",
+                    (str(mbid),))
+        return cur.fetchone()[0]
+
+    except psycopg2.IntegrityError, e:
+        raise BadRequest(str(e))
+    except psycopg2.OperationalError, e:
+        raise ServiceUnavailable(str(e))
+
+    return InternalServerError("Ouch!")
```
mayhem: :)

acousticbrainz/views/api.py
```diff
@@ -7,11 +7,32 @@
 api_bp = Blueprint('api', __name__)
 
 
+@api_bp.route("/<uuid:mbid>/meta", methods=["GET"])
```
mayhem: Not sure about the /meta enpoint. Do you foresee returning other metadata here? Because if this is only to be used for a count then /count might be best.

acousticbrainz/views/api.py
```diff
@@ -7,11 +7,32 @@
...
 def get_low_level(mbid):
-    """Endpoint for fetching low-level information to AcousticBrainz."""
-    return Response(load_low_level(mbid), content_type='application/json')
+    """Endpoint for fetching low-level information to AcousticBrainz.
+
+    Offset can be specified if you need to get another dataset in case there
+    are duplicates.
+    """
+    offset = request.args.get("n")
+    if offset:
```
mayhem: If we're calling it Offset, maybe the arg should be "o"?

alastair: I said n because I thought it was a `number`

mayhem: It is a number. But rather than describing the type, I think we should describe the content...

acousticbrainz/views/api.py
alastair: yes, there was a possibility that we could add other metadata here. any ideas for other metadata? :)

acousticbrainz/views/api.py
mayhem: In IRC we decided to go with "n" as is.

acousticbrainz/views/api.py
mayhem: In IRC we decided to go with /count instead.

mayhem:
After you address the comments, feel free to merge.

tsukanov:
Okay, I'm merging this. Documentation is not updated yet, but I created a ticket for that - http://tickets.musicbrainz.org/browse/AB-31.

:rocket:

PR #76, opened 2015-02-06
AB-31: Documentation update

tsukanov:
*Fixes http://tickets.musicbrainz.org/browse/AB-31.*

### DIFF
--- a/templates/data/data.html
+++ b/templates/data/data.html
@@ -36,6 +36,7 @@
 <pre>
    GET http://acousticbrainz.org/&lt;MBID&gt;/low-level?n=1
 </pre>
+  <p>Documents are counted from 0 to count-1.</p>
   <p><em>Example:</em></p>
 <pre>
    GET <a href="http://acousticbrainz.org/96685213-a25c-4678-9a13-abd9ec81cf35/low-level?n=1">http://acousticbrainz.org/96685213-a25c-4678-9a13-abd9ec81cf35/low-level?n=1</a>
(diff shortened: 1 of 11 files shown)


mayhem:
Why would we count from 1?

And what do you mean "list the taco"? Given that we're just returning the first one, this doesn't make a lot of sense to me.

alastair:
maybe we want to count from 1 because 0 is the same as no ?n= parameter. If we count from 0 (I'm fine with this too) we need to explicitly say that we count from 0-count-1

Yes, now we return just the first one, but our conversations have always said that we're going to work to select the "best" taco from all duplicates, this might be `?n=3`, but there's no way of telling people that that's what they're going to get if they just say /mbid/low-level

alastair:
OK, let's ignore the discussion about /meta stuff. I just want clarification on what we count from.

mayhem:
OK, let's:

* count from 0.
* 0 is the same as no n param
* TODO: document that we count from 0 -> count - 1

tsukanov:
Any suggestions on how I can describe this parameter better?

templates/data/data.html
```diff
@@ -1,52 +1,75 @@
...
+   GET <a href="http://acousticbrainz.org/96685213-a25c-4678-9a13-abd9ec81cf35/count">http://acousticbrainz.org/96685213-a25c-4678-9a13-abd9ec81cf35/count</a>
+</pre>
+  <p>If there's more than one document, you can specify a number argument <em>n</em>:</p>
+<pre>
+   GET http://acousticbrainz.org/&lt;MBID&gt;/low-level?n=1
+</pre>
+  <p><em>Example:</em></p>
+<pre>
+   GET <a href="http://acousticbrainz.org/96685213-a25c-4678-9a13-abd9ec81cf35/low-level?n=1">http://acousticbrainz.org/96685213-a25c-4678-9a13-abd9ec81cf35/low-level?n=1</a>
+</pre>
```
alastair: "documents are counted from 0 to count-1, and matches the suffixes added to documents in the json download"

alastair:
:ship: it

PR #78, opened 2015-02-13
AB-34: Server setup/import additions

tsukanov:
*Fixes http://tickets.musicbrainz.org/browse/AB-34.*


README.md
```diff
@@ -41,6 +41,22 @@ Full installation instructions are available in [INSTALL.md](https://github.com/
...
+AcousticBrainz provides data dumps that you can import into your own server.
+Latest database dump is available at http://acousticbrainz.org/download. You
+need to download full database dump from this page and use it during database
+initialization:
+
+    $ python manage.py init_db -a <path_to_the_archive>
+
+or import archive after database is created:
+
+    $ python manage.py import_data -a <path_to_the_archive>
```
alastair: This will import when the FKs and indexes are created, which is slow. How do we recommend against this?

manage.py
```diff
@@ -26,6 +28,11 @@ def init_db(archive=None):
     More information about populating a PostgreSQL database efficently can be
     found at http://www.postgresql.org/docs/current/static/populate.html.
     """
+    print('Creating user and a database...')
+    exit_code = subprocess.call('sudo -u postgres psql < ' + os.path.join('admin', 'sql', 'create_db.sql'), shell=True)
```
alastair: If the user already exists and the database doesn't, how will this fail? 

e.g. I just deleted the database before I did a test import, but the user still existed

README.md
tsukanov: Maybe delete them for importing and recreate after?

alastair: what if there is already data in the database?

tsukanov: Deleting indexes shouldn't be a problem.

manage.py
tsukanov: It will not create a database. I can add an argument that will allow to skip database creation if that's something that will be helpful.

README.md
alastair: but importing data with duplicate ids will fail.

I'm tending towards having the import create the database itself. If it already exists, you must use `--force` or `--delete` to indicate that you know it will drop and recreate the db

alastair: After discussing this in IRC, I realised I confused the situations of no database, database with no tables, and database with tables. I propose these commands:

 - createdb and tables to run your own server
 - createdb, make tables, import when no db exists previously
 - if the database exists, add a force or drop option to delete the database first before doing the above
 - import a dump into already created database and tables (for updates, not recommended for initial import because of the indexes)

Only the third option is missing, I believe.

tsukanov: Done.

PR #79, opened 2015-02-19
Redirect from /recording/<mbid> to summary page

tsukanov:


acousticbrainz/views/data.py
```diff
@@ -10,7 +10,7 @@
 
 @data_bp.route("/api")
 def api():
-    return redirect("/data")
+    return redirect(url_for(".data"))
```
mayhem: I wonder which slacker left this here.

mayhem:
:+1:

PR #81, opened 2015-03-05
AB-7: Replaced D3.js with Highcharts library on statistics page

tsukanov:
Performance should be significantly better now. Apart from that, there are several other useful features in this library. See http://api.highcharts.com/highstock.

:tada: :chart_with_upwards_trend: :tada:

*Fixes http://tickets.musicbrainz.org/browse/AB-7.*


ianmcorvidae:
Looks fine to me; I don't know highcharts but I trust you know what you're doing. I see something for a range-selector, which would be my main hope -- hopefully there's also still tooltips. But I think it's fine in any case, the d3/nvd3 thing was largely experimental :)

tsukanov:
Yeah, it has both range selector and tooltips.

tsukanov:
Here's how it looks like:
Screenshot of the chart

alastair:
is it faster? if so, :ship: it

tsukanov:
Yeah, should be significantly faster than the previous implementation. :shipit:

PR #83, opened 2015-03-09
[WIP] Datasets

gentlecat:
:books:


tsukanov:
We set up a test server and this branch is live there: http://test.acousticbrainz.org/datasets/create/. Feel free to test and report any issues/suggestions.

tsukanov:
I could use some suggestions about user interface. At least some general direction to have something to work on.
Basic functionality for creation and editing is now working.

cc @mayhem, @alastair

acousticbrainz/__init__.py
```diff
@@ -4,7 +4,7 @@
 from logging.handlers import RotatingFileHandler
 
 # This value must be incremented after schema changes on replicated tables!
-__version__ = 3
+__version__ = 4
```
alastair: Are datasets going to be replicated at the same time as the data?

acousticbrainz/data/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+
+    except (psycopg2.ProgrammingError, psycopg2.IntegrityError, psycopg2.OperationalError) as e:
+        # TODO: Log this.
+        return None, e
+
+    return dataset_id, None
+
+
+def update(dataset_id, dictionary, author_id):
+    try:
```
alastair: Are datasets always going to be updated by passing a full dictionary to this method? what if we have 100,000 items?

acousticbrainz/data/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+    Returns:
+        Tuple with two values: new dataset ID and error. If error occurs first
+        will be None and second is an exception. If there are no errors, second
+        value will be None.
+    """
+    try:
+        connection = psycopg2.connect(current_app.config["PG_CONNECT"])
+        cursor = connection.cursor()
+
+        cursor.execute("""INSERT INTO dataset (id, name, description, author)
```
alastair: I'd like to see this split into a few different methods for reuse, and testability (you have tests, right? ;)

acousticbrainz/data/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+        connection = psycopg2.connect(current_app.config["PG_CONNECT"])
+        cursor = connection.cursor()
+
+        cursor.execute("""UPDATE dataset
+                          SET (name, description, author) = (%s, %s, %s)
+                          WHERE id = %s""",
+                       (dictionary["name"], dictionary["description"], author_id, dataset_id))
+
+        # Replacing old classes with new ones
+        cursor.execute("""DELETE FROM class WHERE dataset = %s""", (dataset_id,))
```
alastair: Is it possible to do a diff between what exists and what is sent, and only make changes as necessary?

acousticbrainz/data/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+    "properties": {
+        "name": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 100
+        },
+        "description": {"type": "string"},
+        "classes": {
+            "type": "array",
+            "minItems": 2,  # must have at least two classes
```
alastair: For creating a dataset at first, there shouldn't be a minimum number of classes

acousticbrainz/data/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+                    "name": {
+                        "type": "string",
+                        "minLength": 1,
+                        "maxLength": 100
+                    },
+                    "description": {"type": "string"},
+                    "recordings": {
+                        # CLASS_MEMBER
+                        "type": "array",
+                        "minItems": 1,  # must have at one recording
```
alastair: Same as above, when you create a class, an item isn't required

alastair:
New dataset
---
 * You should be able to create a dataset with no classes, and classes with no items
 * There no text description when you type in a non-uuid as a recording id
 * It's not clear that you then have to press "add recording to add it" - I thought you needed to press it only if you wanted >1 recording

edit dataset
---
* You should be able to just add and delete classes without having to deal with recordings
* There should be at least a bulk way of adding uuids per class. Maybe a 'edit class' field, with an option to type 1 uuid or a list (newline separated)

acousticbrainz/views/datasets.py
```diff
@@ -0,0 +1,110 @@
...
+        return jsonify(
+            success=True,
+            dataset_id=dataset_id,
+        )
+    return render_template('datasets/editor.html', mode="create")
+
+
+@datasets_bp.route('/<uuid:id>/edit', methods=('GET', 'POST'))
+@login_required
+def edit(id):
```
alastair: For large datasets it'd be nice to have endpoints for dealing with single recordings and classes

acousticbrainz/data/dataset.py
tsukanov: I can add more "fine-grained" methods for updating specific class or adding/removing recordings from it. Current implementation there is just a consequence of the way editing currently works.

alastair: ok, sounds good

admin/sql/create_tables.sql
```diff
@@ -42,4 +42,24 @@ CREATE TABLE "user" (
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 
+CREATE TABLE dataset (
+  id          UUID,
+  name        VARCHAR NOT NULL,
+  description TEXT,
+  author      INT, -- FK to user
+  created     TIMESTAMP WITH TIME ZONE DEFAULT NOW()
```
alastair: We need to make sure we have a future task to be able to "publish" datasets. They should be private by default

tsukanov:
Related IRC discussion logs: http://chatlogs.musicbrainz.org/musicbrainz-devel/2015/2015-04/2015-04-08.html#T08-03-30-589228.

gentlecat:
I need some suggestions about efficiently getting information about recordings. *Do we actually need to support adding recordings that are not in AcousticBrainz?* We'll need to check if MBID is associated with recording (on both server and client) and get basic information about that recording like artist name and title (client). It should be easy to do on the client by getting information from the MusicBrainz Web Service. On the server, however, it might not be that simple if we'll have to rely on information from MusicBrainz instead of using AcousticBrainz database.

alastair:
We talked about this in IRC a bit. I agree that it's too much work now to be able to add non-ab mbids currently.

Checking if an mbid is valid (is a recording) should probably be done server-side. We can do this easily without going to musicbrainz because we have all mbids. We can look at this again in the future if we allow non-ab recordings.

I'm not convinced that we should do metadata lookups from the browser. One way we could do it is to do the lookup on the server, but behind a cache, so that for many people looking at the same dataset, it won't go to musicbrainz all the time.

gentlecat:
I was thinking only about editing, but it seems sensible to do that on the server since we'll need that data for displaying datasets. Did you mean caching on the server for a certain amount of time?

alastair:
merge it! let's start using it and see what works

PR #85, opened 2015-03-20
Added Less compiler setup into install_server.sh

tsukanov:


admin/install_server.sh
```diff
@@ -21,6 +21,11 @@ echo "client_encoding = utf8" >> "$PG_CONF"
 
 service postgresql restart
 
+# Less compiler
+curl -sL https://deb.nodesource.com/setup | sudo bash -
```
alastair: Is the node in our base image too old?
If it is, and we need to install the package ourselves, please do this step manually (copy list file to sources.list.d, import gpg key) instead of using curl|bash

or can we just install node-less directly? (I know this can be problematic because it's often quite old)

tsukanov: I'm getting weird errors after installing less compiler using node and npm from the base image:

```
root@acousticbrainz-server:/vagrant/static/css# lessc
/usr/bin/env: node: No such file or directory
```

Installing node-less directly works and it compiles styling fine. I don't know if we'll need more recent version of nodejs in the future (for compiling JSX and other things).

mwiencek: Just a guess, but that sounds like you need the `nodejs-legacy` package, which provides a symlink to `node`.

alastair: I'm tempted to use node-less for now, since it works, and involves less magic. Let's re-address this if we need advanced support in the future

tsukanov: Yeah, it works with `nodejs-legacy`. Thanks, Michael.

alastair:
:ship: it

PR #86, opened 2015-03-27
Fixed `debug` argument in server.py

tsukanov:
Before that it was kind of useless, same goes for `DEBUG` parameter in config file. Debug mode was always on. :confounded:

### DIFF
--- a/server.py
+++ b/server.py
@@ -5,9 +5,15 @@ import argparse
 application = create_app()
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(description='AcousticBrainz dev server')
-    parser.add_argument("-d", "--debug", help="Turn on debugging mode to see stack traces in the error pages", default=True, action='store_true')
-    parser.add_argument("-t", "--host", help="Which interfaces to listen on. Default: 0.0.0.0", default="0.0.0.0", type=str)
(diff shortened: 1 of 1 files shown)


server.py
```diff
@@ -5,9 +5,15 @@
...
+                        help="Turn on debugging mode to see stack traces on "
+                             "the error pages. This overrides 'DEBUG' value "
+                             "in config file.")
+    parser.add_argument("-t", "--host", default="0.0.0.0", type=str,
+                        help="Which interfaces to listen on. Default: 0.0.0.0.")
+    parser.add_argument("-p", "--port", default="8080", type=int,
+                        help="Which port to listen on. Default: 8080.")
     args = parser.parse_args()
-    application.run(debug=True, host=args.host, port=args.port)
+    application.run(debug=True if args.debug else None,
```
alastair: Are the arguments to `debug` really `True` and `None`?

tsukanov: `None` is a default value there, which means it doesn't override whatever value we have set in the config file. See http://flask.pocoo.org/docs/0.10/api/#flask.Flask.run and http://flask.pocoo.org/docs/0.10/api/#flask.Flask.debug.

alastair:
:ship: it

PR #87, opened 2015-04-01
Changes related to http://tickets.musicbrainz.org/browse/AB-38

swapnilgt:
Currently we show basic 404 page when there's no data for recording. Since we have integration with MusicBrainz, we can load some basic info about recording.


acousticbrainz/data/__init__.py
```diff
@@ -137,15 +137,22 @@ def count_lowlevel(mbid):
 
     return InternalServerError("Ouch!")
 
-
```
tsukanov: https://www.python.org/dev/peps/pep-0008/#blank-lines

acousticbrainz/data/__init__.py
```diff
@@ -186,6 +189,7 @@ def get_summary_data(mbid):
     return InternalServerError("whoops!")
 
 
+
```
tsukanov: Same as above (https://www.python.org/dev/peps/pep-0008/#blank-lines).

acousticbrainz/data/__init__.py
```diff
@@ -137,15 +137,22 @@ def count_lowlevel(mbid):
...
     mbid = str(mbid)
     conn = psycopg2.connect(current_app.config['PG_CONNECT'])
     cur = conn.cursor()
     try:
         cur.execute("SELECT data FROM lowlevel WHERE mbid = %s", (mbid, ))
         if not cur.rowcount:
-            raise NotFound("We do not have data for this track. Please upload it!")
+            # Returning all None as the lowelevel data not found !!
+            return lowlevel, highlevel, genres, moods, other
```
tsukanov: Maybe it's better to return a dictionary there and pass it into the template? I feel like there are way too many variables returned there. This will also help get rid of initialization stuff above.

acousticbrainz/views/data.py
```diff
@@ -46,22 +47,41 @@ def view_high_level(mbid):
 def summary(mbid):
     lowlevel, highlevel, genres, moods, other = get_summary_data(mbid)
 
-    info = _get_track_info(mbid, lowlevel['metadata'])
+    info = None
```
tsukanov: Why do you need that?

acousticbrainz/views/data.py
```diff
@@ -46,22 +47,41 @@ def view_high_level(mbid):
...
+
+    if lowlevel == None: # No metadata in acousticbrainz.. try to fetch data from musicbrainz
+        info = _get_track_info(mbid)
+    else: # Try to get metadata from musicbrainz with second priority as the one retrieved from acoustic
+        info = _get_track_info(mbid, lowlevel['metadata'])
```
tsukanov: You can simplify this by defining call to `_get_track_info` only once.

    info = _get_track_info(mbid, lowlevel['metadata'] if lowlevel else None)

acousticbrainz/views/data.py
```diff
@@ -46,22 +47,41 @@ def view_high_level(mbid):
...
+
```
tsukanov: :question:

acousticbrainz/views/data.py
```diff
@@ -46,22 +47,41 @@ def view_high_level(mbid):
...
+    
+    elif info: # When there is only metadata... TODO: Move into the same template with if conditions...
+        return render_template("data/summary-metadata.html", info = info, tomahawk_url = tomahawk_url, mbid=mbid)
+    
+    else: # When there is no data
+        raise NotFound("MusicBrainz does not have data for this track. Please upload it!")
 
-def _get_track_info(mbid, metadata):
+
+def _get_track_info(mbid, metadata = None): # Defaulting the metadata to None
```
tsukanov: Comment seems useless there since code is pretty obvious. This applies to some other comments you added as well. See http://blog.codinghorror.com/coding-without-comments/.

tsukanov:
Are you still working on that?

swapnilgt:
Yes, I am still working on this. Will work on your comments over this weekend.

tsukanov:
Okay, great. Also, don't forget to mention something about missing data and how to submit it. :wink:

swapnilgt:
Should I create separate branch for pull request now or should I use the same pull request? What will be easier for you?

tsukanov:
No, that was a tip for the future. Update this one. :smiley:

acousticbrainz/views/data.py
```diff
@@ -44,21 +45,44 @@ def view_high_level(mbid):
...
+    if 'lowlevel' in summary:
+        lowlevel = summary['lowlevel']
+    if 'highlevel' in summary:
+        highlevel = summary['highlevel']
+    if 'genres' in summary:
+        genres = summary['genres']
+    if 'moods' in summary:
+        moods = summary['moods']
+    if 'other' in summary:
+        other = summary['other']
```
tsukanov: Code above can be converted to one-liners which would make it easier to read. For example:

```python
lowlevel = summary['lowlevel'] if 'lowlevel' in summary else None
```

Alternatively (which might be a better option), you can modify ``get_summary_data`` function to return ``None`` in dictionary values if data is missing. That way you won't even need to check if these values are defined.

acousticbrainz/views/data.py
```diff
@@ -44,21 +45,44 @@ def view_high_level(mbid):
...
         tomahawk_url = "http://toma.hk/embed.php?artist={artist}&title={title}".format(
             artist=quote_plus(info['artist'].encode("UTF-8")),
             title=quote_plus(info['title'].encode("UTF-8")))
 
-    return render_template("data/summary.html", lowlevel=lowlevel, highlevel=highlevel,
-                           genres=genres, moods=moods, other=other, mbid=mbid,
-                           info=info, tomahawk_url=tomahawk_url)
+    if lowlevel and info: # When all the data is available
+        return render_template("data/summary.html", lowlevel=lowlevel, highlevel=highlevel,
+                                genres=genres, moods=moods, other=other, mbid=mbid,
```
tsukanov: You can just pass ``summary`` there without extracting different parts of it into separate variables. *Template will need to be modified slightly.*

tsukanov:
I think *data/summary-metadata.html* template that you created should share code with existing *data/summary.html*. You can extract code that renders recording metadata into a [macro](http://jinja.pocoo.org/docs/dev/templates/#macros) and put it into a separate file.

swapnilgt:
Instead creating a separate summary-metadata.html, I am thinking of handling this with some basic if statements inside summary.html. What do you think?

tsukanov:
Yeah, that sounds like a better option.

acousticbrainz/data/__init__.py
```diff
@@ -160,23 +162,25 @@ def get_summary_data(mbid):
...
-        highlevel = None
         if cur.rowcount:
             row = cur.fetchone()
             highlevel = row[0]
+            summary['highlevel'] = highlevel
             try:
                 genres, moods, other = interpret_high_level(highlevel)
+                summary['genres'] = genres
+                summary['moods'] = moods
+                summary['other'] = other
```
tsukanov: ``summary['genres'], summary['moods'], summary['other'] = interpret_high_level(highlevel)``?

acousticbrainz/views/data.py
```diff
@@ -44,21 +45,23 @@ def view_high_level(mbid):
 
 @data_bp.route("/<uuid:mbid>", methods=["GET"])
 def summary(mbid):
-    lowlevel, highlevel, genres, moods, other = get_summary_data(mbid)
-
-    info = _get_track_info(mbid, lowlevel['metadata'])
+    summary = get_summary_data(mbid)
+    
+    info = None
```
tsukanov: :question:

acousticbrainz/views/data.py
```diff
@@ -44,21 +45,23 @@ def view_high_level(mbid):
...
             artist=quote_plus(info['artist'].encode("UTF-8")),
             title=quote_plus(info['title'].encode("UTF-8")))
 
-    return render_template("data/summary.html", lowlevel=lowlevel, highlevel=highlevel,
-                           genres=genres, moods=moods, other=other, mbid=mbid,
-                           info=info, tomahawk_url=tomahawk_url)
+    if info:
+        return render_template("data/summary.html", summary=summary, mbid=mbid, info=info, tomahawk_url=tomahawk_url)
+    else: # When there is no data
+        raise NotFound("MusicBrainz does not have data for this track. Please upload it!")
```
tsukanov:
```
NameError: global name 'NotFound' is not defined
```

templates/data/summary.html
```diff
@@ -52,72 +52,77 @@
...
+      {% if other -%}
+        <li>
+          <a href="{{ url_for('data.view_high_level', mbid=mbid) }}">High-level data</a>
+          (<a href="{{ url_for('api.get_high_level', mbid=mbid) }}">raw JSON</a>)
+        </li>
+      {%- endif %}
+    </ul>
+    {% else %}
+    <div class="alert alert-warning">   
+          Sorry, but we don't have more information for this track. You have this track on your hard drive? Great !! You can be the first one to contribute for this track. You just need to run a very simple extractor. Find more information <a href="{{ url_for('index.contribute') }}">here</a>.
```
tsukanov: Not sure about this text. @alastair?

alastair: How about:

We don't have any information about this recording yet. If you have it you can help us increase the size and quality of the AcousticBrainz database. You just need to run a very simple tool. <a href="{{ url_for('index.contribute') }}">Find more information here</a>

acousticbrainz/views/data.py
```diff
@@ -97,6 +101,10 @@ def _get_track_info(mbid, metadata):
         else:
             info['track_number'] = metadata['tags']['tracknumber'][0]
 
-    info['length'] = metadata['audio_properties']['length_formatted']
-
+    # Try to get length from abz data
```
alastair: This can be moved up into the main `if` block

acousticbrainz/data/__init__.py
```diff
@@ -139,13 +139,15 @@ def count_lowlevel(mbid):
...
+    """Fetches the lowlevel and highlevel features from abz database for the specified MBID."""
+    summary = {'lowlevel': None, 'highlevel': None, 'genres': None, 'moods': None, 'other': None}
     mbid = str(mbid)
     conn = psycopg2.connect(current_app.config['PG_CONNECT'])
     cur = conn.cursor()
     try:
         cur.execute("SELECT data FROM lowlevel WHERE mbid = %s", (mbid, ))
         if not cur.rowcount:
-            raise NotFound("We do not have data for this track. Please upload it!")
+            return summary
```
alastair: I like how summary is moved into a dictionary, but this can still be an exception (not the flask one any more). In views/data.py you can catch it and set it to something else

acousticbrainz/views/data.py
```diff
@@ -44,21 +46,21 @@ def view_high_level(mbid):
...
-    lowlevel, highlevel, genres, moods, other = get_summary_data(mbid)
-
-    info = _get_track_info(mbid, lowlevel['metadata'])
-
+    summary = get_summary_data(mbid)
+    
+    info = _get_track_info(mbid, summary['lowlevel']['metadata'] if summary['lowlevel'] else None)
     # Tomahawk player stuff
-    if not ('artist' in lowlevel['metadata']['tags'] and 'title' in lowlevel['metadata']['tags']):
+    if not ('artist' in info and 'title' in info):
```
alastair: It's not directly related to your change, but the tomahawk_url code should be a separate method, taking `metadata` as an argument

templates/data/summary.html
```diff
@@ -52,72 +52,77 @@
...
-    {%- endif %}
-  </ul>
+      {% if other -%}
+        <li>
+          <a href="{{ url_for('data.view_high_level', mbid=mbid) }}">High-level data</a>
+          (<a href="{{ url_for('api.get_high_level', mbid=mbid) }}">raw JSON</a>)
+        </li>
+      {%- endif %}
+    </ul>
+    {% else %}
```
alastair: This might be better with two templates - a base template shows basic metadata + player, then the two templates show either this notice or the extracted data. This way we don't get messy templates.

acousticbrainz/views/data.py
```diff
@@ -44,21 +46,21 @@ def view_high_level(mbid):
 
 @data_bp.route("/<uuid:mbid>", methods=["GET"])
 def summary(mbid):
-    lowlevel, highlevel, genres, moods, other = get_summary_data(mbid)
-
-    info = _get_track_info(mbid, lowlevel['metadata'])
-
+    summary = get_summary_data(mbid)
+    
+    info = _get_track_info(mbid, summary['lowlevel']['metadata'] if summary['lowlevel'] else None)
```
alastair: if you throw an exception from `get_summary_data` you can clean up this call:

```python
try:
    summary = get_summary_data(mbid)
    metadata = summary['lowlevel'].get('metadata') # you get `None` for free with .get
    template = ""
    args = {"some": "args_for_template"}
except TheException:
    metadata = None
    template = ""
    args = {"some": "args_for_template"}
info = _get_track_info(mbid, metadata)
tomahawk_url = _get_tomahawk_url(info) # since we have musicbrainz data we may as well use that
...
return render_template(template, **args)
```

templates/data/summary.html
alastair: Ah, I see you already discussed this with @tsukanov. Because we will be making these pages more complex in the future my preference is to have separate templates. I don't mind if it's done with block inheritance or macros

templates/data/summary.html
swapnilgt: I am not sure if we should we keep the message to be very formal. A little informal touch to the message can be quite attractive to the user. But again, I am not very sure if we want something very formal or informal.

acousticbrainz/data/__init__.py
swapnilgt: If I get this right, you are talking about the exceptions that might be caused during the execution of the sql query right? Isn't this thing already handled with the try block written here?

acousticbrainz/views/data.py
swapnilgt: We can add this as a low priority ticket.. just to keep track.

acousticbrainz/data/__init__.py
alastair: No, I mean that in the case where there is no row we should throw an exception instead of returning an empty dictionary.

acousticbrainz/views/data.py
alastair: Sure, add a ticket if you want. To me, it's a 6 line change to copy this to a new function, in a part of the code you're already working on, which is why I suggested we make the change now.

swapnilgt: Yeah.. changing now itself seems reasonable.

acousticbrainz/data/__init__.py
gentlecat: This should return `HTTP 404`.

```python
FAIL: test_summary (acousticbrainz.views.data_test.DataViewsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/vagrant/acousticbrainz/views/data_test.py", line 28, in test_summary
    self.assertEqual(resp.status_code, 404)
AssertionError: 200 != 404
```

acousticbrainz/views/data.py
```diff
@@ -44,21 +46,25 @@ def view_high_level(mbid):
...
 
-    # Tomahawk player stuff
-    if not ('artist' in lowlevel['metadata']['tags'] and 'title' in lowlevel['metadata']['tags']):
-        tomahawk_url = None
-    else:
-        tomahawk_url = "http://toma.hk/embed.php?artist={artist}&title={title}".format(
-            artist=quote_plus(info['artist'].encode("UTF-8")),
-            title=quote_plus(info['title'].encode("UTF-8")))
+    if info:
+        return render_template("data/summary.html", summary=summary, mbid=mbid, info=info, tomahawk_url=tomahawk_url)
```
gentlecat: This view should return status `404` when we have no data for this track.

```python
FAIL: test_summary (acousticbrainz.views.data_test.DataViewsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/vagrant/acousticbrainz/views/data_test.py", line 28, in test_summary
    self.assertEqual(resp.status_code, 404)
AssertionError: 200 != 404
```

acousticbrainz/views/data.py
```diff
@@ -44,21 +46,25 @@ def view_high_level(mbid):
...
+    else:  # When there is no data
+        raise NotFound("MusicBrainz does not have data for this track. Please upload it!")
```
gentlecat: If there is no *info* it means that this recording doesn't exist in MusicBrainz. In which case you can't really upload anything. Might want to remove "Please upload it!" part.

acousticbrainz/views/data.py
gentlecat: Also, you can replace second `tomahawk_url` on this line with `_get_tomahawk_url(info)` and remove assignment above.

templates/data/summary.html
gentlecat: I think messages like this one should be neutral. Forcing users to do something is not a good idea and it can't be appealing to everyone.

acousticbrainz/data/__init__.py
```diff
@@ -139,13 +139,15 @@ def count_lowlevel(mbid):
...
+    """Fetches the lowlevel and highlevel features from abz database for the specified MBID."""
+    summary = {}
     mbid = str(mbid)
     conn = psycopg2.connect(current_app.config['PG_CONNECT'])
     cur = conn.cursor()
     try:
         cur.execute("SELECT data FROM lowlevel WHERE mbid = %s", (mbid, ))
         if not cur.rowcount:
-            raise NotFound("We do not have data for this track. Please upload it!")
+            raise NotFound
```
gentlecat: You shouldn't use `NotFound` exception from werkzeug there. It's meant for use in case of HTTP errors, not our internal data fetching. So a better thing to do there would be to create your own exception.

acousticbrainz/views/data.py
```diff
@@ -44,21 +46,30 @@ def view_high_level(mbid):
...
 def summary(mbid):
-    lowlevel, highlevel, genres, moods, other = get_summary_data(mbid)
-
-    info = _get_track_info(mbid, lowlevel['metadata'])
-
-    # Tomahawk player stuff
-    if not ('artist' in lowlevel['metadata']['tags'] and 'title' in lowlevel['metadata']['tags']):
-        tomahawk_url = None
+    try:
+        summary = get_summary_data(mbid)
```
gentlecat: Name of this variable is the same as the function. I'm not sure if it's a good idea to do that.

acousticbrainz/data/__init__.py
gentlecat: Or you could just return empty dict like we do in `_get_track_info`.

acousticbrainz/views/data.py
```diff
@@ -44,21 +46,30 @@ def view_high_level(mbid):
...
+    except NotFound:
+        summary = {}
+    
+    info = _get_track_info(mbid, summary['lowlevel']['metadata'] if summary and summary['lowlevel'] else None)
+    if info and summary and summary['lowlevel']:
```
gentlecat: I think `and summary['lowlevel']` checks in this line and the line above are redundant.

gentlecat:
I think we should merge this after things that I pointed out earlier are fixed, unless @alastair has any other suggestions.

alastair:
Yes, I agree. Ready to merge after these points are addressed

gentlecat:
Thanks a lot, @swapnilgt! :shipit:

PR #88, opened 2015-04-29
[WIP] Process many highlevel tracks with the same binary

alastair:
Requires updates to `streaming_extractor_music_svm`

This does 2 things:
- allow you to do `streaming_extractor_music_svm in out in2 out2 in3 out3 profile` to process files quickly. This is quicker because the slowest part of the process is loading the model, not doing the conversion
- Let you specify a highlevel table and profile file on the commandline to `hl_calc.py`. This is if we want to have a table of documents processed with an independent model.
#### still to be done
- [ ] table name input to `hl_calc` is not sanitised, and is passed directly into the sql with string formatting. oops
- [ ] If we process 50 files at once per binary, sometimes it might fail because _1_ file is broken. If this is the case, we should process each file individually so that we skip just the broken one. This will be slow while it processes 49 files, but it'll mean that we get all of their results
- [ ] There may be some weird race conditions - specifically, I'm seeing https://github.com/metabrainz/acousticbrainz-server/blob/hlmulti/high-level/hl_calc.py#L267 fail sometimes. This means that either the mbid wasn't added to `in_progress`, or it was removed by two threads. Also saw once that https://github.com/metabrainz/acousticbrainz-server/blob/hlmulti/high-level/hl_calc.py#L191 failed - obviously two threads picked up the same item.

### DIFF
--- a/high-level/hl_calc.py
+++ b/high-level/hl_calc.py
@@ -35,7 +35,7 @@ class HighLevel(Thread):
 
     def __init__(self, input_data, profile_file):
         Thread.__init__(self)
-        # (mbid, ll_id, ll_data)
+        # (mbid, ll_data, ll_id)
         self.input_data = input_data
         # (mbid, ll_id, hl_data)
         self.output_data = []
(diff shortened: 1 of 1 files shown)


alastair:
I just realised that the `in_progress` pop problem would manifest itself with duplicates! We should use `ll_id` to keep track of these instead.

Also, we should stop using mbids as a key into the threadpool, since threads process more than 1 mbid.

alastair:
Another thought, we could consider marking failed low-level rows somehow, so we don't continue to pick them up and dropping back to slow processing.

PR #89, opened 2015-05-04
More tests and minor tweaks

gentlecat:


acousticbrainz/views/data_test.py
```diff
@@ -31,3 +37,77 @@ def test_summary(self):
...
         resp = self.client.get(url_for('data.summary', mbid=mbid))
         self.assertEqual(resp.status_code, 200)
+
+        # We don't have data for this recording, but it exists in MusicBrainz.
+        resp = self.client.get(url_for('data.summary', mbid='770cc467-8dde-4d22-bc4c-a42f91e7515e'))
+        self.assertEqual(resp.status_code, 404)
+        self.assertIn("We don't have any information about this recording yet.", resp.data)
+
+
+class FakeMusicBrainz(object):
```
alastair: This is fine now, but will we need it to be extended in the future? If yes, make sure we do it smarter (include xml in data dir, maybe) instead of this

alastair:
fine by me

acousticbrainz/views/data_test.py
gentlecat: XML with sample responses from MusicBrainz? Yeah, that's probably a better idea.

alastair: yes, right

PR #90, opened 2015-05-04
Improved incremental dumps (AB-43)

gentlecat:
I made it so that rows in all tables (not only in `lowlevel`, `highlevel`, and `highlevel_json`) are split up for incremental dumps.

*This should fix [AB-43](http://tickets.musicbrainz.org/browse/AB-43).*

### DIFF
--- a/acousticbrainz/data/dump.py
+++ b/acousticbrainz/data/dump.py
@@ -122,54 +122,60 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
 
 
 def _copy_tables(location, start_time=None, end_time=None):
-    """Copies all tables into separate files within a specified directory.
+    """Copies all tables into separate files within a specified location (directory).
 
-    You can also define time frame that will be used during data selection.
-    This will only work on tables that support incremental dumping: "lowlevel"
-    and "highlevel".
+    ou can also define time frame that will be used during data selection.
+    Files in a specified directory will only contain rows that have timestamps
+    within specified time frame. We assume that each table contains some sort
+    of timestamp that can be used as a reference.
     """
     conn = psycopg2.connect(current_app.config["PG_CONNECT"])
     cursor = conn.cursor()
 
-    # Copying tables that can be split up for incremental dumps
-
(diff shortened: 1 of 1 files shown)


acousticbrainz/data/dump.py
```diff
@@ -122,54 +122,60 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
...
-        with open(os.path.join(location, table), "w") as f:
-            cursor.copy_to(f, table, columns=_TABLES[table])
+    # statistics
+    with open(os.path.join(location, "statistics"), "w") as f:
+        print(" - Copying table statistics...")
+        cursor.copy_to(f, "(SELECT %s FROM statistics %s)" %
+                       (", ".join(_TABLES["statistics"]), generate_where("collected")))
+
+    # incremental_dumps
+    with open(os.path.join(location, "incremental_dumps"), "w") as f:
```
alastair: does it make sense to dump the incremental dumps table?
 
- positive feature: anyone who has all of the dumps has a full dump of the db (except for our un-synched tables)
- negative feature: who actually cares about it?
- Maybe it's a good way of seeing what dump you're up to, so that you know which one to get next

gentlecat: Yeah, I'm unsure. It can be useful if we need to implement automated data updates, to know which was the last imported incremental dump.

gentlecat: Maybe we should do something similar to MBS. http://chatlogs.musicbrainz.org/musicbrainz-devel/2015/2015-05/2015-05-06.html#T18-04-12-203451

alastair:
This patch is OK for now - it will mean that incremental dumps will apply (will they? we still have the highlevel/highleveljson problem as well I think).

Do we want to also consider a musicbrainz-style `replication_control` table so that mirrors know what dump they're up to and what incremental to ask for next?

gentlecat:
Yeah, let's think about implementing something similar.

alastair:
OK, merge as is, and let's open another ticket for making the new changes to replication

PR #91, opened 2015-05-04
AB-41: Added MBID verification into data submission process

gentlecat:
*Fixes [AB-41](http://tickets.musicbrainz.org/browse/AB-41).*


gentlecat:
There's one caveat though... I don't know if there's a good way to check whether recoding is not found or MusicBrainz is unavailable. Thoughts?

alastair:
My concern with this is there is now a clear bottleneck if we ever get a high submission rate (especially after we release a new extractor), and that if there are any problems with accessing musicbrainz (not actually completely unheard of), we're going to reject potentially good submissions from contributors.
I'd prefer to do this offline after submission.

gentlecat:
I agree, that will be better.

PR #92, opened 2015-05-04
Updated installation scripts (AB-42)

gentlecat:
*Fixes [AB-42](http://tickets.musicbrainz.org/browse/AB-42).*

### DIFF
--- a/admin/install_hl_extractor.sh
+++ b/admin/install_hl_extractor.sh
@@ -1,9 +1,17 @@
 #!/bin/sh
 
+if [ $# -eq 1 ]
+  then
+    APP_DIR=$1
+else
+    APP_DIR=$(dirname $0)/..
+    echo "Application directory is not specified. Using default: '$APP_DIR'!"
+fi
+
 # Installing dependencies
-apt-get -y install build-essential libyaml-dev libfftw3-dev libavcodec-dev \
-    libavformat-dev python-dev python-numpy-dev python-numpy git \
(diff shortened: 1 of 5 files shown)


admin/install_hl_extractor.sh
```diff
@@ -1,9 +1,17 @@
...
+else
+    APP_DIR=$(dirname $0)/..
+    echo "Application directory is not specified. Using default: '$APP_DIR'!"
+fi
+
 # Installing dependencies
-apt-get -y install build-essential libyaml-dev libfftw3-dev libavcodec-dev \
-    libavformat-dev python-dev python-numpy-dev python-numpy git \
-    libsamplerate0-dev libtag1-dev libqt4-dev swig pkg-config
+apt-get -y --force-yes install build-essential libyaml-dev libfftw3-dev \
```
alastair:
```
       --force-yes
           Force yes; This is a dangerous option that will cause apt to
           continue without prompting if it is doing something potentially
           harmful. It should not be used except in very special situations.
```
Why do we need force yes?

gentlecat: I had issues with signatures on some of these libraries when I tried to install them today. Let me try again and see if it works without `--force-yes`.

gentlecat: Alright, seems to be working.

alastair: ok, fine. If there's a signature error then the process should have included `apt-key add` to load signatures :)

alastair:
fine. :ship:

PR #99, opened 2015-06-06
Updated dataset viewer

gentlecat:
I made it similar to the editor, which is also based on [React](https://facebook.github.io/react/).

*Live at http://test.acousticbrainz.org/ ([example](http://test.acousticbrainz.org/datasets/53ac14a7-c870-4e61-99d6-50336bdf1f27)).*


templates/datasets/viewer.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'base.html' %}
```
alastair: why did you rename this? `view` sounds like a good name

static/js/dataset-editor.js
```diff
@@ -281,7 +282,7 @@ var ClassList = React.createClass({
                 <h4>Classes</h4>
```
alastair: Is this normal react style to add layout directly to code - or can it be extracted out into template files?

templates/datasets/viewer.html
gentlecat: Just wanted naming of templates to be consistent. We have *delete*, *editor*, and *view*. Perhaps I should rename *editor* instead. :confused:

static/js/dataset-editor.js
gentlecat: React is meant to be used for UI, so I'd say yes. Take a look at https://facebook.github.io/react/docs/thinking-in-react.html, for example.

PR #101, opened 2015-06-10
Favicons

gentlecat:
:sparkles: :volcano: :sparkles:


mayhem:
Yay for favicons! \ø/ However, I think we should not be using the full size logo for the the larger favicons. Let's use the medium sized ones that have a little bit less detail in them.

gentlecat:
Oh, right. Picked wrong one for some reason.

gentlecat:
@mayhem, looks better?

templates/base.html
```diff
@@ -8,7 +8,12 @@
     <title>{% block title %}AcousticBrainz{% endblock %}</title>
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link rel="icon" href="{{ url_for('static', filename='img/favicon.png') }}" type="image/x-icon" />
+
+    <link rel="icon" sizes="16x16" href="{{ url_for('static', filename='img/favicon-16.png') }}" type="image/png" />
```
alastair: do we really need so many favicons? what are all of these used for? homscreen icons on an ipad? all my browser windows have 16x16 icons :)

mayhem: You actually use devices that uses favicons? Wow, so modern. :)

more seriously, the list of actual favicons that should be supported to make everything in the world happy is totally redonkulous. I think this is a fair balance.

gentlecat: They are used on high-res screens, [homescreen stuff](https://developer.chrome.com/multidevice/android/installtohomescreen#supporting) on mobile devices, etc.

gentlecat:
Is this OK to merge?

mayhem:
Sorry, didn't see the update. Yep! :+1:

PR #102, opened 2015-06-12
Added back datasets.view_json view

gentlecat:
It is used in the dataset viewer. I thought it was unused and removed it in 6ca97ec07806b8f186768c6fe9ed8311a3e223b6. :bomb:

### DIFF
--- a/acousticbrainz/views/datasets.py
+++ b/acousticbrainz/views/datasets.py
@@ -21,6 +21,14 @@ def view(id):
     )
 
 
+@datasets_bp.route("/<uuid:id>/json")
+def view_json(id):
+    ds = dataset.get(id)
+    if not ds:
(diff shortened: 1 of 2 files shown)


acousticbrainz/views/datasets.py
```diff
@@ -21,6 +21,14 @@ def view(id):
     )
 
 
+@datasets_bp.route("/<uuid:id>/json")
+def view_json(id):
+    ds = dataset.get(id)
```
alastair: can any user get any other user's dataset? did we talk about permissions?

gentlecat: We didn't implement privacy features for datasets yet.

alastair: OK. Fine for merge now then but this should be quite high in the list of
next things

PR #103, opened 2015-06-12
Using thousands separator in numbers in stats block on homepage

gentlecat:
Should make it easier to read. :newspaper:

### DIFF
--- a/acousticbrainz/views/index.py
+++ b/acousticbrainz/views/index.py
@@ -7,6 +7,8 @@ index_bp = Blueprint('index', __name__)
 @index_bp.route("/")
 def index():
     stats, last_collected = get_stats()
+    for key in stats:
+        stats[key] = "{:,}".format(stats[key])
     return render_template("index/index.html", stats=stats, last_collected=last_collected,
                            last_submitted_data=get_last_submitted_tracks())
 


acousticbrainz/views/index.py
```diff
@@ -7,6 +7,8 @@
 @index_bp.route("/")
 def index():
     stats, last_collected = get_stats()
+    for key in stats:
+        stats[key] = "{:,}".format(stats[key])
```
alastair: Is this the standard jinja way to format this? e.g. django has localisation in the templating itself, https://docs.djangoproject.com/en/1.8/topics/i18n/formatting/, which lets us use different separators depending on selected language

gentlecat: It's https://docs.python.org/2/library/string.html#format-specification-mini-language. In CritiqueBrainz we use something similar to Django stuff: https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/frontend/views.py#L27-L28, but we have support for localization there and additional libraries that provide this functionality.

alastair: OK,, no problem. Let's leave it as-is until we add i18n.

:ship: it

PR #104, opened 2015-06-12
Implemented privacy features for datasets

gentlecat:
*Live at http://test.acousticbrainz.org/.*

:see_no_evil:


acousticbrainz/data/dataset.py
```diff
@@ -129,7 +131,7 @@ def get(id):
     try:
         connection = psycopg2.connect(current_app.config["PG_CONNECT"])
         cursor = connection.cursor()
```
alastair: Is it worth using a dictcursor here so that we don't need to use `row[2]` later?

alastair:
Looks good

acousticbrainz/data/dataset.py
gentlecat: Huh, I didn't even know something like that existed. I will try to integrate it there.

PR #108, opened 2015-06-23
Database-related improvements

gentlecat:


acousticbrainz/data/__init__.py
```diff
@@ -1,192 +1,14 @@
...
-from acousticbrainz.utils import sanity_check_data, clean_metadata, interpret_high_level
-from acousticbrainz.data.exceptions import NoDataFoundException
-from werkzeug.exceptions import BadRequest, ServiceUnavailable, InternalServerError, NotFound
-from flask import current_app
-from hashlib import sha256
-import json
-import time
 
+# Be careful when importing this before init_connection function is called!
+connection = None
```
gentlecat: @alastair, any idea if there's a way to avoid importing `connection` in every function it is used (like I've done it now)?  There must be a better way. Doesn't seem right.

gentlecat:
There's a module called `data` in package `acousticbrainz.data`. Same names might be confusing there, but I'm not sure what to call it. Any suggestions?

acousticbrainz/data/__init__.py
gentlecat: *http://chatlogs.musicbrainz.org/musicbrainz-devel/2015/2015-06/2015-06-24.html#T20-22-07-694711*

acousticbrainz/data/__init__.py
```diff
@@ -1,192 +1,39 @@
...
-
-    return InternalServerError("whoops!")
+        logging.error("Failed to initialize database connection: %s" % str(e))
 
 
 def run_sql_script(sql_file_path):
-    cursor = psycopg2.connect(current_app.config['PG_CONNECT']).cursor()
-    with open(sql_file_path) as sql:
-        cursor.execute(sql.read())
+    global _connection
```
alastair: You don't need global when you just read a variable - it's only needed when assigning

gentlecat: Right, thanks!

acousticbrainz/data/dataset.py
```diff
@@ -141,74 +136,47 @@ def get(id):
...
-        return None
+    with data.create_cursor() as cursor:
+        cursor.execute(
+            "SELECT id, name, description, author, created, public "
+            "FROM dataset "
+            "WHERE id = %s",
+            (str(id),)
+        )
+        if cursor.rowcount > 0:
+            row = dict(cursor.fetchone())
```
alastair: :)

acousticbrainz/data/dump.py
```diff
@@ -333,68 +330,67 @@ def dump_highlevel_json(location, incremental=False, dump_id=None):
...
+            # Need to count how many duplicate MBIDs are there before start_time
+            if start_time:
+                cursor.execute("""
+                    SELECT mbid, count(id)
+                    FROM highlevel
+                    WHERE submitted <= %s
+                    GROUP BY mbid
+                    """, (start_time,))
+                counts = cursor.fetchall()
+                for count in counts:
```
alastair: `for mbid, count in counts`

admin/updates/update-6.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+ALTER TABLE class RENAME TO dataset_class;
```
alastair: i like this better, good choice

acousticbrainz/data/dump.py
gentlecat: :+1: a961f60a4042dd8f00f2dd8e8fb681d5ba4f329f

PR #109, opened 2015-06-29
Project structure improvements

gentlecat:
Okay, this one is pretty big, but most changes are just moving packages around and changing imports. Functionality is the same.

The main thing I try to do there is clearly separate different parts of the server (by modifying package structure and dependencies between different packages). This should simplify navigation, maintenance, and make it easier to understand for people unfamiliar with the project, which is always nice. I think this will also help split up project into smaller modules if we need to do that in future.

Most changes are related to separating parts from the web server (Flask and its extensions) and making them independent. The biggest one is `data` package, which is used for database access, caching, data dump creation and importing. Management scripts now use [Click](http://click.pocoo.org/) instead of [Flask-Script](https://flask-script.readthedocs.org/). Configuration file is moved out of web server since all scripts use it.

All that should help with upcoming dataset evaluator implementation.

*It might be easier to just browse this branch from https://github.com/metabrainz/acousticbrainz-server/tree/structure.*

:palm_tree: :cat2:


gentlecat:
Live at http://test.acousticbrainz.org/.

alastair:
:ship: it

PR #110, opened 2015-07-08
Dataset evaluation

gentlecat:
Adds another Python script that does dataset evaluation ([classifier training](https://github.com/MTG/essentia/blob/master/FAQ.md#training-and-running-classifier-models-in-gaia)) using [Gaia](https://github.com/MTG/gaia) library.


db/test/test_dataset_eval.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertIsNotNone(job)
+        self.assertEqual(job["status"], dataset_eval.STATUS_PENDING)
+
+    def test_get_job(self):
+        job_id = dataset_eval._create_job(self.test_dataset_id)
+        job = dataset_eval.get_job(job_id)
+        self.assertIsNotNone(job)
+        self.assertEqual(type(job), dict)
+
+        self.assertIsNone(dataset_eval.get_job("123e4567-e89b-12d3-a456-426655440000"))
```
alastair: Did you know that this isn't a valid uuid4?
https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29
(Don't worry, you don't need to change it)

db/dataset_eval.py
```diff
@@ -0,0 +1,129 @@
...
+        if cursor.fetchone()[0] > 0:
+            raise JobExistsException
+    validate_dataset(db.dataset.get(dataset_id))
+    return _create_job(dataset_id)
+
+
+def validate_dataset(dataset):
+    """Validate dataset by making sure that it matches JSON Schema for complete
+    datasets (JSON_SCHEMA_COMPLETE) and checking if all recordings referenced
+    in classes have low-level information in the database.
```
alastair: For now, this exception is OK. But could we do something more automatically? Perhaps we just ignore mbids with no match in AB. But maybe we need a flag for the user to say it's OK.

alastair: On second thoughts, let's leave it as-is for now. There's lots of interesting things we can do with the data - we'll make changes to this selection process next.

db/test/test_dataset_eval.py
gentlecat: I updated it anyway.

dataset_eval/evaluate.py
```diff
@@ -0,0 +1,129 @@
...
+
+    finally:
+        shutil.rmtree(temp_dir)  # Cleanup
+
+
+def create_groundtruth(dataset):
+    groundtruth = {
+        "type": "testing",      # FIXME: What's that?
+        "version": 1.0,         # FIXME: Is this important? Documentation?
+        "className": "hello",   # FIXME: ???
```
gentlecat: @alastair, any suggestions on what to do with these? Not sure if they even matter.

gentlecat: http://chatlogs.musicbrainz.org/musicbrainz-devel/2015/2015-07/2015-07-24.html#T12-38-43-626576

db/dataset_eval.py
gentlecat: Right, I think we should make basic functionality work first.

gentlecat:
I still want to make some changes to UI for this, but basic functionality should be ready (with a few small problems like catching exceptions from Gaia). Though, I'm sure we'll find plenty of bugs. :beetle:

Would be awesome if someone could help me test this and make sure it's working correctly.

:rainbow:

gentlecat:
Got a couple of questions:
1. Do we need to allow evaluation for datasets that were evaluated and haven't changed since then?
2. Should we allow to evaluate private datasets? Make results of each evaluation public?

gentlecat:
See http://chatlogs.musicbrainz.org/musicbrainz-devel/2015/2015-07/2015-07-25.html#T17-08-07-920044.

dataset_eval/gaia_wrapper.py
```diff
@@ -0,0 +1,39 @@
...
+    results = ClassificationResults()
+    results.readResults(project["resultsDirectory"])
+    best_accuracy, best_result_file, best_params = results.best(1, None)[0]
+
+    cm = ConfusionMatrix()
+    cm.load(best_result_file)
+    return {
+        "parameters": best_params,
+        "accuracy": best_accuracy,
+        "confusion_matrix": dict((k, dict(v)) for k, v in cm.matrix.items()),
```
gentlecat: Is this a good way to store it for rendering tables with results?

alastair: For now I don't think we need to store all of the song ids in this matrix. Just store the number of items in each class

gentlecat: Done in adaf2a5fa54e7fc850c18fe240bc8f199458a849. :rainbow:

gentlecat:
Here's how results table looks like:

alastair:
so pretty

webserver/views/datasets.py
```diff
@@ -30,12 +29,43 @@ def view(id):
     )
 
 
+@datasets_bp.route("/<uuid:dataset_id>/evaluation")
+def view_latest_job(dataset_id):
+    ds = get_dataset(dataset_id)
+    jobs = db.dataset_eval.get_jobs_for_dataset(ds["id"])
+    if not jobs:
+        raise NotFound("Can't find any evaluation jobs for the specified dataset.")
```
alastair: We shouldn't show a 404 here, the page doesn't look like the rest of the evaluation interface. Instead, do a 200, show the evaluation page interface, but have a message that says that this dataset hasn't been submitted for evaluation yet.

db/dataset_eval.py
```diff
@@ -0,0 +1,152 @@
...
+        ID of the newly created evaluation job.
+    """
+    with create_cursor() as cursor:
+        cursor.execute(
+            "SELECT count(*) FROM dataset_eval_jobs WHERE dataset_id = %s AND status IN %s",
+            (dataset_id, (STATUS_PENDING, STATUS_RUNNING))
+        )
+        if cursor.fetchone()[0] > 0:
+            raise JobExistsException
+    validate_dataset(db.dataset.get(dataset_id))
```
alastair: If this fails we should report the error to the user so that they can fix it

webserver/templates/datasets/eval-job.html
```diff
@@ -0,0 +1,81 @@
...
+
+{% block title %}Dataset "{{ dataset['name'] }}" by {{ author.musicbrainz_id }} - AcousticBrainz{% endblock %}
+
+{% block content %}
+  <div id="evaluation-results">
+
+    {% from 'datasets/macros.html' import print_header %}
+    {{ print_header(dataset, author, current_user, 'evaluation') }}
+
+    {% if job['status'] == 'pending' %}
```
alastair: We can actually say "pending", maybe with a "time since added" option. In the future we could have an estimated ETA, but I don't think that's important. How about "n datasets in queue"?

webserver/templates/datasets/eval-job.html
```diff
@@ -0,0 +1,81 @@
...
+      </div>
+
+    {% elif job['status'] == 'running' %}
+      <div id="status" class="alert alert-info">
+        This dataset is being evaluated right now. Results should be available soon.
+      </div>
+
+    {% elif job['status'] == 'failed' %}
+      <div id="status" class="alert alert-danger">
+        Evaluation of this dataset has failed!
```
alastair: Will we show the reason for a failure?

alastair: After a dataset has failed, it'd be nice if the user can re-submit itto the queue. Perhaps we need a way in the future to check that they don't continually submit the same failing dataset

dataset_eval/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+
+def evaluate_dataset(eval_job):
+    db.dataset_eval.set_job_status(eval_job["id"], db.dataset_eval.STATUS_RUNNING)
+    temp_dir = tempfile.mkdtemp()
+
+    try:
+        dataset = db.dataset.get(eval_job["dataset_id"])
+
+        filelist_path = os.path.join(temp_dir, "filelist.yaml")
+        filelist = dump_lowlevel_data(extract_recordings(dataset), os.path.join(temp_dir, "data"))
```
alastair: can we have some logging here so that when we run this we can see what's happening?

dataset_eval/evaluate.py
```diff
@@ -0,0 +1,142 @@
...
+
+        results = gaia_wrapper.train_model(
+            groundtruth_file=groundtruth_path,
+            filelist_file=filelist_path,
+            project_dir=temp_dir,
+        )
+        db.dataset_eval.set_job_result(eval_job["id"], json.dumps(results))
+        db.dataset_eval.set_job_status(eval_job["id"], db.dataset_eval.STATUS_DONE)
+
+    # TODO(roman): Also need to catch exceptions from Gaia.
```
alastair: Gaia exceptions are pretty important. Can we hack this for now? Some ideas:

 - get the stacktrace and look for "gaia2"
 - find a list of common exceptions and check `msg`

dataset_eval/gaia_wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+        groundtruth_file=groundtruth_file,
+        filelist_file=filelist_file,
+        datasets_dir=os.path.join(project_dir, "datasets"),
+        results_dir=os.path.join(project_dir, "results"),
+    )
+    run_tests(project_file)
+    return select_best_model(project_file)
+
+
+def select_best_model(project_file_path):
```
alastair: Docs

webserver/templates/datasets/macros.html
```diff
@@ -0,0 +1,53 @@
...
+    </a>
+
+    <p id="dataset-description">{{ dataset['description'] }}</p>
+
+    <p id="dataset-info">
+      <strong>Author:</strong>
+      <a href="{{ url_for('user.profile', musicbrainz_id=author.musicbrainz_id) }}">
+        {{ author.musicbrainz_id }}
+      </a>
+      <br/><strong>Created:</strong> {{ dataset['created'] }}
```
alastair: Can you make this date more compact? We don't need seconds and miliseconds

alastair:
This is cool! Go ahead and merge it. I'll continue editing the requirements document to add some suggestions for workflow, which we can implement next.

gentlecat:
I'll fix issues that you pointed out and then merge.

dataset_eval/gaia_wrapper.py
gentlecat: What should be documented there?

webserver/templates/datasets/macros.html
gentlecat: Yeah, I think I can.

dataset_eval/evaluate.py
gentlecat: Yes.

db/dataset_eval.py
gentlecat: http://tickets.musicbrainz.org/browse/AB-61

webserver/templates/datasets/eval-job.html
gentlecat: We should get failure reasons with a fix for http://tickets.musicbrainz.org/browse/AB-61. I'm not sure exactly how to check if same failing dataset is being submitted. Maybe compare update timestamps?

webserver/views/datasets.py
gentlecat: http://tickets.musicbrainz.org/browse/AB-62

webserver/templates/datasets/eval-job.html
gentlecat: Not sure how we can compute ETA. Queue details are coming: http://tickets.musicbrainz.org/browse/AB-63.

dataset_eval/evaluate.py
gentlecat: http://tickets.musicbrainz.org/browse/AB-64

dataset_eval/gaia_wrapper.py
alastair: a description of what this method does, what its input is, what its output/return value is

There should be one for train_model, too

webserver/templates/datasets/eval-job.html
alastair: That's a good idea. If it hasn't been updated since it failed, don't let them change it.

However, this doesn't solve the issue of us having a gaia error which is something we need to fix. Do you have any ideas on a way to report errors to us so that we can fix them, and then let us re-queue someone's dataset?  Perhaps we could use sentry.

webserver/templates/datasets/eval-job.html
alastair: For ETA we could have some sort of estimate of "time to process x files" (based on previous runs), and scale the value, but it's not important now.

webserver/templates/datasets/eval-job.html
gentlecat: Yeah, for our error monitoring we can just add calls to `logging` module and handlers (including Sentry) will record that depending on levels set there: https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/loggers.py#L7-L12.

On the website we still need a user-friendly way to display these errors. Maybe we can say "Error occurred during evaluation" and provide a way to see stack trace?

dataset_eval/gaia_wrapper.py
gentlecat: Yeah, I'll add some info about that.

webserver/templates/datasets/eval-job.html
alastair: A stack trace is never user-friendly ;)  Ideally we should understand all of the errors, and be able to describe them nicely (with a suggested fix). We can always send the stack trace to us so that we can debug it further.

It might be nice to have an admin interface which lets us see, at least, all pending, running, and finished jobs, with links to them, and the ability to change the state. We'll need admin users to be able to do this too. Make an issue for it, but make it quite low priority. For now we need an easy way to re-queue a stuck process. Maybe we could just have a "resubmit" button for people to do it themselves, but in the future we remove it. What do you think?

dataset_eval/gaia_wrapper.py
gentlecat: Done. See #120.

webserver/templates/datasets/eval-job.html
gentlecat: Can you create a ticket for this please?

webserver/templates/datasets/eval-job.html
gentlecat: Yes, it's not user-friendly. But considering that Gaia doesn't have proper exceptions, there's not much else we can do there I think. String parsing seems very hacky and fragile.

I plan to create public pages with information about current evaluation jobs, that might be helpful. Perhaps we can show some additional functions for admins there.

Ticket for admin interface: http://tickets.musicbrainz.org/browse/AB-66. Feel free to add details.

alastair: parsing `e.msg` should be quite stable, as long as we know that exceptions from gaia are raised with a known string. For now we don't have anything better, so I don't have a problem with doing this.
The public page would be a good start for admin control as well.

PR #112, opened 2015-07-19
Updated summary data display

gentlecat:
It is now possible to view different submissions. I also did some minor refactoring in database access code and data display.

*This is related to [AB-13](AB-13).* :tv:


webserver/templates/data/summary.html
```diff
@@ -0,0 +1,107 @@
...
+    </table>
+
+    {% if data.highlevel %}
+      <h3>High-level information</h3>
+
+      {% macro print_row(row) %}
+        {# Prints <name, value, probability> table row. #}
+          {% set probability = row[2]|float * 100 %}
+          <tr class="{{ 'success' if probability > 75 else (
+                        'warning' if probability > 25 else
```
gentlecat: Let me know if these should be adjusted.

gentlecat:
Example: http://ab.roman.zone/770cc467-8dde-4d22-bc4c-a42f91e7515e?n=1

alastair:
Minor refactoring? why did 159 files change?!!?

gentlecat:
159 files changed because I removed old unused Bootstrap files and compiled Less.

webserver/templates/data/summary.html
alastair: We definitely don't need this class on the entire row. Just because The Machine has a confidence of >75% of a value doesn't mean we have to say "LOOK, WE HAVE A SAD SONG!". I wonder if we could highlight just the probability number, but I'm tending towards not highlighting anything. In the end we should stop relying on this value and only show data which we're highly confident of.

gentlecat: Alright, I'll remove that. Just thought that it would be nice to have more visual indication.

alastair:
mergeit

PR #113, opened 2015-07-25
Use absolute path when importing config file

gentlecat:


alastair:
Fine, go for it.
My preference would be to not have to hack `sys.path`, but I'm not sure if there's a better way.

gentlecat:
Yeah, not sure what would be a better way to share config file.

PR #115, opened 2015-07-29
Added more logging handlers

gentlecat:
* Email (SMTPHandler)
* [Sentry](https://github.com/getsentry/sentry) ([Raven](https://github.com/getsentry/raven-python))

:memo:

### DIFF
--- a/config.py.sample
+++ b/config.py.sample
@@ -13,7 +13,6 @@ PG_CONNECT_TEST = "dbname=ab_test user=ab_test password= host=localhost port=543
 # Set to True if Less should be compiled in browser. Set to False if styling is pre-compiled.
 COMPILE_LESS = True
 
-LOG_FILE = "./acousticbrainz.log"
 
 MUSICBRAINZ_USERAGENT = "acousticbrainz-server"
 MUSICBRAINZ_HOSTNAME = None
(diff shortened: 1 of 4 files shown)


gentlecat:
We use Sentry in MetaBrainz website and CritiqueBrainz already. It works quite well, makes keeping track of problems easier.

gentlecat:
Also it doesn't look like we do any logging in other parts of the server. I can look into that as well.

alastair:
What log systems do you suggest we set up in production? All of them? Do we have a metabrainz sentry server? Who will get emails?

gentlecat:
*SMTPHandler* sends email for every single exception that occurs. Sentry notifications are [less reduandant](https://docs.getsentry.com/on-premise/notifications/). I'm running it on my own server since nobody wanted to deploy another application on MusicBrainz servers.

alastair:
Yeah, i've used sentry a bit. It's nice. Let's use your instance for now, but I would feel happier moving it somewhere else in the future, especially since we have 3 (?) products now using it.

alastair:
go for it, and set me up a sentry login, please :)

gentlecat:
Yeah, I agree with you. Should poke @mayhem about that once again.

gentlecat:
@alastair, you should receive an invitation via email (might be in spam).

PR #117, opened 2015-07-31
Replaced our function for directory creation with one from the standard library

gentlecat:


db/dump.py
```diff
@@ -73,7 +72,8 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
     Returns:
         Path to created dump.
     """
-    create_path(location)
+    if not os.path.exists(location):
```
alastair: This doesn't actually check if the file at location is a directory or a file (you should use `isdir`)
However, the style in `create_path` is very pythonic - you usually do something and catch an exception rather than first test if you can do it: http://stackoverflow.com/questions/6092992/why-is-it-easier-to-ask-forgiveness-than-permission-in-python-but-not-in-java

gentlecat: Hm, interesting. If it doesn't actually improve anything, let's keep it as it was before. I just didn't know that `os.makedirs

PR #118, opened 2015-08-02
Moved schema version definition into db package

gentlecat:


db/__init__.py
```diff
@@ -6,6 +6,10 @@
 
 import logging
 
+
+# This value must be incremented after schema changes on replicated tables!
+__version__ = 2
```
alastair: Should we call this `version`, or `schema`?

gentlecat: `SCHEMA_VERSION`?

alastair: ok!

gentlecat: Done.

PR #119, opened 2015-08-03
Updated dependencies

gentlecat:


alastair:
Sure. If it works, you don't need to ask about commits like this.

gentlecat:
I think it's always better to have someone check your work, even if it's pretty basic. Mistakes happen, maybe something prevents us from upgrading, etc.

PR #121, opened 2015-08-05
Replaced some print calls with functions from logging module

gentlecat:


alastair:
cool, looks good. Can we try and reuse the logging infrastructure for the webserver too? So that we can send messages to the same place easily.

gentlecat:
Yeah, I want to make logging work for all packages we have there.

PR #123, opened 2015-08-05
Check if high-level data has been computed correctly before trying to interpret it

gentlecat:
*Fixes http://sentry.roman.zone/sentry/ab-production/group/29/.*

### DIFF
--- a/webserver/views/data.py
+++ b/webserver/views/data.py
@@ -93,7 +93,7 @@ def summary(mbid):
     except db.exceptions.NoDataFoundException:
         summary_data = {}
 
-    if "highlevel" in summary_data:
+    if "highlevel" in summary_data and summary_data["highlevel"]:
         genres, moods, other = _interpret_high_level(summary_data["highlevel"])
         summary_data["highlevel"] = {
             "genres": genres,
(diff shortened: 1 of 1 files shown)


webserver/views/data.py
```diff
@@ -93,7 +93,7 @@ def summary(mbid):
     except db.exceptions.NoDataFoundException:
         summary_data = {}
 
-    if "highlevel" in summary_data:
+    if "highlevel" in summary_data and summary_data["highlevel"]:
```
alastair: what can `summary_data["highlevel"]` be which is bad?
How about `if summary_data.get("highlevel", {})`?

gentlecat: It can be an empty dictionary. Probably from high-level extractor that didn't work correctly.

alastair: Ah, I see now. This is not the case from the exception catch on line 94.

I'd prefer the `.get()` syntax rather than the double if statement, since an empty dictionary `{}` will evaluate to false.

gentlecat: Okay, I get your idea now. Took me way too long to understand it since I'm not used to `.get()`. Updated in 5e017461a5a6a4244d77185721e0dcf46bca2fbc.

PR #122, opened 2015-08-05
Checking if there are any releases associated with recording in _get_recording_info function

gentlecat:
*Fixes http://sentry.roman.zone/sentry/ab-production/group/30/ and http://tickets.musicbrainz.org/browse/AB-49.*

PR #124, opened 2015-08-08
Check if "musicbrainz_albumid" is defined in tags

gentlecat:
*Fixes http://sentry.roman.zone/sentry/ab-production/group/32/.*

### DIFF
--- a/webserver/views/data.py
+++ b/webserver/views/data.py
@@ -163,15 +163,13 @@ def _get_recording_info(mbid, metadata):
             info['length'] = time.strftime("%M:%S", time.gmtime(float(good_metadata['length']) / 1000))
 
     elif metadata:
-        info['title'] = metadata['tags']['title'][0]
         info['length'] = metadata['audio_properties']['length_formatted']
-        info['artist_id'] = metadata['tags']['musicbrainz_artistid'][0]
-        info['artist'] = metadata['tags']['artist'][0]
-        info['release_id'] = metadata['tags']['musicbrainz_albumid'][0] \
-            if 'musicbrainz_albumid' in metadata['tags'] else None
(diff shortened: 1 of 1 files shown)


Freso:
Also http://tickets.musicbrainz.org/browse/AB-49

gentlecat:
@Freso, this is going to be fixed by #122.

webserver/views/data.py
```diff
@@ -167,10 +167,11 @@ def _get_recording_info(mbid, metadata):
...
         info['artist_id'] = metadata['tags']['musicbrainz_artistid'][0]
         info['artist'] = metadata['tags']['artist'][0]
-        info['release_id'] = metadata['tags']['musicbrainz_albumid'][0]
+        info['release_id'] = metadata['tags']['musicbrainz_albumid'][0] \
+            if 'musicbrainz_albumid' in metadata['tags'] else None
         info['release'] = metadata['tags']['album'][0]
-        info['track_id'] = metadata['tags']['musicbrainz_releasetrackid'][0] if \
-            'musicbrainz_releasetrackid' in metadata['tags'] else None
+        info['track_id'] = metadata['tags']['musicbrainz_releasetrackid'][0] \
+            if 'musicbrainz_releasetrackid' in metadata['tags'] else None
```
alastair: how about `info['track_id'] = metadata['tags'].get('musicbrainz_releasetrackid', [None])[0]` ?

Alternatively, I'd like to see a helper function which is used for all lines which checks if the field exists (we don't require any tag except recording id to be present) and then does the `[0]`

gentlecat: I like your first idea. Done in 88083614b487c6fecb48116601d9ab893084f7dc.

PR #125, opened 2015-08-08
List of all evaluation jobs for a dataset

gentlecat:
*Fixes [AB-62](http://tickets.musicbrainz.org/browse/AB-62).*


gentlecat:
Accidentally merged it. Reverted and created new pull request with the same contents: #128. Sorry about that.

PR #126, opened 2015-08-11
Save messages from exceptions in the dataset evaluator

gentlecat:
I didn't integrate that with the UI yet because #125 changes related part of the interface anyway. I'll add message display after that pull request is merged.

*This is related to [AB-61](http://tickets.musicbrainz.org/browse/AB-61).*

### DIFF
--- a/dataset_eval/evaluate.py
+++ b/dataset_eval/evaluate.py
@@ -71,8 +71,11 @@ def evaluate_dataset(eval_job):
     # TODO(roman): Also need to catch exceptions from Gaia.
     except db.exceptions.DatabaseException as e:
         logging.info("Evaluation job %s has failed!" % eval_job["id"])
-        db.dataset_eval.set_job_status(eval_job["id"], db.dataset_eval.STATUS_FAILED,
-                                       status_msg=str(e))
+        db.dataset_eval.set_job_status(
+            job_id=eval_job["id"],
+            status=db.dataset_eval.STATUS_FAILED,
+            status_msg=str(e),
+        )
         logging.info(e)
 
     finally:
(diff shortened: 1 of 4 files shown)


dataset_eval/evaluate.py
```diff
@@ -71,7 +71,8 @@ def evaluate_dataset(eval_job):
     # TODO(roman): Also need to catch exceptions from Gaia.
     except db.exceptions.DatabaseException as e:
         logging.info("Evaluation job %s has failed!" % eval_job["id"])
-        db.dataset_eval.set_job_status(eval_job["id"], db.dataset_eval.STATUS_FAILED)
+        db.dataset_eval.set_job_status(eval_job["id"], db.dataset_eval.STATUS_FAILED,
+                                       status_msg=str(e))
```
alastair: Why do you name `status_msg` here but not any of the other parameters?

dataset_eval/evaluate.py
```diff
@@ -71,7 +71,8 @@ def evaluate_dataset(eval_job):
...
         # TODO(roman): Maybe log this exception? Would also be nice to provide
```
alastair: no longer a todo!

db/dataset_eval.py
```diff
@@ -113,7 +113,7 @@ def set_job_result(job_id, result):
     commit()
 
 
-def set_job_status(job_id, status):
+def set_job_status(job_id, status, status_msg=None):
```
alastair: Is `status_msg` optional? Could you add some documentation?

dataset_eval/evaluate.py
gentlecat: Yep. Removed in 3eda350ddc2c8ef822a819175d7cc6528de744ad. :boom:

dataset_eval/evaluate.py
gentlecat: I was trying to make it more clear what this string is. Updated in 0998697ab72b001d3d85de742d0c44e58c2852fe.

db/dataset_eval.py
gentlecat: Yep, it's optional. I added docstring for this function in c7b5eedc69c2ffcb7bcf6d26608bf2f38dff220f. :scroll:

PR #128, opened 2015-08-13
List of all evaluation jobs for a dataset

gentlecat:
*Fixes [AB-62](http://tickets.musicbrainz.org/browse/AB-62).*


alastair:
This is good.
We talked about having an "admin" interface to also be able to see all current jobs. Did you want to make that part of this pull, or a separate one?

alastair:
Watch the back button. There are a few places where it doesn't work as expected. Let me know if you want specific instances

alastair:
ah, the actual cause of the http500 error is
```
ProgrammingError: column "status_msg" does not exist
LINE 1: SELECT id, dataset_id, status, status_msg, result, created, ...
```
which is fine, except it's unusal that it's being eclipsed by the transaction error. I think we need to do 2 things:

* make sure that the http500 template doesn't do any database calls, so that we can report the correct error 
* Don't render the http500 html page for a json query endpoint

gentlecat:
No, I didn't want to make admin interface a part of this pull request. This just fixes AB-62. There's a separate ticket for admin stuff: http://tickets.musicbrainz.org/browse/AB-66.

alastair:
OK. make a ticket for the http500 stuff and this is OK to merge.

gentlecat:
Here's the ticket for error page stuff: http://tickets.musicbrainz.org/browse/AB-74.

PR #129, opened 2015-08-13
History file generation in dataset evaluation process

gentlecat:
:scroll: :honeybee: :honeybee: :honeybee:

### DIFF
--- a/dataset_eval/evaluate.py
+++ b/dataset_eval/evaluate.py
@@ -23,6 +23,7 @@ import os
 import re
 
 SLEEP_DURATION = 30  # number of seconds to wait between runs
+HISTORY_STORAGE_DIR = os.path.join(config.FILE_STORAGE_DIR, "history")
 
 
 def main():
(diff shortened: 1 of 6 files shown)


gentlecat:
Still need to permanently save that file on the server and provide a way to download it.

dataset_eval/evaluate.py
```diff
@@ -23,6 +23,7 @@
 import re
 
 SLEEP_DURATION = 30  # number of seconds to wait between runs
+HISTORY_STORAGE_DIR = os.path.join(config.FILE_STORAGE_DIR, "history")
```
alastair: does it make sense to put the existing high level files here too? what do you think?

alastair:
looks good :rocket:

dataset_eval/evaluate.py
gentlecat: What files?

alastair: I mean all of the ones that we currently use in the high level extractor currently.

gentlecat: *http://chatlogs.musicbrainz.org/musicbrainz-devel/2015/2015-08/2015-08-15.html#T12-14-36-810697* :memo:

PR #133, opened 2015-08-27
Fixed metadata fetching for summary page

gentlecat:


webserver/views/data.py
```diff
@@ -173,14 +173,15 @@ def _get_recording_info(mbid, metadata):
...
         if 'tracktotal' in metadata['tags']:
             info['track_number'] = '%s / %s' % (metadata['tags']['tracknumber'][0],
                                                 metadata['tags']['tracktotal'][0])
         else:
             info['track_number'] = metadata['tags']['tracknumber'][0]
+        return info
 
-    return info
+    else:
+        return None
```
alastair: I don't like returning None from something that normally returns a dict. Can we return `{}` instead? I guess the problem here is that `info` always has at least the mbid, so that we can't check it for emptiness when it gets returned?

gentlecat: Exactly.

gentlecat: Fixed. :ice_cream:

PR #134, opened 2015-09-05
Remove unneeded server.py

alastair:
AB-76 - You can configure uwsgi directly with the create_app
function in the configuration file


gentlecat:
:shipit:

PR #137, opened 2015-09-21
Use a wrapper to call sql scripts in admin

alastair:
Fixes http://tickets.musicbrainz.org/browse/AB-83


gentlecat:
:+1: :ship:

alastair:
We should put this in other brainzes. Do you want to do this?

gentlecat:
I can do that, but not today. Need to finish some MusicBrainz stuff.

alastair:
no problem

PR #138, opened 2015-09-24
add more MBID names for mp3 tags

19379:
on IRC (as marc2k3) earlier, i was talking about submitting MBIDs to listenbrainz, i was linked to the whitelist json file and i noticed many mp3 specific MBID names were missing so here they are.

i put them side by side just to illustrate the differences. it can always be changed.


Freso:
But [TagLib](https://taglib.github.io/) reading those ID3 tags in MP3 files should abstract them out to the tag names already listed.

19379:
well i'm using foobar2000 and it doesn't use taglib. all TXXX frames are accessed by name. i can always workaround this issue at my end by substituting the names before i submit.

19379:
since i've updated my foobar2000 script to handle the remapping, i might as well close this.

PR #139, opened 2015-10-05
[WIP] Sqlalchemy

alastair:
Work in progress, but getting it up now because it looks like it'll be big.

The main change is to move from psycopg2 to sqlalchemy (http://tickets.musicbrainz.org/browse/AB-89), but there are also some other things in here, including better transaction management (http://tickets.musicbrainz.org/browse/AB-85) and splitting up of data methods more to separate them into database access, and data validation/transformation.

Also adding tests to each module as I'm going along. At this point, most of db/data is done but no more.

Still to be done:
 - [x] highlevel data methods
 - [x] dataset db
 - [x] stats db
 - [x] user db
 - [x] dumps
 - [x] initdb
 - [x] move highlevel script to `db`
 - [x] move eval script to `db`
 - [x] verify transactions work properly for multi-table inserts


db/data.py
```diff
@@ -27,6 +27,7 @@
     ['tonal'],
 ]
 
+# TODO: Util methods should not be in the database package
```
gentlecat: `acousticbrainz.utils`?

db/dataset.py
```diff
@@ -103,33 +100,30 @@ def update(dataset_id, dictionary, author_id):
...
-            cls_id = cursor.fetchone()[0]
+            cls_id = result.fetchone()[0]
 
             for recording_mbid in cls["recordings"]:
-                cursor.execute("INSERT INTO dataset_class_member (class, mbid) VALUES (%s, %s)",
+                connection.execute("INSERT INTO dataset_class_member (class, mbid) VALUES (%s, %s)",
                                (cls_id, recording_mbid))
 
-    # If anything bad happens above, it should just rollback by default.
-    commit()
```
gentlecat: Is it commiting automatically now?

db/stats.py
```diff
@@ -1,4 +1,5 @@
-from db import create_cursor, commit, cache
+import db
+import db.cache
```
gentlecat: Maybe memcached wrapper should be out of `db` package as well? What do you think?

gentlecat: There's actually a flaw in this module. See https://bitbucket.org/metabrainz/musicbrainz-server/pull-requests/1644/:

> The cache namespace key is stored in redis as opposed to memcached so that it doesn't get removed by any LRU algorithm.

*(we don't do that)*

db/data.py
alastair: right, `acousticbrainz.something`. I'm thinking that we should have something like

acousticbrainz/manager/data.py - interface for doing stuff
acousticbrainz/db/data.py - lowlevel database, only called by the manager

db/dataset.py
alastair: yes, transactions should automatically commit when `with db.engine.connection()` finishes

db/test/test_data.py
```diff
@@ -1,20 +1,207 @@
...
         self.test_lowlevel_data_json = open(os.path.join(TEST_DATA_PATH, self.test_mbid + '.json')).read()
         self.test_lowlevel_data = json.loads(self.test_lowlevel_data_json)
 
-    def test_submit_low_level_data(self):
-        data.submit_low_level_data(self.test_mbid, self.test_lowlevel_data)
-        self.assertIsNotNone(data.load_low_level(self.test_mbid))
+
+    @mock.patch("db.data.sanity_check_data")
+    @mock.patch("db.data.write_low_level")
+    @mock.patch("db.data.clean_metadata")
```
gentlecat: What are these?

db/stats.py
alastair: OK, we should open a ticket

db/test/test_data.py
alastair: http://www.voidspace.org.uk/python/mock/patch.html

db/stats.py
gentlecat: http://tickets.musicbrainz.org/browse/AB-92 :ticket:

db/testing.py
```diff
@@ -28,36 +28,38 @@ def reset_db(self):
...
 
     def drop_types(self):
-        with create_cursor() as cursor:
-            cursor.execute('DROP TYPE IF EXISTS eval_job_status CASCADE;')
-        commit()
+        with db.engine.connect() as connection:
+            connection.execute('DROP TYPE IF EXISTS eval_job_status CASCADE;')
+
+    def data_filename(self, mbid):
+        """ Get the expected filename of a test datafile given its mbid """
```
gentlecat: Do you want me to complain about docstring formatting for the sake of consistency? I can ignore that if you don't want to worry about it. I mostly use https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Comments#Comments as a reference. IDE/pylint reminds me about this stuff most of the time, so I kind of got used to the style.

admin/sql/create_indexes.sql
```diff
@@ -7,4 +7,6 @@ CREATE UNIQUE INDEX data_sha256_ndx_lowlevel ON lowlevel (data_sha256);
 CREATE INDEX mbid_ndx_highlevel ON highlevel (mbid);
 CREATE INDEX build_sha1_ndx_highlevel ON lowlevel (build_sha1);
 
+CREATE INDEX lower_musicbrainz_id_ndx_user ON "user" (lower(musicbrainz_id));
```
alastair: I'm not sure if I should also make this a unique constraint - would it be possible to add the same user twice with different case?  MusicBrainz should never send us this, but it should be taken care of

gentlecat: That's actually one of the things I was worried about. Right now we are inserting from only one place, but it seems like a good idea to take care of this.

alastair:
I just tested high level calc with query logging on and they are wrapped in a transaction.

Low level insert and dataset editor queries are also wrapped with a transaction.  Ready to merge

PR #140, opened 2015-10-05
AB-90: Return a json error message for API endpoints

alastair:
Fixes AB-90: http://tickets.musicbrainz.org/browse/AB-90
We use a flask pattern from
http://flask.pocoo.org/docs/0.10/patterns/apierrors/
to raise a specific api error and return a json response in the
error handler.
We add two more routes to catch /anything/low-level and .../high-level
so that we can also return json for invalid uuids, not just missing
ones.

### DIFF
--- a/webserver/errors.py
+++ b/webserver/errors.py
@@ -3,8 +3,8 @@ import webserver.exceptions
 
 def init_error_handlers(app):
 
-    @app.errorhandler(webserver.exceptions.InvalidAPIUsage)
-    def handle_invalid_usage(error):
+    @app.errorhandler(webserver.exceptions.APIError)
+    def api_error(error):
         response = jsonify(error.to_dict())
         response.status_code = error.status_code
         return response
--- a/webserver/errors.py
+++ b/webserver/errors.py
@@ -1,7 +1,13 @@
-from flask import render_template
-
+from flask import render_template, jsonify
+import webserver.exceptions
 
 def init_error_handlers(app):
+
+    @app.errorhandler(webserver.exceptions.InvalidAPIUsage)
+    def handle_invalid_usage(error):
+        response = jsonify(error.to_dict())
+        response.status_code = error.status_code
+        return response
 
     @app.errorhandler(400)
     def bad_request(error):
--- a/webserver/exceptions.py
+++ b/webserver/exceptions.py
@@ -1,8 +1,7 @@
-class InvalidAPIUsage(Exception):
-    status_code = 400
+class APIError(Exception):
 
     def __init__(self, message, status_code, payload=None):
-        Exception.__init__(self)
+        super(APIError, self).__init__()
         self.message = message
         self.status_code = status_code
         self.payload = payload
@@ -11,3 +10,11 @@ class InvalidAPIUsage(Exception):
         rv = dict(self.payload or ())
         rv['message'] = self.message
         return rv
+
+class APINotFound(APIError):
+    def __init__(self, message, payload=None):
+        super(APINotFound, self).__init__(message, 404, payload)
+
+class APIBadRequest(APIError):
+    def __init__(self, message, payload=None):
+        super(APIBadRequest, self).__init__(message, 400, payload)
--- a/webserver/exceptions.py
+++ b/webserver/exceptions.py
@@ -1,0 +1,13 @@
+class InvalidAPIUsage(Exception):
+    status_code = 400
+
+    def __init__(self, message, status_code, payload=None):
+        Exception.__init__(self)
+        self.message = message
+        self.status_code = status_code
+        self.payload = payload
+
+    def to_dict(self):
+        rv = dict(self.payload or ())
+        rv['message'] = self.message
+        return rv
--- a/webserver/templates/navbar.html
+++ b/webserver/templates/navbar.html
@@ -28,7 +28,7 @@
       {# Not showing user stuff on error pages. If attempt to load user info fails there, page will not render. #}
       {% if not error %}
         <ul class="nav navbar-nav navbar-right">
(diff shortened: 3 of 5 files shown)


webserver/templates/navbar.html
```diff
@@ -28,7 +28,7 @@
       {# Not showing user stuff on error pages. If attempt to load user info fails there, page will not render. #}
       {% if not error %}
         <ul class="nav navbar-nav navbar-right">
-          {% if not current_user or current_user.is_anonymous() %}
+          {% if not current_user or current_user.is_anonymous %}
```
gentlecat: I'm guessing this is because `is_anonymous` and a bunch of other values have been converted into class properties in one of the latest versions of Flask-Login?

alastair: yeah, I upgraded flask login, so I guess that's it.

django's templating language will take a name and first try it as a property, then a dictionary key, then a function. does jinja do the same?

webserver/exceptions.py
```diff
@@ -0,0 +1,13 @@
+class InvalidAPIUsage(Exception):
+    status_code = 400
```
alastair: I changed status code from optional to required in the constructor (compared to the page I got the snippet from). I missed this. It can be removed

webserver/views/api.py
```diff
@@ -37,7 +38,13 @@ def get_low_level(mbid):
...
         return Response(load_low_level(mbid, offset), content_type='application/json')
     except NoDataFoundException:
-        raise NotFound
+        raise webserver.exceptions.InvalidAPIUsage("Not found", status_code=NotFound.code)
+
+@api_bp.route("/<string:notanmbid>/low-level", methods=["GET"])
+@crossdomain()
+def get_low_level_404(notanmbid):
+    """Specific low-level view for non-mbid url parts to return a json error message"""
+    raise webserver.exceptions.InvalidAPIUsage("Not found", status_code=NotFound.code)
```
gentlecat: Do we need these views to catch a case when non-UUID is being submitted to the API? I guess `<uuid:mbid>` check just raises regular exception.

That's kind of ugly. :confused: What if we remove all validation on the path level and do it manually?

alastair: The thing is, I don't want to get an html error message for the potential case of a non-uuid argument. 
I don't mind this setup, but I would be just as happy with a non-specific uuid matcher and do validation inside the method (return 404 if it's not a valid uuid)

webserver/views/api.py
```diff
@@ -65,10 +78,10 @@ def submit_low_level(mbid):
...
-        raise BadRequest("Cannot parse JSON document: %s" % e)
+        raise webserver.exceptions.InvalidAPIUsage("Cannot parse JSON document: %s" % e, status_code=BadRequest.code)
 
     try:
         submit_low_level_data(mbid, data)
     except BadDataException as e:
-        raise BadRequest(e)
-    return ""
+        raise webserver.exceptions.InvalidAPIUsage("%s" % e, status_code=BadRequest.code)
+    return jsonify({"status": "ok"})
```
gentlecat: :+1:

webserver/views/api.py
gentlecat: Right, that's what I thought. I think just having a regular checking within one view would be better.

webserver/views/api.py
alastair: though, the exception handler returns a key `message`. Perhaps we should be consistent.

webserver/exceptions.py
```diff
@@ -0,0 +1,13 @@
...
+    def __init__(self, message, status_code, payload=None):
+        Exception.__init__(self)
+        self.message = message
+        self.status_code = status_code
+        self.payload = payload
+
+    def to_dict(self):
+        rv = dict(self.payload or ())
+        rv['message'] = self.message
+        return rv
```
alastair: no newline. todo

webserver/templates/navbar.html
gentlecat: Looks like it's kind of similar to that. http://jinja.pocoo.org/docs/dev/templates/#variables

webserver/exceptions.py
```diff
@@ -0,0 +1,13 @@
+class InvalidAPIUsage(Exception):
```
gentlecat: Maybe it would be better to have more generic `APIException` and use a separate exception in cases where we have no data? Something like this:

```python
class APIException(Exception): pass
class BadRequest(APIException): pass
class NoData(APIException): pass
# etc...
```

And just set status associated status code in each instead of setting it in constructor every time. What do you think?

webserver/errors.py
```diff
@@ -1,8 +1,14 @@
-from flask import render_template
-
+from flask import render_template, jsonify
+import webserver.exceptions
 
 def init_error_handlers(app):
 
+    @app.errorhandler(webserver.exceptions.InvalidAPIUsage)
```
gentlecat: Maybe would be better to split up API exceptions and regular ones into different init functions?

webserver/templates/navbar.html
alastair: I tested with the old version of flask login and it works, so we'll be OK to release this without the version upgrade

webserver/exceptions.py
alastair: good idea :100:

webserver/errors.py
alastair: nah, it's all webserver related. We don't need another file.

gentlecat: I meant another function in the same file.

alastair: Ah, right. No, I don't think it's needed (its still an error handler for the website)

PR #141, opened 2015-10-14
Offset is zero indexed ...

Sophist-UK:
... whilst position/previous/next are one indexed.

Obviously next cannot be the same as position, hence this change.

### DIFF
--- a/webserver/views/data.py
+++ b/webserver/views/data.py
@@ -111,7 +111,7 @@ def summary(mbid):
             tomahawk_url=_get_tomahawk_url(recording_info),
             submission_count=submission_count,
             position=offset,
-            previous=offset -1 if offset > 0 else None,
+            previous=offset - 1 if offset > 0 else None,
             next=offset + 1 if offset < submission_count - 1 else None,
             offset=offset,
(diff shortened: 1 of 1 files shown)


alastair:
Thanks.
From looking at this it seems like we already use offset (0 indexing) in the `?n` parameter (which is turned into the offset variable). From a quick glance, I'm not sure this fix is correct - but I won't have time to look at the patch in detail for another few weeks.

Sophist-UK:
Ok - I have found how this is used - on the summary page in web view - and the paging works currently, so I will change this to be a correct fix.

Sophist-UK:
How about this?

alastair:
I'm not sure what you're wanting to do now - `position` is only used in the template to show "x of y" - it's the *x*, and is 1 more than offset, since offset counts from 0 and people count from 1. What are you trying to do with the change?

Sophist-UK:
Well it looked like a bug - but obviously it wasn't.

alastair:
Thanks for taking the time to make a patch (even if didn't work out)!

PR #143, opened 2015-10-15
[WIP] AB-50: Initial schema change for high level split

alastair:
http://tickets.musicbrainz.org/browse/AB-50
http://tickets.musicbrainz.org/browse/AB-56

Initial schema changes to High level tables to split output of models into separate rows. 
During discussion we realised that we could use the data migration as a chance to upgrade to postgresql 9.4 and jsonb data types.
MessyBrainz has shown that we can install postgres 9.4 from the postgres repos on a machine with the same setup as spike and run both at the same time.
We can run a migration row by row to split the data into models while AB is still up and running, and do a final swap over with minimal downtime.
While we're changing the schema, take the chance to split lowlevel json into its own table. No versioning yet, we'll leave that for a later stage.

img_20151015_181537437

(Top is original schema, below is new one)
Take the chance also to swap the fk from highlevel.data -> highlevel_json.id to json.highlevel -> highlevel.id (and same for lowlevel).

Still looking for a good name for `hl_item` or `hl_model`

Todo:
- [x] Low level split migration
- [x] High level split migration
- [x] new schema low level submit method
- [x] new schema lowlevel get method
- [x] new schema high level get method
- [x] Admin interface to set modules import flag?
- [x] Work out how to have submission going to old schema but then quickly switch to new schema

Migration plan:
 - [x] Install postgres 9.4
 - [x] pgbouncer
 - [x] Set up beta site
 - [x] Run migration
 - [ ] merge new schema branch
 - [ ] clean up old schema code
 - [ ] New dump of data


admin/sql/create_db.sql
```diff
@@ -2,5 +2,5 @@
 
 -- Create the user and the database. Must run as user postgres.
 
-CREATE USER acousticbrainz NOCREATEDB NOCREATEUSER;
-CREATE DATABASE acousticbrainz WITH OWNER = acousticbrainz TEMPLATE template0 ENCODING = 'UNICODE';
+-- CREATE USER acousticbrainz NOCREATEDB NOCREATEUSER;
+CREATE DATABASE acousticbrainz94 WITH OWNER = acousticbrainz TEMPLATE template0 ENCODING = 'UNICODE';
```
gentlecat: Is this going to be renamed after migration?

alastair: this was for my testing. On spike we'll be running on 2 different database servers so the database name can be the same. We'll have to sort out a step by step process.

alastair:
Lowlevel splitting is roughly working. Still to be fixed:
 - [ ] Perform many items in a transaction (for speed)
 - [ ] Catch any errors from postgres
 - [ ] Keep counter of where we're up to, to restart if needed
 - [ ] Update the `id` field sequence

Looking through local code, it looks like we actually removed bad utf-8 items from the database already (just the metadata keys), so perhaps this won't be an issue during the conversion.

admin/sql/create_indexes.sql
```diff
@@ -2,10 +2,17 @@ BEGIN;
...
+CREATE UNIQUE INDEX data_sha256_ndx_lowlevel_json ON lowlevel_json (data_sha256);
 
 CREATE INDEX mbid_ndx_highlevel ON highlevel (mbid);
-CREATE INDEX build_sha1_ndx_highlevel ON lowlevel (build_sha1);
+CREATE INDEX build_sha1_ndx_highlevel ON highlevel (build_sha1);
+
+CREATE UNIQUE INDEX data_sha256_ndx_highlevel_meta ON highlevel_meta (data_sha256);
+-- This index should not be unique on data, because it's possible to have the same estimate for
+-- two different recordings. Instead map it over data, model, and the id of the item
+CREATE INDEX data_sha256_ndx_highlevel_model ON highlevel_model (data_sha256, highlevel, model);
```
alastair: Error here, should be `UNIQUE` again.
I think this is right. With a test migration I did I had two identical `data` fields for two different `hl.id`. A unique row in this table is a combination of the input file, the model used on it, and the data. If any of these change we need a new row.

admin/sql/create_models.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+INSERT INTO model (model, model_version, build_sha1, date, include) VALUES ('danceability', 'v2.1_beta1', 'v2.1_beta1-228-g260734a', now(), 't');
```
alastair: By adding this before the migration we don't have to do a lookup for the id each time we do a insert into `highlevel_model`, we just have a lookup table in python. much faster

admin/sql/create_tables.sql
```diff
@@ -5,25 +5,45 @@ CREATE TABLE lowlevel (
...
+  data        JSONB    NOT NULL,
+  data_sha256 CHAR(64) NOT NULL,
+  model       INTEGER -- FK to model.id
+);
+
+CREATE TABLE model (
+  id            SERIAL,
+  model         TEXT NOT NULL,
+  model_version TEXT NOT NULL,
+  build_sha1    TEXT NOT NULL,
```
alastair: In the highlevel json file (http://acousticbrainz.org/98255a8c-017a-4bc7-8dd6-1fa36124572b/high-level?n=0) we have git sha, version, and build sha for both essentia and gaia:
```
version: {
  highlevel: {
    essentia: "2.1-beta1",
    essentia_build_sha: "8e24b98b71ad84f3024c7541412f02124a26d327",
    essentia_git_sha: "v2.1_beta1-228-g260734a",
    extractor: "music 1.0",
    gaia: "2.4-dev",
    gaia_git_sha: "857329b",
    models_essentia_git_sha: "v2.1_beta1"
  },
  lowlevel: {
    essentia: "2.1-beta2",
    essentia_build_sha: "c0e182ba78c2de26e491b7b7a4a2c98c1ed34208",
    essentia_git_sha: "v2.1_beta2-2-gbb40004",
    extractor: "music 1.0"
  }
}
}
```
I don't know what data we should store in this table. Git revisions seem more useful that sha1 of the binary. Should we have essentia or gaia?

admin/split-highlevel.py
```diff
@@ -0,0 +1,75 @@
...
+            """)
+
+        res2 = conn2.execute(q, {"ids": id_list})
+        while True:
+            row = res2.fetchone()
+            if not row:
+                break
+            row = dict(row)
+            print "converting", row["id"]
+            with db.engine94.begin() as connection:
```
alastair: Currently a transaction per original row. I think we could make this do many rows at a time

admin/sql/create_tables.sql
gentlecat: What's in `model`, `model_version`, and `build_sha1`? It seems like version of gaia would be important in case of models, since that's what we use to generate them. And I agree that git revision should be more useful.

alastair: we had another chat and decided to break the entire `highlevel` block into a `version` table. This will contain a json blob of high level *or* lowlevel data. and the `highlevel` and `lowlevel` tables will link to it.

dataset_eval/artistfilter.py
```diff
@@ -0,0 +1,168 @@
...
+    if options["filter_type"] == "artist":
+        print ("Filtering by artist")
+        train, test = split_groundtruth(datadict)
+        print (train, test)
+    else:
+        train = datadict
+        test = {}
+    if options["normalize"]:
+        print("Normalising")
+        train, remaining = normalise_datadict(train, 450)
```
alastair: This is terrible. This value should be based on the smallest size class in the training set. Perhaps there should be a limit somewhere in a configuration file.

hl_extractor/job_calc.py
```diff
@@ -0,0 +1,241 @@
...
+
+                    try:
+                        jdata = json.loads(hl_data)
+                    except ValueError:
+                        print("error %s: Cannot parse result document" % mbid)
+                        print(hl_data)
+                        sys.stdout.flush()
+                        jdata = {}
+
+                    db.data.write_high_level(mbid, ll_id, jdata, build_sha1)
```
alastair: If `jdata` is empty, we need to work out how to add a dummy highlevel item for the combination of `ll_id` and `module`, so that when we ask for the next lot of data this failed example is not here.

alastair:
in the high level calculation we need a way of creating the profile file with just the high level models that we want.

alastair:
We still have some open questions about how to dump highlevel data.

The `models` table can be modified, it's not append-only. Therefore we can't just use `copy` to put it into the database. Instead we should scan each line into a tuple, search its id in the database, if it exists do an update, otherwise do an insert. We never delete from this table, just change it status, so we don't need to consider this case.

We need to decide how to dump highlevel json incremental dumps.
We have 2 situations to consider: a new model is promoted to live, and a model is demoted from live to hidden.
We shouldn't modify files that people have on their filesystem between incremental dumps. Therefore, I'm tempted to suggest that we dump highlevel as one file per model, instead of one file per submission. That is, we have `uuid-1-metadata.json`, `uuid-1-lastfm.json`, `uuid-1-dortmund.json`. This means that if we have a new model `allmusicmood` we can create a new file in the next incremental dump `uuid-1-allmusicmood.json` and it can uncompress in place. 

The case of a demoted model is more difficult. We can't record a deletion in the archive except through an additional metadata file. One proposal from Dmitry is to just do a new full dump when we encounter this case (the highlevel dump is by far the smallest, anyway).
We need to keep track of if a model was demoted between two dumps. Therefore in addition/instead of a `created` field, we need an `updated` and also a `last_state`, so that we can see if the state has changed from active to hidden.

alastair:
We should migrate to postgres 9.5, too!

gentlecat:
:confetti_ball:

alastair:
:disappointed_relieved:

alastair:
wait, that's "disappointed relieved", I wanted "phew"

:sweat_smile:

alastair:
better

PR #144, opened 2015-10-24
[WIP] Updated UI and database schema for new dataset evaluation features

gentlecat:


db/dataset_eval.py
```diff
@@ -138,12 +168,49 @@ def set_job_status(job_id, status, status_msg=None):
...
+            "FROM dataset_snapshot "
+            "WHERE id = %s",
+            (id,)
+        )
+        row = result.fetchone()
+        return dict(row) if row else None
+
+
+def add_dataset_snapshot(data):
+    with db.engine.connect() as connection:
```
alastair: for `INSERT` you should use `db.engine.begin` to use a transaction

PR #148, opened 2015-11-10
Remove duplicate recordings from classes in a dataset before inserting

gentlecat:


db/dataset.py
```diff
@@ -89,6 +89,9 @@ def create_from_dict(dictionary, author_id):
                            (cls["name"], cls["description"], dataset_id))
             cls_id = result.fetchone()[0]
 
+            # Removing duplicate recordings
+            cls["recordings"] = list(set(cls["recordings"]))
```
alastair: This is the same recording id multiple times in the same class, right?  Does the UI also block this?

gentlecat: Yeah, there's a check in the editor. This was coming from CSV import.

PR #149, opened 2015-12-03
Cleanup and fix parts related to statistics

gentlecat:
[Detailed statistics](http://acousticbrainz.org/statistics-graph) page stopped working correctly. It's probably time to revisit this part of code, figure out how it works, and do some cleanup. :beetle:


gentlecat:
@alastair, feel free to push there or merge it. I'm not sure if I'll be able to finish this before the end of the year.

db/stats.py
```diff
@@ -10,23 +10,30 @@
...
+    last_submissions = db.cache.get(cache_key)
+    if not last_submissions:
         with db.engine.connect() as connection:
-            result = connection.execute("""SELECT mbid,
-                                     data->'metadata'->'tags'->'artist'->>0,
-                                     data->'metadata'->'tags'->'title'->>0
-                                FROM lowlevel
-                            ORDER BY id DESC
-                               LIMIT 5
-                              OFFSET 10""")
```
gentlecat: Not sure why we need an offset there.

alastair: because the highlevel computation might not have finished yet. We want to show 5 that we are pretty sure are complete before someone clicks on the link. We could instead join to highlevel

gentlecat: Ah, OK.

db/stats.py
```diff
@@ -1,119 +1,261 @@
...
+    if sorted(stats_names) != sorted(stats.keys()) or last_collected is None:
 
-            result = connection.execute("SELECT now()")
-            last_collected = result.fetchone()[0]
+        stats = dict([(name, 0) for name in stats_names])
 
-        value = stats_parameters
+        with db.engine.connect() as connection:
+            # FIXME(roman): Ideally all these stats should be calculated in one
+            # query which would also get current time, for better accuracy.
```
gentlecat: I think this is what might be causing problems with incorrect stats display.

gentlecat:
OK, I think problem with incorrect values shouldn't happen when new stats are recorded now. Another issue is that we still have old timestamps in the database, which cause calculation of total counts to not work correctly. I'm not sure how to fix that, apart from adjusting timestamp values manually after retrieval.

All that assuming I found the actual cause of that problem. :snake:

alastair:
We've noticed that stats in general are quite messy. One issue is https://github.com/metabrainz/acousticbrainz-server/blob/stats-cleanup/db/stats.py#L145-L149 which tries to combine independent stats based on their timestamp, which may be different

Also, stats only trigger if someone visits the homepage, so we are only guaranteed to have stats at least an hour apart, not every hour.

After discussion with @gentlecat, we'd like to see these changes made before we merge:

- [ ] Tests for adding/getting these stats
- [ ] Replace stats generation with a script that we can call from `manage.py`, which always puts exactly 1 hour between stats and counts submissions by their submit date, not by a `count` and `now`
- [ ] Perhaps store a total instead of summing 2 values
- [ ] Recompute *all* stats (perhaps when we release 9.4 + schema change)

db/stats.py
```diff
@@ -1,119 +1,244 @@
...
+    Arguments:
+      date: a datetime
+    """
+    delta = datetime.timedelta(hours=1)
+    date = date + delta
+    date = date.replace(minute=0, second=0, microsecond=0)
+    return date
+
+def compute_stats(to_date):
+    # Compute stats up to the given date
```
gentlecat: Docstring. :scroll:

db/stats.py
```diff
@@ -1,119 +1,244 @@
...
+        LOWLEVEL_LOSSY,
+        LOWLEVEL_LOSSLESS,
+        LOWLEVEL_LOSSY_UNIQUE,
+        LOWLEVEL_LOSSLESS_UNIQUE,
+        LOWLEVEL_TOTAL,
+        LOWLEVEL_TOTAL_UNIQUE,
+    ]
+    cache_key_prefix = "ac-num-"
+    # TODO: Port this to new implementation (don't use protected parts):
+    stats = db.cache._mc.get_multi(stats_names, key_prefix=cache_key_prefix)
```
gentlecat: I think this was not working before because of exactly that issue in caching wrapper that you found. So it's probably safe to use our custom interface now.

db/stats.py
alastair: yeah, I just pushed this so that I would have it on my other machine. Still some stuff to do

db/stats.py
alastair: yeah, it should be fine to use the wrapper now. However, we can decide if we still want to cache this - especially because we do the calculation offline now, and only once an hour.
Should we

1) Update this cache value once every 10 minutes by counting items in the database
2) Update this cache value once an hour by reading the stats table
3) Just show hourly stats by reading the stats table

db/stats.py
```diff
@@ -1,119 +1,244 @@
...
-            if update_db:
-                for key, value in six.iteritems(stats_parameters):
-                    connection.execute("INSERT INTO statistics (collected, name, value) VALUES (now(), %s, %s) RETURNING collected", (key, value))
-
-            result = connection.execute("SELECT now()")
-            last_collected = result.fetchone()[0]
-
-        value = stats_parameters
+            stats = _get_latest_stats(connection)
+            last_collected = _current_db_time(connection)
```
gentlecat: These two functions are undefined.

db/stats.py
gentlecat: Oh, that's right. I think we can just read values from the database and cache them for some time. 10 minutes sounds good.

I'm not sure about usage of `last_collected` value that we currently have. What if we look at `collected` column in the last inserted row (shouldn't matter which particular statistic it is)?

gentlecat: Also, do we really need to store different stats separately?

alastair: `last_collected` is a good idea. We want to use memcache to reduce the amount that we access the database. What do you mean by separately? Do you mean that we could store a single key with a dict instead?

gentlecat: We can cache `last_collected` like we do the actual values. And yes, single key with a dict since we don't use these values separately anyway.

Looks like `get_stats` function needs an update.

alastair:
I've been looking at the part which computes stats for the sidebar (every 10 minutes). Currently the way it works is that if someone visits the page and the stats are out of date, they get recomputed. Even on spike, this takes about 5 seconds, which is an unacceptable wait for the page to load.

I will set up an hourly cron task, which writes to the `statistics` table. This is used for the detailed stats page.
We will also have a 10 minute task which computes the same value, but adds the data to memcache.
The stats sidebar can read from memcache, and if the value doesn't exist, take the most recent item from the stats table (which should only be up to an hour old)

db/test/test_stats.py
```diff
@@ -1,46 +1,234 @@
...
+                  "lowlevel-total": 35, "lowlevel-total-unique": 20}
+        date2 = datetime.datetime(2016, 01, 11, 00, 00, tzinfo=pytz.utc)
+        data = [
+            {"collected": date1, "stats": stats1},
+            {"collected": date2, "stats": stats2}
+        ]
+
+        formatted = db.stats.format_statistics(data)
+        expected_formatted = [
+            {'data': [[1452380400000, 15], [1452466800000, 20]], 'name': 'Lossless (all)'},
```
alastair: issue with timezones here. This number isn't correct

alastair:
Finally done!

db/stats.py
```diff
@@ -1,120 +1,278 @@
...
         ]
-        db.cache.set('last-submitted-data', last_submitted_data, time=LAST_MBIDS_CACHE_TIMEOUT)
+        db.cache.set(cache_key, last_submissions, time=LAST_MBIDS_CACHE_TIMEOUT)
+
+    return last_submissions
 
-    return last_submitted_data
 
+def compute_stats(to_date):
+    # Compute stats up to the given date
```
gentlecat: Docsting.

db/stats.py
```diff
@@ -1,120 +1,278 @@
...
+    return last_collected, stats
 
-            result = connection.execute("SELECT now()")
-            last_collected = result.fetchone()[0]
 
-        value = stats_parameters
+def format_statistics(data):
+    """
+    :param data: data from load_statistics_data
+    :return: statistics formatted for
```
gentlecat: Can you clarify a bit. Now it's confusing what the formatting is for.

db/stats.py
```diff
@@ -1,120 +1,278 @@
...
+            counts[k].append([ts, stats[k]])
 
-def get_statistics_data():
+    stats = [{"name": stats_key_map.get(key, key), "data": data} for key, data in counts.items()]
+
+    return stats
+
+
+def load_statistics_data(limit=None):
+    # Posgres doesn't let you create a json dictionary using values
```
gentlecat: r/Posgres/Postgres

manage.py
```diff
@@ -133,5 +135,20 @@ def import_data(archive):
 cli.add_command(db.dump_manage.cli, name="dump")
 
 
+@cli.command()
+def compute_stats():
+    db.init_db_engine(config.SQLALCHEMY_DATABASE_URI)
+    import datetime
+    import pytz
+    db.stats.compute_stats(datetime.datetime.now(pytz.utc))
```
gentlecat: This is supposed to be used once when there are no stats, right?

manage.py
```diff
@@ -133,5 +135,20 @@ def import_data(archive):
 cli.add_command(db.dump_manage.cli, name="dump")
 
 
+@cli.command()
+def compute_stats():
+    db.init_db_engine(config.SQLALCHEMY_DATABASE_URI)
+    import datetime
+    import pytz
```
gentlecat: Maybe it's better to put these at the top of the module?

gentlecat:
LGTM. I commented on some minor issues.

db/stats.py
gentlecat: Also, for the sake of consistency we should decide on common way to document arguments, return values, and things like that. I tend to just use https://google.github.io/styleguide/pyguide.html#Comments, but I picked it just because it is easy to read. Maybe there's something better or more conventional.

PR #152, opened 2016-01-05
AB-105: Improve error handling during JS builds

gentlecat:


webserver/static/gulpfile.js
```diff
@@ -42,6 +42,9 @@ function writeResource(stream) {
     }))
     .on('finish', function () {
       deferred.resolve();
+    })
+    .on('error', function (error) {
```
mwiencek: I think that needs to be directly after `stream`, because it only catches errors that occur immediately before it. Unless you tested it and it works, in which case ignore me. :)

gentlecat: I don't think it worked correctly, no. :)

gentlecat:
Well, at least it fails on errors now.
```
local: ./node_modules/.bin/gulp
[15:30:20] Using gulpfile ~/acousticbrainz-server/gulpfile.js
[15:30:20] Starting 'styles'...
[15:30:20] 'styles' errored after 278 ms
[15:30:20] TypeError: Cannot read property 'fail' of undefined
    at Gulp.buildStyles (gulpfile.js:54:10)
    at module.exports (/home/vagrant/acousticbrainz-server/node_modules/orchestrator/lib/runTask.js:34:7)
    at Gulp.Orchestrator._runTask (/home/vagrant/acousticbrainz-server/node_modules/orchestrator/index.js:273:3)
    at Gulp.Orchestrator._runStep (/home/vagrant/acousticbrainz-server/node_modules/orchestrator/index.js:214:10)
    at Gulp.Orchestrator.start (/home/vagrant/acousticbrainz-server/node_modules/orchestrator/index.js:134:8)
    at /home/vagrant/acousticbrainz-server/node_modules/gulp/bin/gulp.js:129:20
    at nextTickCallbackWith0Args (node.js:452:9)
    at process._tickCallback (node.js:381:13)
    at Function.Module.runMain (module.js:449:11)
    at startup (node.js:139:18)

/home/vagrant/acousticbrainz-server/node_modules/q/q.js:155
                throw e;
                ^
Error: directive options not recognised in file /home/vagrant/acousticbrainz-server/webserver/static/styles/main.less line no. 1

Fatal error: local() encountered an error (return code 1) while executing './node_modules/.bin/gulp'

Aborting.
```
Improvement?

PR #155, opened 2016-02-18
Solved AB-79

kartikgupta0909:
This request solves the bug AB-79. It displays a warning whenever a user tries to edit a dataset which has a pending or a running job.

### DIFF
--- a/webserver/views/datasets.py
+++ b/webserver/views/datasets.py
@@ -15,6 +15,9 @@ import db.user
 import jsonschema
 import csv
 import six
+import config
+import db
+import db.dataset_eval
 
 datasets_bp = Blueprint("datasets", __name__)
 
@@ -144,7 +147,11 @@ def edit(dataset_id):
     ds = get_dataset(dataset_id)
     if ds["author"] != current_user.id:
         raise Unauthorized("You can't edit this dataset.")
-
+    jobs = db.dataset_eval.get_jobs_for_dataset(str(dataset_id))
+    for job in jobs:
+        if(job["status"]=="running" or job["status"]=="pending"):
+            flash.warn("There is a pending job for this dataset")
+            break
     if request.method == "POST":
         dataset_dict = request.get_json()
         if not dataset_dict:
--- a/webserver/views/datasets.py
+++ b/webserver/views/datasets.py
@@ -145,12 +145,13 @@ def _parse_dataset_csv(file):
 @login_required
 def edit(dataset_id):
     ds = get_dataset(dataset_id)
+    pending = "0"
     if ds["author"] != current_user.id:
         raise Unauthorized("You can't edit this dataset.")
     jobs = db.dataset_eval.get_jobs_for_dataset(str(dataset_id))
     for job in jobs:
-        if(job["status"]=="running" or job["status"]=="pending"):
-            flash.warn("There is a pending job for this dataset")
+        if(job["status"] in ("running","pending")):
+            pending = "1"            
             break
     if request.method == "POST":
         dataset_dict = request.get_json()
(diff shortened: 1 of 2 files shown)


gentlecat:
Some general comments regarding the workflow:
* It's better to put some info about what was actually done in commit message, in addition to ticket reference (there's a useful link in https://github.com/metabrainz/acousticbrainz-server/blob/master/CONTRIBUTING.md).
* When you start working on a ticket, assign it to yourself and set status "In Progress".
* After you have a pull request, link it in the comments for the ticket.

webserver/views/datasets.py
```diff
@@ -144,7 +147,11 @@ def edit(dataset_id):
     ds = get_dataset(dataset_id)
     if ds["author"] != current_user.id:
         raise Unauthorized("You can't edit this dataset.")
-
+    jobs = db.dataset_eval.get_jobs_for_dataset(str(dataset_id))
+    for job in jobs:
+        if(job["status"]=="running" or job["status"]=="pending"):
```
gentlecat: This can be simplified to `job["status"] in ("running", "pending")`.

webserver/views/datasets.py
```diff
@@ -15,6 +15,9 @@
 import jsonschema
 import csv
 import six
+import config
+import db
```
gentlecat: This import looks unused.

webserver/views/datasets.py
```diff
@@ -144,7 +147,11 @@ def edit(dataset_id):
     ds = get_dataset(dataset_id)
     if ds["author"] != current_user.id:
         raise Unauthorized("You can't edit this dataset.")
-
+    jobs = db.dataset_eval.get_jobs_for_dataset(str(dataset_id))
+    for job in jobs:
+        if(job["status"]=="running" or job["status"]=="pending"):
+            flash.warn("There is a pending job for this dataset")
```
gentlecat: "Pending job" is more of our internal terminology. It might be better to write "This dataset is in evaluation queue." (ideally with a link to see status, etc.)

alastair: I agree here. If we're going to use links, etc, it might be better to not use flash(). I wonder if instead we can set a flag in the context and have an `{% if %}` block in the template.

gentlecat: Dataset pages are using React, so it's probably a matter of inserting more values into one of json endpoints.

kartikgupta0909:
I have updated the pull request to send a variable "pending" from webserver/views/datasets.py to webserver/templates/datasets/edit.html and then print a message "Dataset is in the evaluation queue", if there is a pending job for that dataset.

webserver/views/datasets.py
```diff
@@ -142,9 +145,14 @@ def _parse_dataset_csv(file):
 @login_required
 def edit(dataset_id):
     ds = get_dataset(dataset_id)
+    pending = "0"
     if ds["author"] != current_user.id:
         raise Unauthorized("You can't edit this dataset.")
-
+    jobs = db.dataset_eval.get_jobs_for_dataset(str(dataset_id))
+    for job in jobs:
+        if(job["status"] in ("running","pending")):
```
alastair: don't use `()` in python. Just `if job['status'] in....` is fine

webserver/views/datasets.py
```diff
@@ -142,9 +145,14 @@ def _parse_dataset_csv(file):
 @login_required
 def edit(dataset_id):
     ds = get_dataset(dataset_id)
+    pending = "0"
```
alastair: This should be a boolean, not a string

webserver/views/datasets.py
gentlecat: There also should be a space after a comma. If you are wondering about how to format Python code, check this website: http://pep8.org/.

gentlecat:
See #170.

PR #157, opened 2016-02-24
AB-114: Replace toma.hk embed with youtube search

alastair:
http://tickets.musicbrainz.org/browse/AB-114

### DIFF
--- a/webserver/templates/data/summary-metadata.html
+++ b/webserver/templates/data/summary-metadata.html
@@ -42,8 +42,8 @@
   </div>
 
   <div class="col-md-4 text-center">
-    {% if tomahawk_url %}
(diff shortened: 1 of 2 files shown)


gentlecat:
:boom:

alastair:
We can set width and height on the iframe. Can you make a suggestion?

gentlecat:
200px is the minimum value according to https://developers.google.com/youtube/player_parameters. So I'd say adjust it to fill up the space.
```html
<iframe width="200" height="200" ...
```

gentlecat:
And you can set width to `100%` and let our styling handle the rest. It would probably work better this way on mobile layouts.

PR #159, opened 2016-02-24
Add link to ideas page on the wiki

alastair:


gentlecat:
:bulb: :shipit: :bulb:

alastair:
Closing because it's based on a version of master before we force-pushed it, but I will apply the patch separately

PR #164, opened 2016-02-27
Fix layout inconsistencies between dataset viewer and editor

gentlecat:


alastair:
I'd still like to see a cancel button, or something like the "<- back to dataset list" that you see on the view page.

alastair:
(fyi i rebased this branch off master again and force pushed. please take care when you make changes)

gentlecat:
OK, added cancel button into editor/creator.

PR #165, opened 2016-02-27
AB-63: Dataset list

alastair:
http://tickets.musicbrainz.org/browse/AB-63


alastair:
missing tests :(

webserver/templates/datasets/list.html
```diff
@@ -0,0 +1,27 @@
...
+        </tr>
+        {% for ds in datasets %}
+        <tr>
+            <td>{{ ds["author_name"] }}</td>
+            <td>
+                <a href="{{ url_for('datasets.view', id=ds['id']) }}">
+                    {{ ds["name"] }}
+                </a>
+            </td>
+            <td>{{ ds["created"] }}</td>
```
alastair: should be nicely formatted

webserver/templates/datasets/list.html
```diff
@@ -0,0 +1,27 @@
...
+    <table class="table">
+        <tr>
+            <th>author</th>
+            <th>dataset</th>
+            <th>created</th>
+            <th>evaluation status</th>
+        </tr>
+        {% for ds in datasets %}
+        <tr>
+            <td>{{ ds["author_name"] }}</td>
```
alastair: link to author name

webserver/templates/datasets/list.html
```diff
@@ -0,0 +1,27 @@
...
+{% block title %}Public datasets{% endblock %}
+
+{% block content %}
+    <h3>Public datasets</h3>
+    <table class="table">
+        <tr>
+            <th>author</th>
+            <th>dataset</th>
+            <th>created</th>
+            <th>evaluation status</th>
```
alastair: to get the status we should do just one query, not one query per dataset

webserver/templates/datasets/list.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.html' %}
+
+{% block title %}Public datasets{% endblock %}
+
+{% block content %}
+    <h3>Public datasets</h3>
```
alastair: would be good to have an option to filter by *running*, *in queue*, *finished*, *new*

If no filter, put in running and in queue first

We will need paging before we get too many datasets

gentlecat: Page title should be `<h2 class="page-title">...</h2>`.

webserver/templates/datasets/list.html
gentlecat: Can use `|datetime`.

webserver/templates/datasets/list.html
```diff
@@ -13,7 +13,10 @@
         </tr>
         {% for ds in datasets %}
         <tr>
-            <td>{{ ds["author_name"] }}</td>
+            <td>
+                <a href="{{ url_for('user.profile', musicbrainz_id=ds['author_name']) }}">
+                    {{ ds["author_name"] }}
```
alastair: missing a `</a>`

db/dataset.py
```diff
@@ -183,7 +183,15 @@ def get_public_datasets():
             row = dict(row)
             dataset_id = row["id"].encode('utf-8')
             row["status"] = dataset_status[dataset_id]
-            datasets.append(row)
+            if(status_value == "all" or row["status"] == status_value):
```
gentlecat: Parentheses here aren't necessary. You can try using [pylint](https://www.pylint.org/) which catches most of these problems.

gentlecat: We probably should set it up as a part of Jenkins build too.

db/dataset.py
```diff
@@ -161,7 +161,7 @@ def get(id):
         else:
             return None
 
-def get_public_datasets():
+def get_public_datasets(status_value,page_no):
```
gentlecat: On this level it would be better to work with offsets and limits.

kartikgupta0909: I think using offsets and limits is good too. Will do that.

db/dataset.py
```diff
@@ -183,7 +183,15 @@ def get_public_datasets():
             row = dict(row)
             dataset_id = row["id"].encode('utf-8')
             row["status"] = dataset_status[dataset_id]
```
alastair: If this dataset has not been submitted, it will not have a status. This will fail with a `KeyError`. You should use `dataset_status.get(dataset_id)` instead.

webserver/views/datasets.py
```diff
@@ -16,9 +16,9 @@
 datasets_bp = Blueprint("datasets", __name__)
 
 
-@datasets_bp.route("/")
-def list_datasets():
-    datasets = db.dataset.get_public_datasets()
+@datasets_bp.route("/<status_value>/<int:page_no>")
```
alastair: Be careful about this change. Now you can't go to `/datasets` to view the default page. You should fill in defaults if none are provided.

gentlecat: I think these should be a part of query string anyway.

db/dataset.py
alastair: i've been playing around with this a bit. Here's a query which gets status as well, using a `LATERAL JOIN`, showing a null status if the dataset hasn't been submitted

```sql
explain analyze SELECT dataset.id :: TEXT
     , dataset.name
     , dataset.description
     , dataset.author
     , "user".musicbrainz_id AS author_name
     , dataset.created
     , job.status
  FROM dataset
  JOIN "user"
    ON "user".id = dataset.author
  LEFT JOIN LATERAL (SELECT status
                       FROM dataset_eval_jobs
                      WHERE dataset.id = dataset_eval_jobs.dataset_id
                   ORDER BY updated DESC
                      LIMIT 1)
                   AS job ON TRUE
 WHERE dataset.public = 't'
  and job.status = 'pending'
ORDER BY dataset.created DESC
OFFSET 0 LIMIT 10
```

You can modify the `WHERE job.status` and `OFFSET` and `LIMIT` to choose what items you show

db/dataset.py
```diff
@@ -160,6 +160,62 @@ def get(id):
...
+                 , dataset.name
+                 , dataset.description
+                 , dataset.author
+                 , "user".musicbrainz_id AS author_name
+                 , dataset.created
+                 , job.status
+            FROM dataset
+            JOIN "user"
+              ON "user".id = dataset.author
+              LEFT JOIN LATERAL (SELECT status
```
mineo: Note that `INSTALL.md` list Postgres 9.2 as a dependency, but lateral joins are a 9.3 feature.

alastair: And `jsonb` is a 9.4 feature ;)

mineo: I didn't know AB used JSONB. In that case, ignore my comment (I see you've
already updated INSTALL.md).

alastair:
I've made some changes to this, including simplifying the SQL that we use and adding tests.
The pager still needs to be fixed. It should only show if there are actually next/previous items to show.

db/dataset.py
```diff
@@ -102,15 +105,19 @@ def get(id):
...
 
-def get_public_datasets(status_value, offset, limit):
-    result = ''
+def get_public_datasets(status, offset, limit):
+    if status == "all":
+        statuses = db.dataset_eval.VALID_STATUSES
+    elif status in db.dataset_eval.VALID_STATUSES:
+        statuses = [status]
+    else:
+        raise Exception("Unknown status")
```
gentlecat: This should probably be a `ValueError`. Otherwise it's hard to catch it without doing `except Exception`, which is bad.

alastair: thanks for the reminder. I just put `Exception` in as a temporary value, but ValueError is perfect for this.

alastair:
Pagination is here now. It looks like this

screenshot from 2016-06-13 18-00-49

and page 2 looks like this

screenshot from 2016-06-13 18-00-53

and if you choose just 1 status it looks like this

screenshot from 2016-06-13 18-01-06


I don't do any pagination shrinking, so if we get to more than ~20 pages we might have to think again about how to make it clearer

alastair:
Merged! Still waiting on a `/datasets` endpoint which links to both the list page and the accuracy page. Perhaps with the rewrite of `/data` to move documentation to rtd

webserver/templates/datasets/list.html
```diff
@@ -0,0 +1,46 @@
...
+        <tr>
+            <th>Author</th>
+            <th>Dataset</th>
+            <th>Created</th>
+            <th>Evaluation status</th>
+        </tr>
+        {% for ds in datasets %}
+        <tr>
+            <td>
+                <a href="{{ url_for('user.profile', musicbrainz_id=ds['author_name']) }}">
```
gentlecat: This tag is not closed.

webserver/templates/datasets/list.html
```diff
@@ -0,0 +1,46 @@
...
+        {% if status == 'pending' %}<li class="active">{% else %}<li>{% endif %}
+        <a href="{{url_for('.list_datasets', status='pending')}}">Pending</a></li>
+        {% if status == 'running' %}<li class="active">{% else %}<li>{% endif %}
+        <a href="{{url_for('.list_datasets', status='running')}}">Running</a></li>
+        {% if status == 'done' %}<li class="active">{% else %}<li>{% endif %}
+        <a href="{{url_for('.list_datasets', status='done')}}">Done</a></li>
+        {% if status == 'failed' %}<li class="active">{% else %}<li>{% endif %}
+        <a href="{{url_for('.list_datasets', status='failed')}}">Failed</a></li>
+    </ul>
+    <p></p>
```
gentlecat: Empty?

webserver/templates/macros.html
```diff
@@ -11,3 +11,40 @@
...
+	      <span aria-hidden="true">&raquo;</span>
+		{% else %}
+		<li>
+          <a href="{{nextpage}}" aria-label="Next">
+			<span aria-hidden="true">&raquo;</span>
+		  </a>
+		{% endif %}
+		</li>
+	  </ul>
+	</nav>
```
gentlecat: Can you clean up indentation here? It's very hard to read.

webserver/forms.py
```diff
@@ -23,3 +26,12 @@ class DatasetEvaluationForm(Form):
         (dataset_eval.FILTER_ARTIST, "By Artist"),
     ])
     normalize = BooleanField("Normalize classes")
+
+class DatasetListForm(Form):
```
gentlecat: I don't see where this is used.

gentlecat:
Can you give some time to take a look in future?

PR #166, opened 2016-02-28
Better dataset validation

gentlecat:


gentlecat:
*http://ci.musicbrainz.org/job/acousticbrainz-server/17/* :eyes:

utils/dataset_validator.py
```diff
@@ -0,0 +1,109 @@
...
+DATASET_NAME_LEN_MAX = 100
+
+MIN_CLASSES = 2
+CLASS_NAME_LEN_MIN = 1
+CLASS_NAME_LEN_MAX = 100
+
+MIN_RECORDINGS_IN_CLASS = 2
+
+
+def validate(dataset, complete=False):
```
alastair: documentation

alastair: what's complete? is this validation for building the model, instead of general validation?

alastair: this does two types of validation - that the structure of the dataset objects is correct, and that the contents of the dataset follow a set of rules. Do you think it makes sense to combine them. My thought is that since dataset objects are always created by the app, this is less needed.

gentlecat: I added a docsting with some info. Datasets are not always created by the app. This is also used to validate datasets that are submitted to the "API" endpoint. It is used by our creation UI, but it's still important to validate.

gentlecat:
Fixed conflicts and made some tweaks to the way completeness of the dataset is checked.

utils/dataset_validator.py
```diff
@@ -0,0 +1,116 @@
...
+        raise ValidationException("Class name must be a string.")
+    if not (DATASET_NAME_LEN_MIN < len(dataset["name"]) < DATASET_NAME_LEN_MAX):
+        raise ValidationException("Class name must be between %s and %s characters" %
+                                  (DATASET_NAME_LEN_MIN, DATASET_NAME_LEN_MAX))
+
+    # Description (optional)
+    if "description" in dataset:
+        if not isinstance(dataset["description"], basestring):
+            raise ValidationException("Description must be a string.")
+        # TODO: Do we need to check the length there?
```
alastair: no

utils/dataset_validator.py
```diff
@@ -0,0 +1,116 @@
...
+        raise ValidationException("Dataset must have a name.")
+    if not isinstance(dataset["name"], basestring):
+        raise ValidationException("Class name must be a string.")
+    if not (DATASET_NAME_LEN_MIN < len(dataset["name"]) < DATASET_NAME_LEN_MAX):
+        raise ValidationException("Class name must be between %s and %s characters" %
+                                  (DATASET_NAME_LEN_MIN, DATASET_NAME_LEN_MAX))
+
+    # Description (optional)
+    if "description" in dataset:
+        if not isinstance(dataset["description"], basestring):
```
alastair: This check is for if people submit via the api, right? I wonder if we need the string checks. We could just cast to a string. This is for numbers and null, right?

utils/dataset_validator.py
```diff
@@ -0,0 +1,116 @@
...
+        ValidationException: A general exception for validation errors.
+        IncompleteDatasetException: Raised in cases when one of "completeness"
+            requirements is not satisfied.
+    """
+    if not isinstance(dataset, dict):
+        raise ValidationException("Dataset must be a dictionary.")
+    _check_dict_structure(dataset, ["name", "description", "classes", "public"])
+
+    # Name
+    if "name" not in dataset:
```
alastair: can we be more generic here - reusing the stuff you have in `_check_dict_structure` where you check for unexpected items. Just raise in there also if the expected items are not there.

utils/dataset_validator.py
```diff
@@ -0,0 +1,116 @@
...
+def _validate_classes(classes):
+    if not isinstance(classes, list):
+        raise ValidationException("Classes need to be in a list.")
+    for cls in classes:
+        _validate_class(cls)
+
+
+def _validate_class(cls):
+    if not isinstance(cls, dict):
+        raise ValidationException("Class must be a dictionary.")
```
alastair: can we show what class should be the dictionary? Does it make sense to render the whole thing in the error message (would be huge), could always say "element 4"

utils/dataset_validator.py
```diff
@@ -0,0 +1,116 @@
...
+        # TODO: Do we need to check the length there?
+
+    # Classes
+    if "classes" not in dataset:
+        raise ValidationException("Dataset must have a list of classes.")
+    _validate_classes(dataset["classes"])
+
+    # Publicity
+    if "public" not in dataset:
+        raise ValidationException("You need to specify if dataset is public or not.")
```
alastair: can we be consistent in these messages? `field "public" not present` or similar

alastair:
This is fine, I just want a little more consistency in the content of the error messages, and describing in more detail where the problem is. If you need the changes, you can merge and update the other stuff later

alastair:
How will we show these errors in the API compared to in the dataset editor? Perhaps API errors need to have more detail ("class Rock, element 3 is not a UUID"), but the dataset editor should highlight the bad item with just "not a UUID"

utils/dataset_validator.py
gentlecat: I don't think I like making modifications to a dataset here. This just checks if it has exactly the right format.

gentlecat:
I made the messages more informative and consistent. It would be nice to have a better integration with the UI for this, but I don't have time to implement it right now.

PR #167, opened 2016-03-01
Badjson

alastair:

Postgres has very strict utf-8 handling, and we get funny errors like 
```
DataError: (psycopg2.DataError) invalid input syntax for type json
LINE 3:                    VALUES (1, '{"lowlevel":{"average_loudnes...
                                      ^
DETAIL:  Unicode low surrogate must follow a high surrogate.
CONTEXT:  JSON data, line 1: ...te":44100},"tags":{"album":["IU...IM"],"artist":[...
```

Catch `DataError` and return http 400.


alastair:
created against master, should be against hlsplit

PR #168, opened 2016-03-01
AB-55: Reject submissions with invalid-according-to-postgres json

alastair:

Postgres has very strict utf-8 handling, and we get funny errors like 
```
DataError: (psycopg2.DataError) invalid input syntax for type json
LINE 3:                    VALUES (1, '{"lowlevel":{"average_loudnes...
                                      ^
DETAIL:  Unicode low surrogate must follow a high surrogate.
CONTEXT:  JSON data, line 1: ...te":44100},"tags":{"album":["IU...IM"],"artist":[...
```

Catch `DataError` and return http 400.

### DIFF
--- a/db/data.py
+++ b/db/data.py
@@ -8,6 +8,7 @@ import db
 import db.exceptions
 
 from sqlalchemy import text
+import sqlalchemy.exc
 
 _whitelist_file = os.path.join(os.path.dirname(__file__), "tagwhitelist.json")
 _whitelist_tags = set(json.load(open(_whitelist_file)))
@@ -124,7 +125,11 @@ def submit_low_level_data(mbid, data):
         )
 
     # The data looks good, lets see about saving it
-    write_low_level(mbid, data)
+    try:
+        write_low_level(mbid, data)
+    except sqlalchemy.exc.DataError as e:
+        raise db.exceptions.BadDataException(
+                "data is badly formed")
 
 def insert_version(connection, data, version_type):
     # TODO: Memoise sha -> id
--- a/db/data.py
+++ b/db/data.py
@@ -125,11 +125,7 @@ def submit_low_level_data(mbid, data):
         )
 
     # The data looks good, lets see about saving it
-    try:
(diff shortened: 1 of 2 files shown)


db/data.py
```diff
@@ -124,7 +125,11 @@ def submit_low_level_data(mbid, data):
         )
 
     # The data looks good, lets see about saving it
-    write_low_level(mbid, data)
+    try:
+        write_low_level(mbid, data)
+    except sqlalchemy.exc.DataError as e:
+        raise db.exceptions.BadDataException(
+                "data is badly formed")
```
alastair: I'm not sure what we should say here. I don't really want to expose the word `postgres` in this message

db/data.py
```diff
@@ -124,7 +125,11 @@ def submit_low_level_data(mbid, data):
         )
 
     # The data looks good, lets see about saving it
-    write_low_level(mbid, data)
+    try:
+        write_low_level(mbid, data)
+    except sqlalchemy.exc.DataError as e:
```
alastair: Should we catch all this exception or just the specifc unicode one which we expect? Should it be caught here or in `write_low_level`?

db/data.py
alastair: this message is returned to the client

db/data.py
gentlecat: This doesn't seem like the right place to do that. I'd rather catch exceptions inside `write_low_level`, that's where queries are performed.

gentlecat:
@brainzbot, test this please.

alastair:
:scream:

alastair:
@gentlecat better?

gentlecat:
Yep. There's some issue with dumps test. Is it related to the change?

alastair:
Nah, this is the failing dump from ages ago that we @ignored (this patch is
really old)

gentlecat:
OK. :shipit:

alastair:
merged to master

PR #169, opened 2016-03-02
AB-79: Indication of outdated evaluation results

kartikgupta0909:
Comparing the timestamp for dataset and job to check if the results are valid.

### DIFF
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -151,6 +151,7 @@ def get(id):
         )
         if result.rowcount > 0:
             row = dict(result.fetchone())
+ 	    print row
             row["classes"] = _get_classes(row["id"])
             return row
         else:
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -115,7 +115,7 @@ def update(dataset_id, dictionary, author_id):
             dictionary["description"] = None
 
         connection.execute("""UPDATE dataset
-                          SET (name, description, public, author) = (%s, %s, %s, %s)
+                          SET (name, description, public, author, last_edited) = (%s, %s, %s, %s, now())
                           WHERE id = %s""",
                        (dictionary["name"], dictionary["description"], dictionary["public"], author_id, dataset_id))
 
@@ -144,7 +144,7 @@ def get(id):
     """
     with db.engine.connect() as connection:
         result = connection.execute(
-            "SELECT id::text, name, description, author, created, public "
+            "SELECT id::text, name, description, author, created, public ,last_edited "
             "FROM dataset "
             "WHERE id = %s",
             (str(id),)
@@ -201,7 +201,6 @@ def get_by_user_id(user_id, public_only=True):
         for row in result:
             datasets.append(dict(row))
         return datasets
-
 
 def delete(id):
     """Delete dataset with a specified ID."""
--- a/webserver/static/js/eval-jobs-viewer.js
+++ b/webserver/static/js/eval-jobs-viewer.js
@@ -107,6 +107,7 @@ var EvaluationJobsViewer = React.createClass({
                         status={active_job.status}
                         statusMsg={active_job.status_msg}
                         result={active_job.result}
+                        outdated={active_job.outdated}
                         onReturn={this.handleReturn} />
                 );
             }
@@ -134,6 +135,7 @@ var JobList = React.createClass({
                         id={cls.id}
                         created={cls.created}
                         status={cls.status}
+                        outdated={cls.outdated}
                         onViewDetails={this.props.onViewDetails} />
                 );
             }.bind(this));
@@ -165,6 +167,7 @@ var JobRow = React.createClass({
         id: React.PropTypes.string.isRequired,
         created: React.PropTypes.string.isRequired,
         status: React.PropTypes.string.isRequired,
+        outdated: React.PropTypes.string.isRequired,
         onViewDetails: React.PropTypes.func.isRequired
     },
     handleViewDetails: function (event) {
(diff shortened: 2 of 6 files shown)


webserver/static/js/eval-jobs-viewer.js
```diff
@@ -107,6 +107,7 @@ var EvaluationJobsViewer = React.createClass({
                         status={active_job.status}
                         statusMsg={active_job.status_msg}
                         result={active_job.result}
+			valid={active_job.valid}
```
alastair: we use spaces to indent, not tabs

db/dataset.py
```diff
@@ -144,13 +144,14 @@ def get(id):
...
         result = connection.execute(
-            "SELECT id::text, name, description, author, created, public "
+            "SELECT id::text, name, description, author, created, public ,last_edited "
             "FROM dataset "
             "WHERE id = %s",
             (str(id),)
         )
         if result.rowcount > 0:
             row = dict(result.fetchone())
+ 	    print row
```
alastair: :scream:

webserver/static/js/eval-jobs-viewer.js
```diff
@@ -237,10 +239,22 @@ var JobDetails = React.createClass({
                 </div>;
                 break;
             case JOB_STATUS_DONE:
-                status = <div className="alert alert-success">
-                    This evaluation job has been completed on {this.props.updated}.
-                    You can find results below.
-                </div>;
+		
+		switch (this.props.valid) {
+		case '1':
```
alastair: see the way we do the job status above, we define some constants, e.g. `JOB_STATUS_DONE`, so that it's easier to read. You could do the same here - `JOB_DATASET_VALID`, `_NOTVALID`?
I'm not sure valid is the terminology we want to use here. I guess you want to say "the results are still valid", but the word is a little confusing because it can also mean things like "not made correctly". What about using `OUTDATED`?

webserver/views/datasets.py
```diff
@@ -41,8 +41,13 @@ def eval_jobs(dataset_id):
     ds = get_dataset(dataset_id)
     jobs = db.dataset_eval.get_jobs_for_dataset(ds["id"])
     # TODO(roman): Remove unused data ("confusion_matrix", "dataset_id").
+    last_edited_time = ds["last_edited"]
     for job in jobs:
         if "result" in job and job["result"]:
+	    cmp_time = max(job["created"],last_edited_time)
```
alastair: in function calls we put a space after commas - `max(job["created"], last_edited_time)`

webserver/views/datasets.py
```diff
@@ -41,8 +41,13 @@ def eval_jobs(dataset_id):
...
+            job["valid"] = '0'
```
alastair: Same as your other patch - we should use booleans here if it can only have 2 values

webserver/views/datasets.py
```diff
@@ -41,8 +41,13 @@ def eval_jobs(dataset_id):
...
+	    if(cmp_time == job["created"]):
```
alastair: python doesn't need `(` for `if` statements

db/dataset.py
```diff
@@ -115,7 +115,7 @@ def update(dataset_id, dictionary, author_id):
             dictionary["description"] = None
 
         connection.execute("""UPDATE dataset
-                          SET (name, description, public, author) = (%s, %s, %s, %s)
+                          SET (name, description, public, author, last_edited) = (%s, %s, %s, %s,current_timestamp)
```
alastair: watch the space after the comma here too. I tend to use `now()` instead of `current_timestamp`. they both mean the same thing

webserver/static/js/eval-jobs-viewer.js
```diff
@@ -237,10 +239,22 @@ var JobDetails = React.createClass({
...
+                    status = <div className="alert alert-success">
+                        This evaluation job has been completed on {this.props.updated}.
+                        You can find results below.
+                    </div>;
+		    break;	
+		case '0':
+		    status = <div className="alert alert-success">
+                        This evaluation job has been completed on {this.props.updated}.
+                         You can find results below.But these results are outdated since the dataset was edited after the job was initiated.
```
alastair: Space after the .

alastair: It would also be a good idea to show an indication on the job list:

screen shot 2016-03-02 at 10 50 20 pm

perhaps instead of saying [Done], for outdated jobs we can say [Done, Outdated], in yellow

alastair: For this sentence, how about saying 

> The dataset has been changed since this job was run, so the results may be out of date.

webserver/views/datasets.py
```diff
@@ -41,8 +41,13 @@ def eval_jobs(dataset_id):
...
+		job["valid"] = '1'
```
alastair: After looking at this statement again, it could be written more clearly. How about
```python
job["outdated"] = False
if last_edited_time > job["created"]:
    job["outdated"] = True
```

or even

```python
job["outdated"] = last_edited_time > job["created"]
```

webserver/views/datasets.py
```diff
@@ -183,7 +188,7 @@ def edit(dataset_id):
             "datasets/edit.html",
             mode="edit",
             dataset_id=str(dataset_id),
-            dataset_name=ds["name"],
```
gentlecat: You didn't need to remove the comma. Useful tip I've seen before: http://dave.cheney.net/2014/10/04/that-trailing-comma. :)

gentlecat:
Is this pull request a substitute for #155? Old one should be closed then.

gentlecat:
See #170.

PR #170, opened 2016-03-04
Changed the schema for dataset table and worked on AB-79

kartikgupta0909:
Made the Suggested changes for AB-79, please neglect the previous pull requests.
1.Changed the schema for dataset
2.Compared the last_edited timestamp of the dataset with the created time stamp of the job
3.Created a boolean variable outdated
4.Use the value of this variable to generate useful message

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -1,4 +1,3 @@
-
 BEGIN;
 
 CREATE TABLE lowlevel (
@@ -114,4 +113,3 @@ CREATE TABLE dataset_snapshot (
 );
 
 COMMIT;
-
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -1,3 +1,4 @@
+
 BEGIN;
 
 CREATE TABLE lowlevel (
@@ -78,7 +79,8 @@ CREATE TABLE dataset (
   description TEXT,
   author      INT NOT NULL, -- FK to user
   public      BOOLEAN NOT NULL,
-  created     TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  created     TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  last_edited TIMESTAMP WITH TIME ZONE DEFAULT NOW()
 );
 
 CREATE TABLE dataset_class (
@@ -112,3 +114,4 @@ CREATE TABLE dataset_snapshot (
 );
 
 COMMIT;
+
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -101,7 +101,7 @@ def create_from_dict(dictionary, author_id):
 
             # Removing duplicate recordings
             cls["recordings"] = list(set(cls["recordings"]))
-     
+
             for recording_mbid in cls["recordings"]:
                 connection.execute("INSERT INTO dataset_class_member (class, mbid) VALUES (%s, %s)",
                                (cls_id, recording_mbid))
@@ -147,7 +147,7 @@ def get(id):
     """
     with db.engine.connect() as connection:
         result = connection.execute(
-            "SELECT id::text, name, description, author, created, public ,last_edited "
+            "SELECT id::text, name, description, author, created, public, last_edited "
             "FROM dataset "
             "WHERE id = %s",
             (str(id),)
@@ -204,6 +204,7 @@ def get_by_user_id(user_id, public_only=True):
         for row in result:
             datasets.append(dict(row))
         return datasets
+
 
 def delete(id):
     """Delete dataset with a specified ID."""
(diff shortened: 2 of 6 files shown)


admin/sql/create_tables.sql
```diff
@@ -1,3 +1,4 @@
+
```
alastair: in a few places you have added some extra whitespace. You should be careful that you, or your editor doesn't do this. See if there is a setting that might be causing this

db/dataset.py
```diff
@@ -101,7 +101,7 @@ def create_from_dict(dictionary, author_id):
 
             # Removing duplicate recordings
             cls["recordings"] = list(set(cls["recordings"]))
-
+
```
alastair: whitespace?

db/dataset.py
```diff
@@ -147,7 +147,7 @@ def get(id):
     """
     with db.engine.connect() as connection:
         result = connection.execute(
-            "SELECT id::text, name, description, author, created, public "
+            "SELECT id::text, name, description, author, created, public ,last_edited "
```
alastair: be careful about formatting here, it should be `public, last_edited`

db/dataset.py
```diff
@@ -205,7 +205,6 @@ def get_by_user_id(user_id, public_only=True):
             datasets.append(dict(row))
         return datasets
 
-
```
alastair: you've deleted a line which already existed the PEP8 formatting guide suggests 2 newlines between functions. https://www.python.org/dev/peps/pep-0008/

webserver/static/scripts/datasets/eval-jobs-viewer.js
```diff
@@ -1,3 +1,4 @@
+/** @jsx React.DOM */
```
alastair: @gentlecat says that this is a requirement for older versions of react. I don't think we need it. @gentlecat can you confirm?

webserver/static/scripts/datasets/eval-jobs-viewer.js
```diff
@@ -184,7 +188,14 @@ var JobRow = React.createClass({
                 status = <span className="label label-danger">Failed</span>;
                 break;
             case JOB_STATUS_DONE:
-                status = <span className="label label-success">Done</span>;
+                switch (this.props.outdated) {
```
alastair: if `this.props.outdated` is a boolean, you could just do:
```javascript
if (this.props.outdated) {
  status = ...
} else {
  status = ...
}
```

webserver/static/scripts/datasets/eval-jobs-viewer.js
```diff
@@ -206,6 +217,7 @@ var JobDetails = React.createClass({
         created: React.PropTypes.string.isRequired,
         updated: React.PropTypes.string.isRequired,
         status: React.PropTypes.string.isRequired,
+	outdated: React.PropTypes.string.isRequired,
```
alastair: This looks like a tab

alastair:
This patch is looking really good. Just a few small formatting issues. You can make one more commit and push it to this branch, and we'll be able to merge it.

webserver/static/scripts/datasets/eval-jobs-viewer.js
gentlecat: https://facebook.github.io/react/blog/2014/10/16/react-v0.12-rc1.html#the-jsx-pragma-is-gone

db/dataset.py
gentlecat: http://pep8.org/#blank-lines

alastair:
great :+1: 

We only need 2 more things; 
* Because the current AcousticBrainz website already has data in this table, we need an sql update file, since we can't drop and recreate it. See this example here: https://github.com/metabrainz/acousticbrainz-server/commit/7cadb2b072762b62075355d4399a9118e5353c8f (`admin/updates/update-2.sql`). This should be in the form of `ALTER TABLE x ADD COLUMN y`, so that if you ran `create_tables` before this patch was merged, you can run the update script to make the tables consistent with the setup scripts.  You don't need to change the version/schema number for this, because this is a private table.
* When you do the update for rows that already exist, think carefully about what the value of the `last_updated` field should be. I don't think it should be `now()`, what do you think?
* We should have some tests. I would like to see one that creates a dataset, updates it, and then checks that the update time is after the time when the dataset was first created

alastair:
Be careful about your git commit messages. There are some generally accepted recommendations. For example, read this blog post: http://chris.beams.io/posts/git-commit/

alastair:
This is merged, thanks!
See this nice outdated evaluation:
http://acousticbrainz.org/datasets/61265979-235e-42b9-9a99-243e600275e3/evaluation

I made some small changes to formatting in your test. Please take a look at them to see how we like to format code, especially parameters in functions:
https://github.com/metabrainz/acousticbrainz-server/commit/88f22dc6a2bf55a4db69581c9f9294874b275d26

PR #171, opened 2016-03-06
Fix and improve displaying of controls on user page

gentlecat:


webserver/views/user.py
```diff
@@ -10,8 +10,9 @@
 
 @user_bp.route("/<musicbrainz_id>")
 def profile(musicbrainz_id):
-    if current_user.is_authenticated and \
-       current_user.musicbrainz_id == musicbrainz_id:
```
gentlecat: Old version breaks when username in the URL doesn't have letters in the same case as the original username. For example, when logged in I see controls at `/user/Gentlecat`, but not at `/user/gentlecat`.

webserver/views/user.py
```diff
@@ -10,8 +10,9 @@
 
 @user_bp.route("/user/<musicbrainz_id>")
 def profile(musicbrainz_id):
-    if current_user.is_authenticated and \
-       current_user.musicbrainz_id == musicbrainz_id:
```
gentlecat: Old version breaks when username in the URL doesn't have letters in the same case as the original username. For example, when logged in I see controls at `/user/Gentlecat`, but not at `/user/gentlecat`.

PR #174, opened 2016-03-18
Improvements for JS related to datasets

gentlecat:


alastair:
Looks good. Added one more `if (isMounted())` guard around the queries in the viewer, not the editor. This code is the same, should we be looking at moving it out to a helper?

PR #175, opened 2016-03-29
Add dataset information for default classifier models

dbogdanov:
I have not tested a rendered HTML yet. 

We should also host ```rosamerica_mp3_filelist.yaml``` file somewhere, and link to in in the description. The same can be done for the rest of music collections to provide at least partial information about the tracks.


alastair:
Merged, looks good thanks

PR #176, opened 2016-03-29
AB-66: Admin section

gentlecat:
*Fixes http://tickets.musicbrainz.org/browse/AB-66.*


db/user.py
```diff
@@ -13,8 +13,11 @@ def create(musicbrainz_id):
 def get(id):
     """Get user with a specified ID (integer)."""
     with db.engine.connect() as connection:
-        result = connection.execute('SELECT id, created, musicbrainz_id FROM "user" WHERE id = %s',
-                       (id,))
+        result = connection.execute("""
```
alastair: I've started using `text` for multiline queries, which allows us to add placeholders with `:id` and have the second parameter to `execute` a dictionary `{"id": myid}` which I think is a lot easier to reason about

gentlecat: Updated.

webserver/__init__.py
```diff
@@ -74,4 +74,10 @@ def create_app():
     app.register_blueprint(user_bp)
     app.register_blueprint(datasets_bp, url_prefix='/datasets')
 
+    # Admin section
+    from flask_admin import Admin
+    from webserver.admin import views as admin_views
```
alastair: can we make this themed like the rest of the site?

alastair:
This is a good start, but I'll hold off on completing the review until we have some actual functionality. I'd like to see 
* A user editor, which lets us make people admin
* A dataset viewer which lets us delete datasets, jobs, and change job states (we should reuse stuff from https://github.com/metabrainz/acousticbrainz-server/pull/165)

gentlecat:
I can add an admin editor, but dataset stuff should probably be separate from this. I added this to be able to start working on dataset challenges.

webserver/__init__.py
gentlecat: Yes, but it requires overriding built-in templates, which I don't want to do too much. From the [docs](https://flask-admin.readthedocs.org/en/latest/introduction/#overriding-the-built-in-templates):
> To take full control over the style and layout of the admin interface, you can override all of the built-in templates. Just keep in mind that the templates will change slightly from one version of Flask-Admin to the next, so once you start overriding them, you need to take care when upgrading your package version.

Besides this is not meant for public view.

gentlecat:
Admin editor is there. You can add users from the website or using *manage.py* script.

PR #178, opened 2016-04-10
API improvements

gentlecat:
Decided to organize all the API endpoints into a single URL prefix: `/api/v1/...`. This seems a bit cleaner and it will be easy to extend in future (datasets, challenges, etc.)

I also included basic structure for Sphinx docs. Example of a build is at https://acousticbrainz.readthedocs.org/en/api/api.html.

Things I plan to add next are:
* Datasets (CRUD)
* Evaluation queue (add dataset, remove dataset, get status, get queue)


gentlecat:
@alastair, if you want access to https://readthedocs.org/projects/acousticbrainz/, give me your username.

alastair:
My rtd username is `alastairp`

gentlecat:
Added you.

webserver/templates/data/summary.html
```diff
@@ -98,13 +98,13 @@ <h3 class="text-center">
     <ul>
       <li>
         <a href="{{ url_for('data.view_low_level', mbid=metadata['mbid'], n=offset) }}">Low-level</a>
-        (<a href="{{ url_for('api.get_low_level', mbid=metadata['mbid'], n=offset) }}">raw JSON</a>)
+        (<a href="{{ url_for('api_v1_core.get_low_level', mbid=metadata['mbid'], n=offset) }}">raw JSON</a>)
       </li>
       {% if data.highlevel %}
         <li>
           {# FIXME(roman): Offset might not point to the right high-level data there. #}
```
alastair: I think this is fixed now with the changes I made to the hl load with the database changes: https://github.com/metabrainz/acousticbrainz-server/blob/master/db/data.py#L334

webserver/views/api/v1/core.py
```diff
@@ -6,21 +6,21 @@
...
+    """Endpoint for getting number of submissions for a given recording."""
     return jsonify({
         'mbid': mbid,
-        'count': count_lowlevel(mbid),
+        'count': count_lowlevel(str(mbid)),
     })
 
 
-@api_bp.route("/<string:mbid>/low-level", methods=["GET"])
+@bp_core.route("/<uuid:mbid>/low-level", methods=["GET"])
```
alastair: Should we update the docs on https://acousticbrainz.org/data too?

You've changed the argument from string to uuid. Is this because it's invalid to make a  query without a mbid?

webserver/views/api/v1/core.py
```diff
@@ -29,14 +29,14 @@ def get_low_level(mbid):
...
+        return jsonify(db.data.load_low_level(str(mbid), offset))
     except NoDataFoundException:
         raise webserver.exceptions.APINotFound("Not found")
 
 
-@api_bp.route("/<string:mbid>/high-level", methods=["GET"])
+@bp_core.route("/<uuid:mbid>/high-level", methods=["GET"])
 @crossdomain()
 def get_high_level(mbid):
     """Endpoint for fetching high-level data.
```
alastair: After reading the documentation website, I'm not sure about this style of writing. Yes, this function "implements an endpoint", but when you read it in the documentation I think it would read better as "Get high level data for a given MusicBrainz Recording id" (or similar) instead. What do you think?

webserver/views/api/v1/core.py
```diff
@@ -45,14 +45,14 @@ def get_high_level(mbid):
     an offset as a query parameter in the form of ?n=x, where x is an integer
     starting from 0
```
alastair: "Use the `/mbid/count/` (link) endpoint to get the total number of documents for this mbid."

alastair:
How do we tell people about the deprecated `/mbid/low-level` endpoints? blog post?

alastair:
Apart from the changes, this looks fine

webserver/views/api/v1/core.py
gentlecat: I'd say move this page to sphinx completely.

It's supposed to be UUID (MBID). Now it's more self-documenting, looks more correct is these docs I added. I'm just doing the conversion below, as you can see.

webserver/views/api/v1/core.py
gentlecat: Yeah, for sure. I'll make an update.

webserver/templates/data/summary.html
gentlecat: Right. Removed. 💥

webserver/views/api/v1/core.py
gentlecat: Done. https://acousticbrainz.readthedocs.org/en/api/api.html#endpoint-reference

gentlecat:
Yeah, I guess blog post is fine. It's probably worth noting that we are still working on some parts and that they might change. I'm mostly worried about submissions from AcousticBrainz Client.

webserver/views/api/v1/core.py
alastair: Good point. We shouldn't remove this page completely (because it has other things like the highlevel model descriptions), but we can link specific API stuff to the RTD docs. Should we also move the example files there? Or perhaps we can just link to a json view of an existing document anyway.

gentlecat: Linking is probably better. I don't think this small scrollable window with all this data is very useful.

alastair:
LGTM. Please create a draft blog post explaining the API changes, to go live when we turn on HTTPS.

gentlecat:
Alright. I'll write something.

gentlecat:
@alastair, https://gist.githubusercontent.com/gentlecat/0eec763e892a9d942d159ab6f87225bb/raw/f2f4166cd5c7d0826a08d29f988567f32665b8de/draft.txt

PR #180, opened 2016-04-11
Show updates when using gulp-watch

gentlecat:
This is pretty much a copy of commit in MusicBrainz server: https://github.com/metabrainz/musicbrainz-server/commit/2c34ea.

### DIFF
--- a/webserver/static/gulpfile.js
+++ b/webserver/static/gulpfile.js
@@ -118,7 +118,6 @@ gulp.task('styles', function () {
 gulp.task('scripts', buildScripts);
 
 gulp.task('watch', ['styles', 'scripts'], function () {
-  let watch = require('gulp-watch');
 
   gulp.watch(path.resolve(STATIC_DIR, '**/*.less'), ['styles']);
 
@@ -144,8 +143,9 @@ gulp.task('watch', ['styles', 'scripts'], function () {
     }
   }
 
+  let watch = require('gulp-watch');
   watch(path.resolve(SCRIPTS_DIR, '**/*.js'), function (file) {
-    _.each(CACHED_BUNDLES, function (bundle, resourceName) {
+    CACHED_BUNDLES.forEach(function (bundle, resourceName) {
       rebundle(bundle, resourceName, file);
     });
   });
--- a/webserver/static/gulpfile.js
+++ b/webserver/static/gulpfile.js
@@ -120,14 +120,7 @@ gulp.task('scripts', buildScripts);
 gulp.task('watch', ['styles', 'scripts'], function () {
   let watch = require('gulp-watch');
 
-  watch(path.resolve(STATIC_DIR, '**/*.less'), function () {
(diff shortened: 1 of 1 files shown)


webserver/static/gulpfile.js
```diff
@@ -113,13 +112,22 @@ function buildScripts() {
...
+gulp.task('styles', function () {
+  return buildStyles(writeManifest);
+});
 gulp.task('scripts', buildScripts);
 
 gulp.task('watch', ['styles', 'scripts'], function () {
   let watch = require('gulp-watch');
 
-  watch(path.resolve(STATIC_DIR, '**/*.less'), buildStyles);
+  watch(path.resolve(STATIC_DIR, '**/*.less'), function () {
```
alastair: I'm pretty sure you can provide a gulp task as a callback to watch, instead of a function, so by setting it to `['styles']` you should get the normal gulp output without having to reproduce it in another callback

alastair: https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpwatchglob-opts-tasks

gentlecat: That doesn't work for some reason.

alastair: I worked this out, it's because it's the `watch` method on `gulp`, not part of `gulp-watch`

webserver/static/gulpfile.js
```diff
@@ -113,13 +112,14 @@ function buildScripts() {
...
+gulp.task('styles', function () {
+  return buildStyles(writeManifest);
+});
 gulp.task('scripts', buildScripts);
 
 gulp.task('watch', ['styles', 'scripts'], function () {
-  let watch = require('gulp-watch');
 
-  watch(path.resolve(STATIC_DIR, '**/*.less'), buildStyles);
+  gulp.watch(path.resolve(STATIC_DIR, '**/*.less'), ['styles']);
```
alastair: This gives output like this now when you save a js file:
```
[15:56:38] Starting 'styles'...
[15:56:38] Finished 'styles' after 3.05 ms
```
just like when you run `gulp styles`

webserver/static/gulpfile.js
```diff
@@ -135,12 +135,17 @@ gulp.task('watch', ['styles', 'scripts'], function () {
...
+        writeManifest();
+        process.stdout.write('done.\n');
+      });
     }
   }
 
+  let watch = require('gulp-watch');
   watch(path.resolve(SCRIPTS_DIR, '**/*.js'), function (file) {
-    _.each(CACHED_BUNDLES, function (bundle, resourceName) {
+    CACHED_BUNDLES.forEach(function (bundle, resourceName) {
```
alastair: this wasn't working for me, but when I changed from `_` to `.forEach` it did. No idea what's going on

gentlecat:
LGTM

PR #181, opened 2016-04-18
API key generation

gentlecat:
New table and UI.

### DIFF
--- a/webserver/static/scripts/profile.js
+++ b/webserver/static/scripts/profile.js
@@ -1,0 +1,33 @@
+const $ = require("jquery");
+
+$(document).ready(function () {
+    let apiKey = $("#api-key");
+    if (apiKey) {  // Only when key is displayed (user is viewing their own page)
+
+        // Not showing confirmation dialog if there's no active key (key value is not displayed).
+        let ignoreConfirmation = apiKey.css("display") === "none";
+
+        $("#btn-generate-api-key").click(function () {
+            if (ignoreConfirmation || confirm('Are you sure you want to generate new API key?\n' +
+                    'Current key will be revoked!')) {
+                $.ajax({
+                    type: "POST",
+                    url: "/user/generate-api-key",
+                    success: function (data) {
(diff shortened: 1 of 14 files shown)


webserver/static/scripts/profile.js
```diff
@@ -0,0 +1,33 @@
...
+
+$(document).ready(function () {
+    let apiKey = $("#api-key");
+    if (apiKey) {  // Only when key is displayed (user is viewing their own page)
+
+        // Not showing confirmation dialog if there's no active key (key value is not displayed).
+        let ignoreConfirmation = apiKey.css("display") === "none";
+
+        $("#btn-generate-api-key").click(function () {
+            if (ignoreConfirmation || confirm('Are you sure you want to generate new API key?\n' +
```
alastair: I really don't like javascript popups. Can we do it with new pages or javascript to change the contents of the page?

alastair:
What's the purpose of API keys? Who will be using them?
 * Website frontend
 * Other apps who need to read/write user information

Some questions:
 * Is it OK for a user to have only 1 key?
 * Do all keys have the same permissions? r/w?
 * Should we just take the plunge and put oauth in like we have for CB?

webserver/static/scripts/profile.js
gentlecat: Yeah, I guess we could, but that's more work which I'm not sure is necessary for what we are doing. What's wrong with popups?

gentlecat:
Our website will keep using a cookie for authentication.

> Other apps who need to read/write user information

Just datasets for now.

> Is it OK for a user to have only 1 key?

There might be a need for multiple keys if you have several apps and don't want to reuse the same key. I think for now one should be enough. In future, if needed, it will be straightforward to add support for multiple.

webserver/static/scripts/profile.js
gentlecat: Can also remove it completely and just send the request.

PR #182, opened 2016-04-18
Basic API for creating, retrieving, and deleting datasets

gentlecat:


gentlecat:
Currently classes in a dataset don't have a public ID that can be referenced in a URL. Maybe this should change with http://tickets.musicbrainz.org/browse/AB-121?

webserver/decorators.py
```diff
@@ -1,9 +1,25 @@
...
+from werkzeug.exceptions import Unauthorized
 from six import string_types
 
 
+def auth_required(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        api_key = request.args.get('key')
+        # TODO(roman): Add support for API keys when this is merged:
+        # https://github.com/metabrainz/acousticbrainz-server/pull/181
```
alastair: Done. Would you like me to merge this PR as-is, or would you like to add this support first?

webserver/errors.py
```diff
@@ -1,13 +1,12 @@
 from flask import render_template, jsonify
-import webserver.exceptions
+from webserver.views.api import exceptions as api_exceptions
```
alastair: I'm not a fan of renaming modules when importing them. Can we just refer to exceptions as `webserver.views.api.exceptions.APIError` ?

webserver/views/datasets.py
```diff
@@ -308,7 +308,11 @@ def edit(dataset_id):
 def delete(dataset_id):
     ds = get_dataset(dataset_id)
     if ds["author"] != current_user.id:
-        raise Unauthorized("You can't delete this dataset.")
+        return jsonify(
```
alastair: Why was this changed from the exception to a custom error message? Is it because you needed the `error=True` value in the returned data?

webserver/views/api/v1/datasets.py
```diff
@@ -0,0 +1,106 @@
...
+                "recordings": ["770cc467-8dde-4d22-bc4c-a42f91e"]
+            }
+
+
+    :resheader Content-Type: *application/json*
+    """
+    dataset_dict = request.get_json()
+    if not dataset_dict:
+        raise api_exceptions.APIBadRequest("Data must be submitted in JSON format.")
+    # TODO: Catch validation error from the next call
```
alastair: to be done before merge, or later?

webserver/decorators.py
gentlecat: Let's add it later. I have other things to work on currently.

webserver/errors.py
gentlecat: That's too long and just `exceptions` would be confusing in this context.

webserver/views/api/v1/datasets.py
gentlecat: Later. Not critical.

webserver/views/datasets.py
gentlecat: Because it was a regular werkzeug exception.

webserver/views/api/v1/datasets.py
```diff
@@ -0,0 +1,106 @@
...
+        .. sourcecode:: json
+
+            {
+                "name": "Happy",
+                "description": "Recordings that represent happiness.",
+                "recordings": ["770cc467-8dde-4d22-bc4c-a42f91e"]
+            }
+
+
+    :resheader Content-Type: *application/json*
```
gentlecat: Done.

webserver/views/datasets.py
alastair: ah, right :+1:. Can we replace this with [`APIUnauthorized`](https://github.com/metabrainz/acousticbrainz-server/blob/0461cbe92aab16718703e25f5df589e4110895ca/webserver/views/api/exceptions.py#L18)?

gentlecat: I'd rather not.

1. This is not an API.
2. If the way API exceptions output content changes, this might break the UI (and this is the only place this endpoint is used). So I'd rather keep it more explicit here than try to find mysterious bugs in future.

webserver/views/api/v1/datasets.py
```diff
@@ -0,0 +1,106 @@
...
+                }
+            ]
+        }
+
+    :reqheader Content-Type: *application/json*
+    :<json string name: *Required.* Name of the dataset.
+    :<json string description: *Optional.* Description of the dataset.
+    :<json boolean public: *Optional.* ``True`` to make dataset public, ``false`` to make it private. New datasets are
+        public by default.
+    :<json array classes: *Optional.* Array of objects containing information about classes to add into new dataset. For
```
alastair: Submitting json with no `public` or `classes` results in errors thrown on the server:

    ValidationException: Field `public` is missing from dataset dictionary.

alastair:
I added some implementation for api key auth and returning the validation error when creating a dataset.
I'm happy to merge if you have no problems with the additions.

webserver/login/__init__.py
```diff
@@ -33,6 +34,21 @@ def load_user(user_id):
...
+    if key:
+        parts = key.split(" ")
+        if len(parts) == 2 and parts[0] == "Token":
+            user = db.user.get_by_api_key(parts[1])
+        else:
+            raise Unauthorized
+    if user:
+        return User.from_dbrow(user)
+    else:
+        return None
```
gentlecat: This affects authentication outside of the API as well, right? We probably don't want that.

gentlecat:
Your changes apart from API key authentication look good.

webserver/login/__init__.py
alastair: It does, as I understand flask will either use its normal session auth, then if that doesn't work it'll use this (for the whole site). It shouldn't affect users browsing the main site (e.g. dataset page), since they won't be sending a header.
We could take a look at see if you can restrict this loader to just a particular blueprint, but I like this method better than a standard `@auth_required` decorator because it seamlessly integrates the `current_user` in the same way as normal browser auth.

gentlecat: Yeah, that seems to be cleaner. I just don't want API keys to be used for access to anything outside of the API (so main site).

alastair: I had a bit of a look at flask-login, it looks like it only works on the app level. There are a few things we could do here instead:

* Check in the `request_loader` the value of `request.path` and return `None` if we're not in the API blueprint (I don't like this)
* Set a flag that says we were authed by token (where? `flask.g`?) and check this flag in a decorator on non-api endpoints, denying access if it's set (probably replacing the flask `@login_required` decorator, but this means we modify non-api endpoints for the sake of api auth)
* Go back to doing auth in a specific decorator for the API (this means setting `flask_login.current_user` properly. Is that all we have to do to make a user appear logged in?)

I'm not sure we have to restrict web access to only session auth. As I understand we're already using session auth in the API anyway, when the website accesses an API endpoint with js. Do you have a specific reason why you don't like having the opposite too?
Theoretically the keys are supposed to be secure, and I don't think there's anything else that a user could do on the website which they can't do by the API. I don't mind having the possibility of token auth for the website, even if no one uses it.

PR #184, opened 2016-04-25
Use py.test as a test runner

gentlecat:
It behaves pretty similar to nose, and doesn't make much difference in our case.

To run all tests just use `py.test` command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,13 +15,8 @@ __pycache__/
 target/
 
 # Unit test / coverage reports
-/htmlcov
-/cover
-.tox/
+htmlcov
 .coverage
-.cache
-nosetests.xml
-coverage.xml
 
 # Logs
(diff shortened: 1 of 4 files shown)


alastair:
what's the reason for this? Because `unittest` is deprecated? Does it do everything that we need it to do?

gentlecat:
Not `unittest`, nose:

> *Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.*
> \- https://nose.readthedocs.org/en/latest/

And yes, it does what we've been using before (`unittest` with our modifications). It also has some other useful features, but I didn't look at them closely. http://pytest.org/

alastair:
:+1:

PR #186, opened 2016-05-04
Delete temporary files if hl calculation fails

alastair:
Whenever we return early from the hl calculation file we need
to make sure that we remove any temporary files that we created.
This includes:
 - input (lowlevel) file if we fail to write data to it
 - input and output (highlevel) file if we fail to run the extractor

Tests:
Ran manually with a `streaming_extractor_music_svm` file which was not `+x` (failed with `OSError`), and a dummy script which returned `1` (failed with `subprocess.CalledProcessError`). Without these changes, tmp files remained, with the changes they were cleaned up.
I'm not sure of the best way to test the `IOError`s for the input and output file. If these open/reads fail I think we have more serious problems on the server

### DIFF
--- a/hl_extractor/hl_calc.py
+++ b/hl_extractor/hl_calc.py
@@ -50,6 +50,8 @@ class HighLevel(Thread):
             f.close()
         except IOError:
             print("IO Error while writing temp file")
+            # If we return early, remove the ll file we created
+            os.unlink(name)
             return "{}"
 
         # Securely generate a temporary filename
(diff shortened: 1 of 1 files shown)


hl_extractor/hl_calc.py
```diff
@@ -62,21 +64,27 @@ def _calculate(self):
...
+            # output file that we created
+            os.unlink(out_file)
             return "{}"
-
-        fnull.close()
-        os.unlink(name)
+        finally:
+            # At this point we can remove the source file,
+            # regardless of if we failed or if we succeeded
+            fnull.close()
```
gentlecat: Can we use a context manager instead of this?

alastair: hmm, so we do 
```python
with open(devnull) as fnull:
    try:
        subprocess.check_call(, , fnull)
    except:
        blah
```
?
It seems a little odd to use it just for this file handle. Although, I guess all a file context manager does is close the file handle in a finally block (https://docs.python.org/2/library/contextlib.html#contextlib.closing). By doing it ourselves we save a level of indentation.

gentlecat: Yeah, sounds fine to keep it. Lately I started using these a lot just to make sure that I never forget to close anything.

hl_extractor/hl_calc.py
```diff
@@ -62,21 +64,27 @@ def _calculate(self):
...
+            os.unlink(name)
 
         try:
             f = open(out_file)
             hl_data = f.read()
             f.close()
-            os.unlink(out_file)
         except IOError:
             print("IO Error while removing temp file")
```
alastair: This `IOError` no longer gets thrown on *removing* the temp file, just when reading it.

gentlecat:
LGTM

PR #188, opened 2016-05-05
Create snapshot of a dataset when it's submitted for evaluation

gentlecat:


admin/updates/update-4.sql
```diff
@@ -0,0 +1,28 @@
...
+);
+
+ALTER TABLE dataset_snapshot ADD CONSTRAINT dataset_snapshot_pkey PRIMARY KEY (id);
+
+ALTER TABLE dataset_snapshot
+  ADD CONSTRAINT dataset_id_fk_dataset
+  FOREIGN KEY (dataset_id)
+  REFERENCES dataset (id);
+
+DELETE FROM dataset_eval_jobs;
```
gentlecat: Not quite sure how to handle existing evaluation jobs there. Can we remove them? Any suggestions on that?

gentlecat:
One problem: There might be a case where dataset that is associated with a snapshot is deleted and that snapshot has to stay. Then we have no way to linking it back to the author, which is not so good.

alastair:
Why might the snapshot have to stay? I wouldn't be against requiring a dataset to be present if we need the snapshot. Don't allow the user to delete it.

gentlecat:
For example when dataset is submitted for a challenge and a snapshot associated with it and evaluation job has to stay available. That way we make sure that evaluation results are reproducible.

One option is to archive/hide dataset instead of actually deleting it.

admin/updates/update-4.sql
alastair: we discussed this in person. This update script should be modified to create a snapshot for all existing datasets and update the columns in `dataset_eval_jobs` to point to the right things

alastair:
We discussed some things:

 - I don't think that we should let a user delete a dataset if it's been snapshotted and added to a challenge. Some possibilities here are to let them delete everything (snapshot, entry to contest, dataset) if they really want to, or just have a "hidden" flag in a dataset so that it exists but no longer shows up for the user


- We should merge the two snapshotting systems (this one, and the artist-filtered training/testing splits) into the single json blob. This requires functionality to do the artist filtering in the web, not in the eval script. I need to add this metadata to the webserver to do this.

admin/updates/update-4.sql
gentlecat: OK, that's done.

gentlecat:
@alastair, can you look at tests for new stuff that lists datasets? I can't figure out what's wrong.

gentlecat:
*Resolved conflicts.*

alastair:
Deleting a pending eval job leaves behind the snapshot. Should we also remove the snapshot, especially if it's not been submitted to a contest?

gentlecat:
Not sure yet.

db/dataset.py
```diff
@@ -192,3 +193,116 @@ def delete(id):
...
+        dictionary: {
+            "id": <ID of the snapshot>,
+            "dataset_id": <ID of the dataset that this snapshot is associated with>,
+            "created": <creation time>,
+            "data": <actual content of a snapshot (see `create_snapshot` function)>
+        }
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT id::text
```
alastair: We've been tending to use `::text` when getting uuids from the database, so I updated this

db/dataset.py
```diff
@@ -192,3 +193,116 @@ def delete(id):
...
+                 , dataset_id::text
+                 , data
+                 , created
+              FROM dataset_snapshot
+             WHERE id = :id
+        """), {"id": id})
+        row = result.fetchone()
+        if not row:
+            raise db.exceptions.NoDataFoundException("Can't find dataset snapshot with a specified ID.")
```
alastair: I see you have a `TODO` also to decide if we raise an exception on dataset `get`. My feeling for this is that we shouldn't have exceptions - they're for _exceptional_ circumstances, and data not existing isn't really exceptional in my view.
In other places in AB we don't raise exceptions here.
From my experience in other places (e.g. pymusicbrainzngs) raising exceptions is just a pain. None or an empty list tends to work better.

db/dataset.py
```diff
@@ -192,3 +193,116 @@ def delete(id):
...
+        return dict(row)
+
+def delete_snapshot(connection, snapshot_id):
```
alastair: I'm not sure what to do here. Our habit has been to have standalone methods which work in a transaction, but snapshot deletion should happen at the same time as dataset_eval_job deletion, therefore in the same transaction. We could copy the sql code into that method, but then that potentially leaves duplicate code, which is difficult to maintain.
I'm not sure what we should do here. Some options:
 * We currently only delete snapshots when we delete a pending eval job, so just move the sql there now and deal with this problem later
 * Come up with a style guide for these helper methods that can be called with a connection as an argument. Maybe `_delete_snapshot_c(connection, snapshot_id)` and then if we need the method publicly accessible we create `delete_snapshot` which creates a connection and calls this method.

db/dataset.py
```diff
@@ -192,3 +193,116 @@ def delete(id):
...
+    Args:
+        connection: an SQLAlchemy connection.
+        snapshot_id (string/uuid): ID of a snapshot.
+    """
+    query = sqlalchemy.text("""
+        DELETE FROM dataset_snapshot
+              WHERE id = :snapshot_id""")
+    connection.execute(query, {"snapshot_id": snapshot_id})
+
+def get_snapshots_for_dataset(dataset_id):
```
alastair: I added these methods but currently they're only used in a test. Not sure if they're going to be useful for snapshot stuff in general. Perhaps just move the sql query directly to the test for now?

db/test/test_dataset.py
```diff
@@ -139,6 +139,34 @@ def test_last_edited(self):
...
+        snapshots = dataset.get_snapshots_for_dataset(id)
+        self.assertEqual(len(snapshots), 0)
+
+        snap_id = dataset.create_snapshot(id)
+        snapshots = dataset.get_snapshots_for_dataset(id)
+        self.assertEqual(len(snapshots), 1)
+        self.assertEqual(snapshots[0]["id"], snap_id)
+
+        snap = dataset.get_snapshot(snap_id)
+        self.assertEqual(snap["data"], self.test_data)
```
alastair: I've left this `assert` failing, so that we can decide if we want this to be true.
`test_data` has a `"public": True` field which is not in the snapshot. The snapshot has `"id": n` fields in each class. Does it make sense to include/not include these fields in the snapshot, therefore making the dict equal, or are `dataset_class.id` values useful in the snapshot data?

db/dataset.py
gentlecat: I think it's fine to raise an exception in this case, as long as behavior is obvious and documented. That's how I've been doing it in code that was merged recently, and here I'm just trying to make it consistent throughout the `db` package. It also makes it easier to catch all database-related exceptions that might happen using the base class, which helps avoid having more boilerplate code.

> From my experience in other places (e.g. pymusicbrainzngs) raising exceptions is just a pain. None or an empty list tends to work better.

What are the issues exactly? And what happens if `None` or an empty list are "normal" values?

db/dataset.py
gentlecat: Before we decided to have "private" functions that accept connection and can be combined with other ones. If the same function is required to be a part of "public" interface, we create a wrapper function for it.

db/dataset.py
gentlecat: I'm not sure if it's a good idea to add stuff that is unused. Makes things even more confusing. If something changes we'll have to spend time updating these functions that are just ran by tests. And tests too, potentially.

db/test/test_dataset.py
gentlecat: This is a bug. There should be no IDs in classes.

PR #189, opened 2016-05-23
Endpoint for submitting data without MBID

hellska:
Added a function to check for low-level data duplicates
Added a skeleton for a function that generates uuids
Added the tests for the two functions
Modified the SANITY_CHECK_KEYS


gentlecat:
@brainzbot, add to whitelist.

db/data.py
```diff
@@ -460,3 +461,32 @@ def get_summary_data(mbid, offset=0):
...
     return summary
+
+def check_low_level_duplicates(data):
+    """
+        Check if data comes from an audio file already analysed
+        Uses the md5_encoded field
+    Args: data
+        A dictionary containing the low-level data extracted using essentia
+    """
+    if 'md5_encoded' in data['metadata']['audio_properties']:
```
alastair: If we check for this field in `SANITY_CHECK_KEYS`, perhaps we don't need to have a separate check here

db/test/test_data.py
```diff
@@ -263,6 +264,11 @@ def test_add_get_model(self):
 
     def test_get_summary_data(self):
         pass
+    
+    def test_check_low_level_duplicates(self):
```
alastair: you can use `@unittest.skip` to skip an incomplete test. It will report it as skipped in the output, so that we don't forget about finishing it

db/data.py
```diff
@@ -460,3 +461,32 @@ def get_summary_data(mbid, offset=0):
...
+        input_md5 = data['metadata']['audio_properties']['md5_encoded']
+        query = text("select data->'metadata'->'tags'->'musicbrainz_recordingid' from lowlevel_json where data->'metadata'->'audio_properties'->'md5_encoded' ? :md5 ")
```
alastair: Look at the way that we format SQL statements in this file, especially keywords in capitals, and the position of keywords, fields, and commas:
https://github.com/hellska/acousticbrainz-server/blob/e5ad20e9d4807abbe4be030c104a2a1531bcd9ab/db/data.py#L135-L138

alastair: You can get the recording ID from the lowlevel table here. You should join it on the id field of `lowlevel` and `lowlevel_json`

alastair: If we do a lookup on the md5 we will need to create an index on this field and add it to an SQL update file (in `admin/updates`)

alastair: the json syntax `data->'metadata'->'tags'->'musicbrainz_recordingid'` is valid, but a bit cumbersome. I learned recently that there is another format for doing this kind of child lookup:
http://www.postgresql.org/docs/9.5/static/functions-json.html
`data#>'{metadata,tags,musicbrainz_recordingid}'` should also work

admin/updates/update-2-add_gin_index_lowleve_json_data-md5_encoded.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+CREATE INDEX data_metadata_audio_properties_md5_encoded_idx_lowlevel_json ON lowlevel_json USING gin ((data#>'{metadata,audio_properties,md5_encoded}'));
```
gentlecat: Could use a line break here before `ON`.

You also should include this into https://github.com/metabrainz/acousticbrainz-server/blob/master/admin/sql/create_indexes.sql.

db/data.py
```diff
@@ -460,3 +461,40 @@ def get_summary_data(mbid, offset=0):
         pass
 
     return summary
+
+def check_low_level_duplicates(data):
+    """
+        Check if data comes from an audio file already analysed
+        Uses the md5_encoded field
+    Args: data
+        A dictionary containing the low-level data extracted using essentia
```
gentlecat: A bit of nitpicking. Most of the docstrings are formatted based on this https://google.github.io/styleguide/pyguide.html#Comments. So that docstring could look like this:
```
Checks if data with the same md5_encoded value has already been submitted.

Args:
    data: Submitted data.
```

gentlecat: And then the name of the function can be more explicit like `find_md5_duplicate`. Perhaps even accept just that hash and return existing ID if it was found. It should not create a new one, just that one thing for finding a duplicate.

db/test/test_data.py
gentlecat: Or you can just write a test that fails and finish it with the rest of implementation.

alastair: Yep, that's another valid option. `self.assertTrue(False)`

db/data.py
```diff
@@ -460,3 +462,66 @@ def get_summary_data(mbid, offset=0):
...
+def generate_datasetitem_uuidv3():
+    """Generate a random UUIDv4 for dataset items that are not present in the DB as recordings
+    
+    Returns: 
+        itemuuid: The generated random UUIDv3 assigned to a specific dataset item
+    """
+    try:
+        itemuuid = uuid4()
+        return str(itemuuid)
+    except:
```
gentlecat: You should specify which exception to catch.

gentlecat: Actually, I don't see a reason we need `try except` here.

db/test/test_data.py
```diff
@@ -260,10 +263,26 @@ def test_add_get_model(self):
...
 
-
     def test_get_summary_data(self):
         pass
-
+    
+    def test_find_md5_duplicates(self):
+        
+        self.assertEqual(db.data.find_md5_duplicates(self.test_md5), None)
+        # implement the test for successful query
```
gentlecat: Can you implement this?

db/data.py
```diff
@@ -460,3 +462,66 @@ def get_summary_data(mbid, offset=0):
...
+        raise "Error Generating the random uuid"
+    
+    return None
+    
+def generate_datasetitem_uuidv4(data):
```
gentlecat: It looks like names of these two functions are the opposite.

gentlecat:
Why do we need to generate a UUID based on a namespace and not just a random one?

db/data.py
```diff
@@ -460,3 +462,66 @@ def get_summary_data(mbid, offset=0):
...
+            # returns the retrieved mbid
+            return row[0]
+        else:
+            return None
+
+def generate_datasetitem_uuidv3():
+    """Generate a random UUIDv4 for dataset items that are not present in the DB as recordings
+    
+    Returns: 
+        itemuuid: The generated random UUIDv3 assigned to a specific dataset item
```
Freso: Does this generate UUIDv3 or v4?

hellska: The comment was wrong it generates UUIDv3.
The reaso to choose UUIDv3 is because it generates the same uuid with the same pair of namespace/string, it seems more consistent than a random uuid. I implemented both so we can choose the one that best fits project needs.

gentlecat: It might be better to add utility functions with implementation of the main feature. Keeping unused code around doesn't seem like a good idea.

hellska: Yes I agree that keeping unused code is a bad idea :/
So we keep the UUIDv4 already used in Acousticbrainz or we go for UUIDv3?

gentlecat: Depends on how you are going to use them. If it's for generating temporary IDs for submissions without MBID, I'd say use v4 (random).

admin/updates/update-2-add_gin_index_lowleve_json_data-md5_encoded.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
```
alastair: spelling mistake in the filename

db/data.py
```diff
@@ -460,3 +462,42 @@ def get_summary_data(mbid, offset=0):
...
+    Args: 
+        md5: The md5_encoded from the low-level data to check
+    
+    Returns:
+        The retrieved mbid if the md5 already exists
+        None if the md5 is not found
+    """
+
+    query = text("""
+                 SELECT lowlevel.mbid
```
alastair: please double-check the indentation of this sql statement. SQL keywords should be right alighed

db/data.py
```diff
@@ -460,3 +462,42 @@ def get_summary_data(mbid, offset=0):
...
+                 FROM lowlevel, lowlevel_json
```
alastair: in other places in the server we use a JOIN, instead of a FROM/WHERE. It means exactly the same, but for consistency we should always write it with the JOIN
```
SELECT lowlevel.mbid
  FROM lowlevel
  JOIN lowlevel_json USING (id)
 WHERE lowlevel_json.data #> '{metadata,audio_properties,md5_encoded}' ? :md5
```

hellska:
Code review done.

db/data.py
```diff
@@ -460,3 +462,41 @@ def get_summary_data(mbid, offset=0):
...
+    
+    Returns: 
+        itemuuid: The generated random UUIDv4 assigned to a specific dataset item
+    """
+    itemuuid = uuid4()
+    
+    if itemuuid:
+        return str(itemuuid)
+    else:
+        return None
```
gentlecat: This function doesn't make much sense to me. Can just use `uuid4()`. It shouldn't fail. And even if it does we've got bigger problems and would like it to crash, not fail silently.

webserver/views/api/v1/core.py
```diff
@@ -10,6 +10,19 @@
 bp_core = Blueprint('api_v1_core', __name__)
 
 
+@bp_core.route("/datasets", methods=["POST"])
```
gentlecat: I actually already implemented this submission part in my PR with dataset API. https://github.com/metabrainz/acousticbrainz-server/pull/182/files#diff-f8f1fe605fbc564266894db22f5b32ffR23 Perhaps we can merge some of the parts that were implemented and you can just reuse them. @alastair, what do you think? I can modify the PR to include only endpoints that are done.

gentlecat:
I thought you were also going to implement endpoint for accepting submissions without MBID. Is that still on your roadmap? It was one of the things that we planned to finish before mid-term evaluation.

webserver/views/api/v1/core.py
gentlecat: OK, I updated PR #182.

hellska:
The submission without MBID is still in the roadmap, I implemented the client side submission process to insert dataset items without MBID and discussed with Alastair the idea of implementing a new endpoint to submit without MBID and merge the endpoint for standard submission with the new one later. Implementation without mbid in the URL.

hellska:
Start to implement the low-level data submission without MBIDs. I created the endpoint in the core.py because of the idea to merge the endpoint with the standard submission.

db/test/test_data.py
```diff
@@ -260,9 +263,20 @@ def test_add_get_model(self):
...
+    def test_find_md5_duplicates(self):
+        
+        self.assertEqual(db.data.find_md5_duplicates(self.test_md5), None)
+        self.assertIsNone(db.data.find_md5_duplicates(self.test_md5))
+        db.data.submit_low_level_data(self.test_mbid, self.test_lowlevel_data)
+        self.assertEqual(str(db.data.find_md5_duplicates(self.test_md5)), self.test_mbid)
+        
+    def test_generate_datasetitem_uuidv4(self):
+        
+        self.assertTrue(db.data.generate_datasetitem_uuidv4())
```
alastair: I think @gentlecat pointed this out previously, about not needing to have a method to generate the uuid. We definitely don't need a test for it either. This is part of the standard library, so we can expect it to work. If you need a uuid you should just use `str(uuid.uuid4())` directly.

db/test/test_data.py
```diff
@@ -260,9 +263,20 @@ def test_add_get_model(self):
         get_id = db.data._get_model_id("modelname", "v1")
         self.assertEqual(modelid, get_id)
 
-
     def test_get_summary_data(self):
         pass
+    
+    def test_find_md5_duplicates(self):
+
```
alastair: don't need this newline

db/test/test_data.py
```diff
@@ -260,9 +263,20 @@ def test_add_get_model(self):
...
         self.assertEqual(modelid, get_id)
 
-
     def test_get_summary_data(self):
         pass
+    
+    def test_find_md5_duplicates(self):
+        
+        self.assertEqual(db.data.find_md5_duplicates(self.test_md5), None)
+        self.assertIsNone(db.data.find_md5_duplicates(self.test_md5))
```
alastair: You're checking the same thing twice here. I'd just keep `assertIsNone`

db/data.py
```diff
@@ -460,3 +462,37 @@ def get_summary_data(mbid, offset=0):
...
+    Args: 
+        md5: The md5_encoded from the low-level data to check
+    
+    Returns:
+        The retrieved mbid if the md5 already exists
+        None if the md5 is not found
+    """
+
+    query = text(
+            """SELECT lowlevel.mbid
```
alastair: you can select `lowlevel.mbid::text` here to return a string instead of a UUID object. We use this in other places in `data.py`

db/data.py
```diff
@@ -460,3 +462,37 @@ def get_summary_data(mbid, offset=0):
...
+                 FROM lowlevel
+                 JOIN lowlevel_json USING (id)
+                WHERE lowlevel_json.data#> '{metadata,audio_properties,md5_encoded}' ? :md5""")
+    with db.engine.begin() as connection:
+        result = connection.execute(query, {"md5": md5})
+        row = result.fetchone()
+        if row:
+            # returns the retrieved mbid
```
alastair: This comment isn't needed. In general you should use comments to explain the reasoning behind a decision, rather than just saying in text the same thing that the statement does. We don't have any comments like this in other methods in this file.

db/test/test_data.py
```diff
@@ -326,3 +341,6 @@ def test_clean_metadata(self):
         db.data.clean_metadata(d)
         self.assertFalse('unknown_tag' in d['metadata']['tags'])
         self.assertTrue('file_name' in d['metadata']['tags'])
+
```
alastair: Did you add these newlines to the end of the file? They're not needed

db/test/test_data.py
hellska: Oh, I thought it was related to check I was doing int the function not to the function itself.

hellska:
Code reviewed!

alastair:
Please make sure you give informative commit messages. Here's a good introduction: http://chris.beams.io/posts/git-commit/
Your messages should always describe _what_ (and if possible, _why_) you made the change. "Code Review" doesn't tell us anything about the change, meaning we need to look at the diff to see what has changed.
If the changes that you make modify distinct parts of the file, you can always make many commits

hellska:
Oh sorry, I thought it was a good message to avoid redundancy, just because I changed only those parts related with the comments, but I will be more explicit and detailed in the future!!! Thanx for the post that is really good!

hellska:
Merged metabrainz/master!

alastair:
I suspect you didn't run the tests. Please run them and make sure there are no failures.

hellska:
Now it seems ok! Run the tests and fixed some wrong imports.

webserver/views/api/v1/core.py
```diff
@@ -109,3 +111,48 @@ def _validate_offset(offset):
     else:
         offset = 0
     return offset
+
+@bp_core.route("/low-level", methods=["POST"])
+def submit_low_level_nombid():
```
alastair: This looks pretty good. a few comments:
 * Perhaps we can name the url endpoint to say that it's specifically one to generate an id for data with no mbid
 * I think that a submission to this endpoint with a musicbrainz tag should be an error. It can be up to the client to see if a data file has an mbid and use the correct endpoint
 * If the data already exists (the md5 is already in the database), I wonder if we can report this somehow:
  * Use HTTP 201 CREATED if we make it, and 200 otherwise
  * Redirect to the existing item if it's already there (will require changing the client to know how to understand the redirect)

hellska: In the last commit I implemented all your suggestion, the only missing thing is the redirect. I don't understand what you mean, the client in the current implementation expects to receive only the gid(mbid,msid) to complete the dataset dictionary, so why do we have to redirect?

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+                        msid = responsedata['payload'][0]['ids']['recording_msid']
+                    else:
+                        msid = None
+                return msid
+            else:
+                raise TypeError('Request to messybrainz failed with HTTP status code:'+str(r.status_code))
+        else:
+            raise TypeError('Bad Request Data')
+    else:
+        raise TypeError('Input data is not a dictionary')
```
alastair: You check on line 16 if this is a list, not a dictionary

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+    Args:
+        querydata: A dictionary containing at least artis and song title
+
+    Returns:
+        msid: string messybrainz id -> if it is found in the messybrainz database
+        None: if no reference is fount in messybrainz database
+
+    """
+    msurl = 'https://messybrainz.org/submit'
+    if isinstance(querydata, list):
```
alastair: Why do you have a list here? The only place that you use this code, you only submit data for a single item. Perhaps you could just pass a single dictionary here, which means you will need less if statements

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+                if 'recording_mbid' in responsedata['payload'][0]['ids'].keys() and 'recording_msid' in responsedata['payload'][0]['ids'].keys():
+                    if responsedata['payload'][0]['ids']['recording_mbid'] != '':
+                        msid = responsedata['payload'][0]['ids']['recording_mbid']
+                    elif responsedata['payload'][0]['ids']['recording_msid'] != '':
+                        msid = responsedata['payload'][0]['ids']['recording_msid']
+                    else:
+                        msid = None
+                return msid
+            else:
+                raise TypeError('Request to messybrainz failed with HTTP status code:'+str(r.status_code))
```
alastair: in python, a `TypeError` means that the type of the variables used as the input to a function was incorrect: https://docs.python.org/2/library/exceptions.html#exceptions.TypeError
If you fail to make a query to messybrainz, it doesn't mean that the data was wrong, it means that there was a problem accessing MessyBrainz

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+        msid: string messybrainz id -> if it is found in the messybrainz database
+        None: if no reference is fount in messybrainz database
+
+    """
+    msurl = 'https://messybrainz.org/submit'
+    if isinstance(querydata, list):
+        if 'artist' in querydata[0].keys() and 'title' in querydata[0].keys():
+            jsondata = ujson.dumps(querydata)
+            r = requests.post(msurl, data=jsondata)
+            if r.status_code == 200:
```
alastair: requests has a neat method, `r.raise_for_status()` (http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status) which will do nothing if the status code is 200, otherwise raise an exception.

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+                #get the correct id
+                responsedata = r.json()
+                if 'recording_mbid' in responsedata['payload'][0]['ids'].keys() and 'recording_msid' in responsedata['payload'][0]['ids'].keys():
```
alastair: in a case like this, where you are always accessing the same data in an object, it's a good idea to create a temporary variable to store it:
```python
data = responsedata['payload'][0]['ids']
if 'recording_mbid' in data and 'recording_msid' in data:  # note that if you use 'in' with a dictionary you don't need 'keys'

File: webserver/external/messybrainz.py [hunk 27]
```diff
@@ -0,0 +1,36 @@
...
+                    if responsedata['payload'][0]['ids']['recording_mbid'] != '':
+                        msid = responsedata['payload'][0]['ids']['recording_mbid']
```
alastair: If you get an `mbid` back from messybrainz, you should return this fact back to the insert code, because you should set the lowlevel type to mbid, not msid.

webserver/views/api/v1/core.py
```diff
@@ -126,33 +127,41 @@ def submit_low_level_nombid():
...
+                    try:
+                        data['metadata']['tags']['musicbrainz_trackid'] = [mbid]
+                        submit_low_level_data(mbid, data)
+                    except BadDataException as e:
+                        logging.warn(str(e))
+                        raise webserver.views.api.exceptions.APIBadRequest("%s" % e)
+                    status = "OK"
+                    outputmsg = {"status": status, "itemuuid": mbid, "action": action}
+                    return jsonify(outputmsg), 201
+                else:
```
alastair: What is the error case that you're trying to catch here?  In HTTP, error codes are grouped based on their type. Errors that start with 2 mean "success", 3 mean "redirect", 4 mean "error caused by something the client did", and 5 "error caused by something the server did". 

I can think of these causes for an error on this endpoint, related to messybrainz data:
 * The submission doesn't have an artist or title. I would return HTTP 400 here, because the data that the user sent is bad, and it will never be accepted unless the user changes the data that they submit
 * We failed to generate a messybrainz id. This is our problem, and therefore should be a 5xx error. HTTP 503 sounds like a good option here. The idea with a 503 is that the user can probably retry with the same data later, and we'll accept it.

If we raise `NoDataFoundException` in the API then we return a status code of 404, which doesn't make sense given the two possible failure cases.

webserver/views/api/v1/core.py
```diff
@@ -126,33 +127,41 @@ def submit_low_level_nombid():
...
+                    artist = artist[0]
+                title = data['metadata']['tags']['title']
+                if isinstance(title, list):
+                    title = title[0]
+                artist_data = [{'artist': artist, 'title': title}]
+                mbid = messybrainz.get_messybrainz_id(artist_data)
+                if mbid is not None:
+                    action = "messybrainz_id"
+                    try:
+                        data['metadata']['tags']['musicbrainz_trackid'] = [mbid]
```
alastair: If we get an msid, then this is not an mbid, and definitely shouldn't be added to the `musicbrainz_trackid` tag. Perhaps we could make a new tag, `messybrainz_recordingid` in this case

webserver/views/api/v1/core.py
```diff
@@ -126,33 +127,41 @@ def submit_low_level_nombid():
...
-            data['metadata']['tags']['musicbrainz_trackid'] = [mbid]
-            action = "uuidV4_generate"
+            if 'artist' in data['metadata']['tags'].keys() and 'title' in data['metadata']['tags'].keys():
+                artist = data['metadata']['tags']['artist']
+                if isinstance(artist, list):
+                    artist = artist[0]
+                title = data['metadata']['tags']['title']
+                if isinstance(title, list):
+                    title = title[0]
+                artist_data = [{'artist': artist, 'title': title}]
```
alastair: I'm not sure if we explained to you the full behaviour of messybrainz here - ideally we want to submit as much data as possible. If we have extra metadata (album name, track number, year, artist mbid, etc) then we should submit as much to messybrainz as possible. This gives us the best possible chance of reconciling the data at a later stage.

webserver/views/api/v1/core.py
```diff
@@ -126,33 +127,41 @@ def submit_low_level_nombid():
...
     
-    if 'musicbrainz_recordingid' in data['metadata']['tags'].keys():
-        mbid = data['metadata']['tags']['musicbrainz_recordingid']
-        action = "mbid_submitted"
-    elif 'musicbrainz_trackid' in data['metadata']['tags'].keys():
-        mbid = data['metadata']['tags']['musicbrainz_trackid']
-        action = "mbid_submitted"
+    if 'musicbrainz_recordingid' in data['metadata']['tags'].keys() or 'musicbrainz_trackid' in data['metadata']['tags'].keys():
+        raise webserver.views.api.exceptions.APIBadRequest('### The data files contains an ID this endopoint only accept nonmbid submissions!')
     elif 'md5_encoded' in data['metadata']['audio_properties'].keys():
```
alastair: This is quite a complex method now. I would like to see a number of tests that confirm that various metadata values cause the correct lookups to be made. Every time you have a different control flow (for example, an exception or an if statement) you should consider having another test to verify the behaviour. Looking at your exceptions, here are some suggestions for tests:

* Bad json submitted to the endpoint
* A submission with a recording id tag
* A submission with no md5 value
* A submission with an md5 twice in a row
* A submission with no artist or title tag
* An error calling messybrainz
* A submission with additional metadata (album, track number, etc)

for the methods which call listenbrainz, you can create a `mock` of the messybrainz lookup method: http://www.voidspace.org.uk/python/mock/
so that you don't actually call this method during testing

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+    This function queries messybrainz to get a msid according to the 'artist/title' present in the item metadata
+    Args:
+        querydata: A dictionary containing at least artis and song title
+
+    Returns:
+        msid: string messybrainz id -> if it is found in the messybrainz database
+        None: if no reference is fount in messybrainz database
+
+    """
+    msurl = 'https://messybrainz.org/submit'
```
alastair: This should be a config parameter in `config.py`, so that people can run their own messybrainz server during development. We don't want to submit bad data to messybrainz when we are working on new features.

alastair: An additional feature that we need to add here is user authentication for messybrainz. Our preference is that only specific services can access it, so we should have some sort of authentication system. 
If you want, you could do this yourself. It's a related feature, and would be a good introduction into setting up and working on another codebase. I can write a ticket describing what needs to be done.
If you would prefer to continue working on this part of acousticbrainz, let me know and I'll make the necessary changes in messybrainz for you.

gentlecat: I'd say we should finish what we planned for GSoC project first.

alastair:
despite the huge number of comments I made, this is a good patch! It works in the general case, which is a good first step. almost everything I suggested are improvements to the general base of the code, to make it more robust and easier to follow. One more revision should be enough to clean up these issues.

Feel free to address the comments in a number of different commits and push them one change at a time.

webserver/external/messybrainz.py
hellska: The list is needed because messybrainz only accept lists in the [API submit endpoint](https://github.com/metabrainz/messybrainz-server/blob/master/webserver/views/api.py#L30)

alastair: Yep, that's true.
However, it's important to consider inputs and outputs of functions and how you are going to be using them. If you think about the flow of how data comes into AcousticBrainz via this method, it can help to make a decision on the best parameters to have.
We have an endpoint which takes a single json file from the extractor. From this, we extract the title and artist, and make a request to messybrainz.
At no point in this workflow do we consider more than one recording at a time (the usecase for having a list of items). Therefore, it would be reasonable to make the `get_messybrainz_id` take as parameters only data for one song.
If the messybrainz API requires input to be a list, that's fine. You can just wrap the data in a list before you make the call:
```python
jsondata = ujson.dumps([querydata])
```
In fact, you already make an assumption that there is only ever one item in this list when you check `if 'artist' in querydata[0].keys()`, so by making the input singular you reduce the amount of logic and code that you need for this test

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+    if isinstance(querydata, list):
+        if 'artist' in querydata[0].keys() and 'title' in querydata[0].keys():
```
alastair: (there's another small bug here, if querydata is an empty list then this will fail)

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,36 @@
...
+            jsondata = ujson.dumps(querydata)
+            r = requests.post(msurl, data=jsondata)
```
alastair: requests has a nice feature where you can do
```python
r = requests.post(msurl, json=querydata)
```
and it will automatically encode the data for you

webserver/external/messybrainz.py
hellska: I put this as TODO once we merge the PR195 that already implement the changes to accept the type of submission in the [submit_low_level_data](https://github.com/metabrainz/acousticbrainz-server/pull/195/files#diff-973d6aba9f5f7b7acda5e7780e0c45a7R129)

webserver/views/api/v1/core.py
hellska: If we add the new tag the id_type that we pass to the function in [PR195](https://github.com/metabrainz/acousticbrainz-server/pull/195/files#diff-973d6aba9f5f7b7acda5e7780e0c45a7R86) does not make much sense and we can define the id_type from the tag in which the value is stored!

hellska:
[WIP] some advices are still not implemented, working on that now, but I'm dividing things in different commits as suggested!

webserver/external/messybrainz.py
hellska: I simplified the code as much as possible, the list is created in the request itself:
`r = requests.post(msurl, json=[querydata])`

webserver/external/messybrainz.py
hellska: This check does not exists any more, but what error would you use in this case? Would you create a specific exception for this? I do not see any related in the builtin-exceptions.

webserver/views/api/v1/core.py
hellska: I understand the comment, but I need some further explanation of messybrainz :)
I figured out that we cannot send artist as list to messybrainz, while all the [tags in our json are lists](https://github.com/metabrainz/acousticbrainz-server/blob/master/db/test_data/0dad432b-16cc-4bf0-8961-fd31d124b01b.json#L2430). One solution could be to explode the artist list and create N recording one per each artist in the list.
The song title can be sent as a list!

webserver/external/messybrainz.py
```diff
@@ -1,35 +1,33 @@
 import requests
-import ujson
+import config
```
gentlecat: In webserver you shouldn't reference config file directly. Use `current_app.config`.

webserver/external/messybrainz.py
```diff
@@ -1,35 +1,33 @@
...
-                    if responsedata['payload'][0]['ids']['recording_mbid'] != '':
-                        msid = responsedata['payload'][0]['ids']['recording_mbid']
-                    elif responsedata['payload'][0]['ids']['recording_msid'] != '':
-                        msid = responsedata['payload'][0]['ids']['recording_msid']
-                    else:
-                        msid = None
-                return msid
-            else:
-                raise TypeError('Request to messybrainz failed with HTTP status code:'+str(r.status_code))
+    msurl = config.MESSYBRAINZ_URL+'submit'
```
gentlecat: You can just move this into a `post` call and get rid of a variable.

webserver/external/messybrainz.py
```diff
@@ -1,35 +1,33 @@
...
+                    msid = responsedata['payload'][0]['ids']['recording_mbid']
+                    id_type = 'mbid'
+                elif responsedata['payload'][0]['ids']['recording_msid'] != '':
+                    msid = responsedata['payload'][0]['ids']['recording_msid']
+                    id_type = 'msid'
+                else:
+                    msid = None
+            return msid, id_type
         else:
             raise TypeError('Bad Request Data')
```
gentlecat: Be a bit more descriptive here. What's bad about it?

gentlecat: Also, it's not `TypeError`. Probably `ValueError`.

webserver/external/messybrainz.py
```diff
@@ -1,35 +1,33 @@
...
-                return msid
-            else:
-                raise TypeError('Request to messybrainz failed with HTTP status code:'+str(r.status_code))
+    msurl = config.MESSYBRAINZ_URL+'submit'
+    if isinstance(querydata, dict):
+        if 'artist' in querydata and 'title' in querydata:
+            r = requests.post(msurl, json=[querydata])
+            r.raise_for_status()
+            responsedata = r.json()
+            if 'recording_mbid' in responsedata['payload'][0]['ids'] and 'recording_msid' in responsedata['payload'][0]['ids']:
```
gentlecat: You reference `responsedata['payload'][0]['ids']` 6 times here. Can move it out into `ids` variable to make it easier to read.

webserver/external/messybrainz.py
```diff
@@ -1,35 +1,33 @@
...
     """
     This function queries messybrainz to get a msid according to the 'artist/title' present in the item metadata
     Args:
-        querydata: A dictionary containing at least artis and song title
+        querydata: A dictionary containing basic info for one track
+                   minimum fields required: [artis, song title]
 
     Returns:
         msid: string messybrainz id -> if it is found in the messybrainz database
         None: if no reference is fount in messybrainz database
```
gentlecat: Looks like this is outdated.

webserver/views/api/v1/core.py
hellska: [dataset_creation_toolkit 849ed99] The 'return ... 503' is still not optimal, but in this way we return the correct HTTP status code to the submitter.

webserver/views/api/v1/core.py
hellska: Check the commit [f7b4884](https://github.com/hellska/acousticbrainz-server/commit/f7b48846b25843c5830fef4c87841e0d8da2808f) that implement almost all those cases. There is a TODO note for the error calling Messybrainz.

hellska:
The py.test runs successfully but I still have those pylint errors, what's going on with this?

y-van-z:
Most reported errors are about code lines over 79 characters.

gentlecat:
Yeah, don't worry about those.

db/data.py
```diff
@@ -1,4 +1,5 @@
 from hashlib import sha256
+from uuid import uuid3, uuid4, NAMESPACE_URL
```
gentlecat: Doesn't look like these are used currently.

db/test/test_data.py
```diff
@@ -5,15 +5,18 @@
...
 
 class DataDBTestCase(DatabaseTestCase):
 
     def setUp(self):
         super(DataDBTestCase, self).setUp()
         self.test_mbid = "0dad432b-16cc-4bf0-8961-fd31d124b01b"
+        self.test_md5 = "19967dc1f0699ba7c63660c0f1c4125f"
         self.test_lowlevel_data_json = open(os.path.join(TEST_DATA_PATH, self.test_mbid + '.json')).read()
         self.test_lowlevel_data = json.loads(self.test_lowlevel_data_json)
+        self.test_lowlevel_data_uuidv3 = '7b199968-d177-3da4-b86c-94cb439fb750'
```
gentlecat: Unused?

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,34 @@
...
+    This function queries messybrainz to get a msid according to the 'artist/title' present in the item metadata
+    Args:
+        querydata: A dictionary containing basic info for one track
+                   minimum fields required: [artis, song title]
+
+    Returns:
+        msid: string messybrainz id -> if it is found in the messybrainz database
+        None: if no reference is fount in messybrainz database
+    """
+    if isinstance(querydata, dict):
```
gentlecat: A better way to write this would be to do the opposite check: if it's *not* a dictionary. That way you can more easily follow implementation of that function without having to go deeper and deeper inside `if`s to understand  how it's supposed to work in normal case.

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,34 @@
...
+        if 'artist' in querydata and 'title' in querydata:
```
gentlecat: Here too.

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,34 @@
...
+            r = requests.post(current_app.config['MESSYBRAINZ_URL']+'submit', json=[querydata])
+            r.raise_for_status()
+            responsedata = r.json()
```
gentlecat: `response_data`

gentlecat:
I changed the title to be more descriptive of what's there. See if it's correct.

gentlecat:
*This branch is going to be rebased after #195 is merged.*

hellska:
Rebased on metabrainz/master.
Now I have some errors when I run the tests, this is an excerpt of the errors:

`
>       self.assertEqual(resp.status_code, 401)
E       AssertionError: 400 != 401

webserver/views/api/v1/test/test_datasets.py:22: AssertionError
-------------- 
>       self.assertRedirects(resp, url_for("user.profile", musicbrainz_id=self.test_user_mb_name))

webserver/views/test/test_datasets.py:188: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../venv2-acousticbrainz/local/lib/python2.7/site-packages/flask_testing/utils.py:213: in assertRedirects
    self.assertTrue(response.status_code in (301, 302))
E   AssertionError: False is not true
`

Maybe it's useless to solve those issues because someone else is already working on that!

alastair:
We won't be completing this feature, so closing

PR #191, opened 2016-06-09
Gaia Config and essentia extractor

hellska:
Removed ‘—download’ option from waf configure
Changed the name of the essentia extrator from
streaming_extractor_music_svm to essentia_streaming_extractor_music_svm
and changed the destination dir to hl_etractor


alastair:
Thanks. With the streaming -> essentia_streaming change do we also need to edit the highlevel extractor code to call the correct extractor?

hellska:
No, it's renamed during the copy.
And the destination dir is 'hl_extractor', there is a typo in the commit message.

alastair:
Thanks!

PR #193, opened 2016-06-13
Added an option for remote evaluation of dataset

kartikgupta0909:
This is about ticket number AB-132.


db/dataset_eval.py
```diff
@@ -11,12 +11,14 @@
...
-
+# Eval locations are defined in `eval_location_type` type. See schema definition.
+EVAL_LOCAL = "local"
+EVAL_REMOTE = "remote"
 # Filter types are defined in `eval_filter_type` type. See schema definition.
 FILTER_ARTIST = "artist"
 
 
-def evaluate_dataset(dataset_id, normalize, filter_type=None):
+def evaluate_dataset(dataset_id, normalize, filter_type=None, eval_location=EVAL_LOCAL):
```
alastair: I would make `eval_location` a required variable, and put it after `normalize`

db/test/test_dataset_eval.py
```diff
@@ -133,3 +133,12 @@ def test_delete_job(self):
...
         dataset_eval.delete_job(job_id)
         self.assertIsNone(dataset_eval.get_job(job_id))
+
+    def test_eval_job_location(self):
+        job1_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, None, dataset_eval.EVAL_REMOTE)
+        job1 = dataset_eval.get_job(job1_id)
+
+        job2_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, None, dataset_eval.EVAL_LOCAL)
+        job2 = dataset_eval.get_job(job2_id)
+        self.assertEqual(job1["eval_location"], dataset_eval.EVAL_REMOTE)
```
alastair: you should move this `assert` to line 140, directly after you get job1

webserver/views/test/test_datasets.py
```diff
@@ -201,3 +202,15 @@ def test_recording_info(self):
         self.assert200(resp)
         resp = self.client.get(url_for("datasets.recording_info", mbid=self.test_uuid))
         self.assert404(resp)
+
+    def test_evaluate_location_options(self):
+        self.temporary_login(self.test_user_id)
+        dataset_id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        resp = self.client.get(url_for("datasets.evaluate", dataset_id=dataset_id))
+        self.assertStatus(resp, 200)
```
alastair: You should put a newline here to separate the fact that these are two separate tests

webserver/views/test/test_datasets.py
```diff
@@ -201,3 +202,15 @@ def test_recording_info(self):
...
+        evaluate_form = forms.DatasetEvaluationForm()
+        evaluate_form.filter_type.data = forms.DATASET_EVAL_NO_FILTER
+        evaluate_form.evaluation_location.data = forms.DATASET_EVAL_REMOTE
+        evaluate_form.normalize.data = True        
+        resp = self.client.get(url_for("datasets.evaluate", dataset_id=dataset_id, form=evaluate_form))
```
alastair: You should be clearer about what this test is doing. Are you checking that the form shows when you load the page, or that the form values are set when you submit the page?
If it's the second, this should be a `post` request, not `get`. You should also check that the methods in the views are actually called (https://github.com/metabrainz/acousticbrainz-server/blob/33eb2403678e6126f36c09eb609922b390a429ac/webserver/views/datasets.py#L112)
One way of doing this is by using the mocks module: https://github.com/metabrainz/acousticbrainz-server/blob/e2da6e7c664d061cb6ba4beb504dd60b5eb67db2/db/test/test_data.py
This means that we don't actually have to run the methods

admin/updates/update-2-add_dataset_eval_location.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
```
alastair: @gentlecat suggested here that we change the filenames of our update files to have the current date `yyyy-mm-dd` instead of a number, this way if two people make a change to alter the database at the same time we won't have conflicting numbers. Can you change this?

Make sure you don't have 3 newlines at the end of the file

db/dataset_eval.py
kartikgupta0909: I did that because if this functions is might be called somewhere else which I might not have noticed we wont have to change that. It can take a default value on its on, and wherever we need to give it a value of our choice we are doing that. However this might be against good software practice, so if you think we need to change that I can do that.

db/test/test_dataset_eval.py
kartikgupta0909: Will correct this.

webserver/views/test/test_datasets.py
kartikgupta0909: Agreed that they should be post rather than get, and will change that.

admin/updates/update-2-add_dataset_eval_location.sql
kartikgupta0909: Sure, I ll do that.

db/dataset_eval.py
```diff
@@ -11,12 +11,14 @@
 STATUS_RUNNING = "running"
 STATUS_DONE = "done"
 STATUS_FAILED = "failed"
-
+# Eval locations are defined in `eval_location_type` type. See schema definition.
```
alastair: Think carefully about what you want to say in comments. They normally shouldn't just duplicate what the code says. One suggestion people make about comments is "say why, not what". You can read the code to see what it does.
Don't think that you have to copy a comment from similar code if you don't think that it gives any value.
I would consider writing here something like

`# Location to run an evaluation (on the AB server of the user's local server), defined in postgres type 'eval_location_type'`

kartikgupta0909: I did that because there were similar comments for the other constants. However the one you suggested looks better.

webserver/templates/datasets/evaluate.html
```diff
@@ -21,6 +21,10 @@ <h2 class="page-title">Evaluate dataset "{{ dataset['name'] }}"</h2>
         <div class="col-sm-4">{{ form.filter_type(class="form-control", required="required") }}</div>
       </div>
       <div class="form-group">
+        <label class="col-sm-2 control-label">{{ form.evaluation_location.label.text }}</label>
```
alastair: we talked briefly in IRC about having a "feature flag" so that we can merge this patch and release it but disable this form element.

You can have a section in `config.py` and have a variable called `FEATURE_something`. If it is set to `False`, don't show this option here, and set `eval_location` to `local` always. Make sure that it is easy to remove this setting when we make the feature permanent (hopefully we only want 2 or 3 if statements if possible)

webserver/templates/datasets/evaluate.html
```diff
@@ -21,6 +21,10 @@ <h2 class="page-title">Evaluate dataset "{{ dataset['name'] }}"</h2>
...
+        <div class="col-sm-4">{{ form.evaluation_location(class="form-control", required="required") }}</div>
+      </div>
+      <div class="form-group">
```
alastair: It would be nice to show some documentation here for users who don't know what this is. This should be another component of the project that we add. I'm not sure what the best way to do this is. One option is to make our own help pages and link to them from here, or to add our documentation to the github wiki for acousticbrainz-server. What do you think?

kartikgupta0909: I think a help page describing in detail how to use this tool would be good. And giving links for the page on the evaluate page would be a good idea. Maybe we could give it a tag or something which says "Make this evaluation your contribution to AB by evaluating on your own machine", and then the user could click on that and get to the page where he/she can read the instructions.

webserver/templates/datasets/evaluate.html
kartikgupta0909: I think this can be done. I ll look into it.

db/dataset_eval.py
alastair: Right. Always think twice when you copy things (comments or code). Don't duplicate content just because it's already there. Also consider if it can be implemented better (integrating both parts into 1 piece of code), or if you should change both the old and the new code to be tidier, or easier to understand

db/dataset_eval.py
alastair: There are two ways of making sure that we don't miss code from other places:
 * use grep to check if the method is called anywhere else
 * use our tests

Ideally we should be using tests, but I know that we don't always have tests covering all of the website. You should always be aiming to make the "correct" change, instead of the "easiest", even if it means that you need to go to some more effort to change other parts of the code.

webserver/templates/datasets/evaluate.html
```diff
@@ -20,6 +20,12 @@ <h2 class="page-title">Evaluate dataset "{{ dataset['name'] }}"</h2>
         <label class="col-sm-2 control-label">{{ form.filter_type.label.text }}</label>
         <div class="col-sm-4">{{ form.filter_type(class="form-control", required="required") }}</div>
       </div>
+      {% if config['EVAL_LOCATION'] %}
```
gentlecat: I don't see this value in example config file. This will fail if it's not defined. Better use `config.get('EVAL_LOCATION')`.

gentlecat: It might also be better to call this flag `FEATURE_ EVAL_LOCATION` or `FEATURE_REMOTE_EVAL`.

webserver/views/datasets.py
```diff
@@ -112,7 +112,8 @@ def evaluate(dataset_id):
             db.dataset_eval.evaluate_dataset(
                 dataset_id=ds["id"],
                 normalize=form.normalize.data,
-                filter_type=form.filter_type.data,
+                eval_location=form.evaluation_location.data,
+                filter_type=form.filter_type.data
```
gentlecat: Please keep the tailing comma so that changes like these don't show up. Only one line should change here.

db/dataset_eval.py
gentlecat: As a side note, in Python 3 there is a feature that lets you force to explicitly specify keyword arguments. https://www.python.org/dev/peps/pep-3102/

gentlecat:
@brainzbot, add to whitelist.

db/dataset_eval.py
```diff
@@ -11,12 +11,15 @@
 STATUS_RUNNING = "running"
 STATUS_DONE = "done"
 STATUS_FAILED = "failed"
-
```
gentlecat: Keep sets of "constants" separate. That way it would be easier to read them.

db/dataset_eval.py
```diff
@@ -11,12 +11,15 @@
...
+# Location to run an evaluation (on the AB server of the user's local server), 
+# defined in postgres type 'eval_location_type'
+EVAL_LOCAL = "local"
+EVAL_REMOTE = "remote"
 # Filter types are defined in `eval_filter_type` type. See schema definition.
 FILTER_ARTIST = "artist"
 
 
-def evaluate_dataset(dataset_id, normalize, filter_type=None):
+def evaluate_dataset(dataset_id, normalize, eval_location, filter_type=None):
```
gentlecat: New argument is undocumented.

db/dataset_eval.py
```diff
@@ -248,15 +252,16 @@ def add_dataset_snapshot(connection, data):
     return snapshot_id
 
 
-def _create_job(connection, dataset_id, normalize, filter_type=None):
+def _create_job(connection, dataset_id, normalize, eval_location, filter_type=None):
+
```
gentlecat: ✂️

webserver/forms.py
```diff
@@ -6,7 +6,8 @@
 
 
 DATASET_EVAL_NO_FILTER = "no_filtering"
-
```
gentlecat: Same here.

kartikgupta0909:
Updated the PR with the changes suggested by Gentlecat.

alastair:
I've merged this into master, and made a few small changes
@gentlecat can you give a quick 👀 ?

config.py.sample
```diff
@@ -57,3 +57,6 @@ MEMCACHED_NAMESPACE = "AB"
 #MAIL_FROM_DOMAIN = "acousticbrainz.org"
 
 FILE_STORAGE_DIR = "./files"
+
+# Feature Flags
+FEATURE_EVAL_LOCATION = False
```
gentlecat: I'm wondering if we should add a file with default configuration values so that people who use the server don't have to copy stuff from example one all the time unless it's something critical. See https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/default_config.py.

db/dataset_eval.py
gentlecat: Looks like that's fixed now.

db/dataset_eval.py
```diff
@@ -261,16 +271,16 @@ def add_dataset_eval_set(connection, data):
     return snapshot_id
 
 
-def _create_job(connection, dataset_id, normalize, filter_type=None):
+def _create_job(connection, dataset_id, normalize, eval_location, filter_type=None):
```
gentlecat: Maybe it would be nice to check if value in `eval_location` is valid before trying to insert. Like it's done with a `filter_type`.

gentlecat:
Looks OK.

config.py.sample
alastair: ahh, good idea. My idea was to make sure we always got it from the config file with a default if it didn't exist, but this is good

db/dataset_eval.py
alastair: :+1:

config.py.sample
kartikgupta0909: Added the default config file.

PR #195, opened 2016-06-19
Schema change to support non-MBID submissions

hellska:
Add the submission_type column to identify if a record is a recording with an MBID
or a recording with a UUIDv4 generated by the AB server

Add an update sql script to modify existing table
Modify default table creation with the new column

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -2,11 +2,11 @@ BEGIN;
 
 CREATE TABLE lowlevel (
   id              SERIAL,
-  gid             UUID NOT NULL,
-  build_sha1      TEXT NOT NULL,
+  gid             UUID    NOT NULL,
+  build_sha1      TEXT    NOT NULL,
   lossless        BOOLEAN                  DEFAULT 'n',
   submitted       TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
-  is_mbid         BOOLEAN NOT NULL
+  is_mbid         id_type NOT NULL
 );
 
 CREATE TABLE lowlevel_json (
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -2,11 +2,11 @@ BEGIN;
 
 CREATE TABLE lowlevel (
   id              SERIAL,
-  mbid            UUID NOT NULL,
+  gid             UUID NOT NULL,
   build_sha1      TEXT NOT NULL,
   lossless        BOOLEAN                  DEFAULT 'n',
   submitted       TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
-  submission_type BOOLEAN NOT NULL         DEFAULT FALSE
+  is_mbid         BOOLEAN NOT NULL
 );
 
 CREATE TABLE lowlevel_json (
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -1,12 +1,12 @@
 BEGIN;
 
 CREATE TABLE lowlevel (
-  id              SERIAL,
-  gid             UUID    NOT NULL,
-  build_sha1      TEXT    NOT NULL,
-  lossless        BOOLEAN                  DEFAULT 'n',
-  submitted       TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
-  is_mbid         id_type NOT NULL
+  id         SERIAL,
+  gid        UUID      NOT NULL,
+  build_sha1 TEXT      NOT NULL,
+  lossless   BOOLEAN                  DEFAULT 'n',
+  submitted  TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  gid_type   gid_type  NOT NULL
 );
 
 CREATE TABLE lowlevel_json (
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -1,11 +1,12 @@
 BEGIN;
 
 CREATE TABLE lowlevel (
-  id          SERIAL,
-  mbid        UUID NOT NULL,
-  build_sha1  TEXT NOT NULL,
-  lossless    BOOLEAN                  DEFAULT 'n',
-  submitted   TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  id              SERIAL,
+  mbid            UUID NOT NULL,
+  build_sha1      TEXT NOT NULL,
+  lossless        BOOLEAN                  DEFAULT 'n',
+  submitted       TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  submission_type BOOLEAN NOT NULL         DEFAULT FALSE
 );
 
 CREATE TABLE lowlevel_json (
--- a/admin/sql/create_types.sql
+++ b/admin/sql/create_types.sql
@@ -1,4 +1,4 @@
 CREATE TYPE eval_job_status AS ENUM ('pending', 'running', 'done', 'failed');
 CREATE TYPE model_status AS ENUM ('hidden', 'evaluation', 'show');
 CREATE TYPE version_type AS ENUM ('lowlevel', 'highlevel');
-CREATE TYPE id_type AS ENUM ('mbid', 'mesbid');
+CREATE TYPE gid_type AS ENUM ('mbid', 'msid');
--- a/admin/sql/create_types.sql
+++ b/admin/sql/create_types.sql
@@ -1,3 +1,4 @@
 CREATE TYPE eval_job_status AS ENUM ('pending', 'running', 'done', 'failed');
 CREATE TYPE model_status AS ENUM ('hidden', 'evaluation', 'show');
 CREATE TYPE version_type AS ENUM ('lowlevel', 'highlevel');
+CREATE TYPE id_type AS ENUM ('mbid', 'mesbid');
--- a/admin/updates/20160621-non-mbid-submission.sql
+++ b/admin/updates/20160621-non-mbid-submission.sql
@@ -1,10 +1,10 @@
 BEGIN;
 
-CREATE TYPE id_type AS ENUM ('mbid', 'msid');
+CREATE TYPE gid_type AS ENUM ('mbid', 'msid');
 ALTER TABLE "lowlevel" RENAME COLUMN mbid TO gid;
-ALTER TABLE "lowlevel" ADD COLUMN is_mbid id_type;
-UPDATE "lowlevel" SET is_mbid = 'mbid';
-ALTER TABLE "lowlevel" ALTER COLUMN is_mbid SET NOT NULL;
-CREATE INDEX is_mbid_ndx_lowlevel ON lowlevel (is_mbid);
+ALTER TABLE "lowlevel" ADD COLUMN gid_type gid_type;
+UPDATE "lowlevel" SET gid_type = 'mbid';
+ALTER TABLE "lowlevel" ALTER COLUMN gid_type SET NOT NULL;
+CREATE INDEX gid_type_ndx_lowlevel ON lowlevel (gid_type);
 
 COMMIT;
--- a/admin/updates/20160621-non-mbid-submission.sql
+++ b/admin/updates/20160621-non-mbid-submission.sql
@@ -1,6 +1,6 @@
 BEGIN;
 
-CREATE TYPE id_type AS ENUM ('mbid', 'mesbid');
+CREATE TYPE id_type AS ENUM ('mbid', 'msid');
 ALTER TABLE "lowlevel" RENAME COLUMN mbid TO gid;
 ALTER TABLE "lowlevel" ADD COLUMN is_mbid id_type;
 UPDATE "lowlevel" SET is_mbid = 'mbid';
--- a/admin/updates/20160621-non-mbid-submission.sql
+++ b/admin/updates/20160621-non-mbid-submission.sql
@@ -1,0 +1,8 @@
+BEGIN;
+
+ALTER TABLE "lowlevel" RENAME COLUMN mbid TO gid;
+ALTER TABLE "lowlevel" ADD COLUMN is_mbid BOOLEAN;
+UPDATE "lowlevel" SET is_mbid = TRUE;
+ALTER TABLE "lowlevel" ALTER COLUMN is_mbid SET NOT NULL;
+
+COMMIT;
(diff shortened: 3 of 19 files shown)


admin/sql/create_tables.sql
```diff
@@ -1,11 +1,12 @@
...
-  mbid        UUID NOT NULL,
-  build_sha1  TEXT NOT NULL,
-  lossless    BOOLEAN                  DEFAULT 'n',
-  submitted   TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  id              SERIAL,
+  mbid            UUID NOT NULL,
+  build_sha1      TEXT NOT NULL,
+  lossless        BOOLEAN                  DEFAULT 'n',
+  submitted       TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  submission_type BOOLEAN NOT NULL         DEFAULT FALSE
```
gentlecat: A better name for this column would probably be something like `is_mbid`.

admin/sql/create_tables.sql
```diff
@@ -1,11 +1,12 @@
 BEGIN;
 
 CREATE TABLE lowlevel (
-  id          SERIAL,
-  mbid        UUID NOT NULL,
-  build_sha1  TEXT NOT NULL,
-  lossless    BOOLEAN                  DEFAULT 'n',
-  submitted   TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  id              SERIAL,
+  mbid            UUID NOT NULL,
```
gentlecat: I assume this is no longer going to be MBID. Should we replace `id` column with this?

gentlecat:
We now name update files like this: `<date>-<a couple of hyphen separated words summarizing the change>.sql`. In your case it would be something like `20160619-non-mbid-submissions.sql`.

admin/sql/create_tables.sql
gentlecat: Also not sure if default is needed. This should be explicitly specified when adding a column.

admin/sql/create_tables.sql
alastair: We can't replace `id`, as we have duplicate `mbid`s with multiple `id`s, and `id` is also used as an FK into the ll_json and hl tables.
MusicBrainz uses `gid` for this field (global ID?), which might make more sense, but may be confusing (it would mean one thing in AB and another thing in MB)

alastair:
With this change we now have to be explicit when we access the `lowlevel` table and say which data we are interested in. Here are some questions we need to ask:

 * Should we calculate high-level data for dataset lowlevel submissions?
 * If we visit http://acousticbrainz.org/mbid for a non-musicbrainz id, should we show data?
 * If we do show data, we need to get metadata from the submission, instead of from a MusicBrainz lookup
 * Do statistics count these submissions, or should we count them separately?
 * Do we dump these files in json for users to download?

There are 2 ways we could do this. One is to change all of the queries to explicitly add a `is_mbid` WHERE statement, but we could also create separate views for each of these values and select from the views instead. @gentlecat any preference here? I haven't used views in a real situation, but it seems useful here.

Another option which would be more work here would be to have a separate table for dataset submissions. We could continue to join it against `lowlevel_json`, but have to be careful about what we use as the shared primary key.

I'd like to see these changes made here instead of in the related submission PR, because the selection of items is tightly related to this database change.

admin/sql/create_tables.sql
gentlecat: Right, `gid` seems fine to me then.

hellska: The <mbid> column becomes <gid> and we have to change all the queries that interrogate the lowlevel table to reflect this change.

hellska:
This commit reflect the changes to the lowlevel table. It does not modify the Acousticbrainz behaviour. The is_mbid field is set by default to True in the _write_low_level function.
TODO: 
- Change the submit_low_level to receive the value of is_mbid column
- Filter the queries to the lowlevel table using the is_mbid column

alastair:
This is looking a lot better. The `mbid`->`gid` change is good. Since this is a large schema change I'm going to meet with @mayhem in person this week to give it a sanity-check to make sure that it works for our needs. Otherwise, we should be able to merge it once you've finished.

db/data.py
```diff
@@ -106,6 +106,15 @@ def submit_low_level_data(mbid, data):
         else:
             data['metadata']['audio_properties']['lossless'] = False
 
+        if 'is_mbid' in data['metadata']['tags']:
```
alastair: Where are you planning to set this? In the client?  I don't like the idea of using the `tags` dictionary as a side-channel to pass data around. Consider that it should be reproduceable to generate the json of this data using `streaming_extractor_music`. Since we already have a separate submission point for non-mbid submissions (https://github.com/metabrainz/acousticbrainz-server/pull/189/files#diff-9ac046b866a00d70d3223c15467563cfR115), can we pass `is_mbid` in as an argument to this function?

db/data_jsonb.py
```diff
@@ -21,7 +21,7 @@ def migrate_low_level(old_ll_row):
     with db.engine94.begin() as connection:
```
alastair: gah, I need to delete this file. it's not used anywhere

hellska: I can do it here, but you mean the data_jsonb.py or the reference to the the engine94?

db/data.py
hellska: Yes I think it makes much more sense. We determine the ID type in the endpoint function that can pass it as an argument!

db/data_jsonb.py
alastair: don't worry. I'll do it.

admin/sql/create_tables.sql
```diff
@@ -1,11 +1,12 @@
...
-  mbid        UUID NOT NULL,
-  build_sha1  TEXT NOT NULL,
-  lossless    BOOLEAN                  DEFAULT 'n',
-  submitted   TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  id              SERIAL,
+  gid             UUID    NOT NULL,
+  build_sha1      TEXT    NOT NULL,
+  lossless        BOOLEAN                  DEFAULT 'n',
+  submitted       TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  is_mbid         id_type NOT NULL
```
alastair: Now that this is a type which indicates many things, I don't think `is_mbid` is a good name. In the ticket I suggested that we call it `gid_type` (http://tickets.musicbrainz.org/browse/AB-135?focusedCommentId=39881&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_39881)

admin/sql/create_types.sql
```diff
@@ -1,3 +1,4 @@
 CREATE TYPE eval_job_status AS ENUM ('pending', 'running', 'done', 'failed');
 CREATE TYPE model_status AS ENUM ('hidden', 'evaluation', 'show');
 CREATE TYPE version_type AS ENUM ('lowlevel', 'highlevel');
+CREATE TYPE id_type AS ENUM ('mbid', 'mesbid');
```
alastair: We have standardised on `msid` as the abbreviation for messybrainz ids.

You can see in the github view of this commit that there is an indicator at the end of this line saying "no newline". It's a generally accepted "rule" to always have one newline at the end of a file.

admin/sql/create_tables.sql
hellska: Yes, I see 👍 
The ENUM type is called id_type, should we call it also gid_type?

admin/updates/20160621-non-mbid-submission.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+CREATE TYPE id_type AS ENUM ('mbid', 'msid');
+ALTER TABLE "lowlevel" RENAME COLUMN mbid TO gid;
```
gentlecat: Are there any constraints that need to be updated with this change?

admin/updates/20160621-non-mbid-submission.sql
```diff
@@ -0,0 +1,10 @@
...
+ALTER TABLE "lowlevel" ADD COLUMN is_mbid id_type;
```
gentlecat: I would call this column `id_type`.

admin/updates/20160621-non-mbid-submission.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+CREATE TYPE id_type AS ENUM ('mbid', 'msid');
```
gentlecat: Is it `id_type` or `gid_type`? This doesn't match *create_types.sql* file.

db/data.py
```diff
@@ -83,13 +83,14 @@ def clean_metadata(data):
     return data
 
 
-def submit_low_level_data(mbid, data):
+def submit_low_level_data(mbid, data, is_mbid):
```
gentlecat: Again, `id_type`.

db/data_jsonb.py
gentlecat: data_migrate.py too?

db/dump.py
```diff
@@ -36,7 +36,7 @@
     ),
     "lowlevel": (
         "id",
-        "mbid",
+        "gid",
```
gentlecat: Did you forget to add type here as well?

db/test/test_stats.py
```diff
@@ -42,7 +42,7 @@ def test_get_last_submitted_recordings(self, dbget):
                     "version": {"essentia_build_sha": "sha"},
                 },
             }
-            db.data.write_low_level(rand_mbid, data)
+            db.data.write_low_level(rand_mbid, data, 'mbid')
```
gentlecat: `'mbid'` value should be stored in a variable somewhere.

db/testing.py
```diff
@@ -68,5 +69,5 @@ def load_low_level_data(self, mbid):
         the database.
         """
         with open(self.data_filename(mbid)) as json_file:
-            db.data.submit_low_level_data(mbid, json.loads(json_file.read()))
+            db.data.submit_low_level_data(mbid, json.loads(json_file.read()), 'mbid')
```
gentlecat: Same issue with type here. You should reference a variable. Something like `GID_TYPE_MBID`.

admin/updates/20160621-non-mbid-submission.sql
hellska: I would call gid_tipe because if we call it id_type it can be confusing and one can think that it is referenced to the lowlevel.id column while it is related to to lowlevel.gid column. Maybe we can change the name of the enum and call it git_type_enum.

admin/updates/20160621-non-mbid-submission.sql
hellska: I checked around and don't see any constraint related to the mbid column.

hellska:
Is there any tool that you use to inspect the schema and do consistency check when make a schema update?

alastair:
Unfortunately not, the best we have at the moment is for you to open the PR and we'll check it. @gentlecat and I were speaking a bit about how we could make this better, but we don't have a plan yet

admin/updates/20160621-non-mbid-submission.sql
alastair: yes, `gid_type` is good. If it lets you, I would call the enum the same too. It's not worth calling it `gid_type_enum` because in this case we already know that it's an enum.

db/dump.py
```diff
@@ -40,6 +40,7 @@
         "build_sha1",
         "lossless",
         "submitted",
+        "gid_type"
```
gentlecat: Can add a trailing comma here too. http://dave.cheney.net/2014/10/04/that-trailing-comma

db/testing.py
```diff
@@ -11,7 +11,7 @@
 
 ADMIN_SQL_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', 'admin', 'sql')
 TEST_DATA_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'test_data')
-
+GID_TYPE_MBID = 'mbid'
```
gentlecat: All of the possible types for `gid` column should be defined in one place and imported when needed.

db/data.py
```diff
@@ -150,7 +151,7 @@ def insert_version(connection, data, version_type):
     row = result.fetchone()
     return row[0]
 
-def write_low_level(mbid, data):
+def write_low_level(mbid, data, is_mbid):
```
gentlecat: Did you forget to change this to `gid_type` or is it supposed to be `is_mbid`?

db/test/test_data.py
```diff
@@ -24,8 +24,8 @@ def test_submit_low_level_data(self, clean, write, sanity):
         clean.side_effect = lambda x: x
         sanity.return_value = None
 
-        db.data.submit_low_level_data(self.test_mbid, self.test_lowlevel_data)
-        write.assert_called_with(self.test_mbid, self.test_lowlevel_data)
+        db.data.submit_low_level_data(self.test_mbid, self.test_lowlevel_data, 'mbid')
```
gentlecat: I think these `'mbid'` arguments can be replaced with `GID_TYPE_MBID`.

db/test/test_stats.py
```diff
@@ -272,15 +274,17 @@ def add_empty_lowlevel(mbid, lossless, date):
...
-                 VALUES (:mbid, :build_sha1, :lossless, :submitted)
+            INSERT INTO lowlevel (gid, build_sha1, lossless, submitted, gid_type)
+                 VALUES (:mbid, :build_sha1, :lossless, :submitted, :gid_type)
               RETURNING id
         """)
         result = connection.execute(query,
                 {"mbid": mbid, "build_sha1": build_sha1,
-                 "lossless": lossless, "submitted": date})
+                 "lossless": lossless, "submitted": date,
+                 "gid_type": gid_type})
```
gentlecat: `gid_type` variable is used once here and can be replaced with `GID_TYPE_MSID`, similar to other places.

gentlecat:
I think it looks good apart from some minor things. We can merge this now and do more cleanup afterwards.

admin/updates/20160621-non-mbid-submission.sql
gentlecat: 👍

db/data.py
hellska: Check the code I already changed this, but I don't know why github does not updated this!

gentlecat: Oh, sorry. I was looking at the old commit.

gentlecat:
There are some conflicts, but I'm pretty sure we can merge this. Just need for @alastair to be online and confirm.

gentlecat:
OK, I rebased, squashed, and merged this in a05763593cbecb802ace895e0a26024220ef8a51. Should be working on acousticbrainz.org and beta.acousticbrainz.org now.

PR #196, opened 2016-06-20
Add information about OAuth and Admin

alastair:


README.md
```diff
@@ -67,6 +67,29 @@ commands:
 
 Full installation instructions are available in [INSTALL.md](https://github.com/metabrainz/acousticbrainz-server/blob/master/INSTALL.md) file.
 
+## Login and Datasets
```
alastair: I just realised that this is also duplicated here: 
https://github.com/metabrainz/acousticbrainz-server/blob/master/INSTALL.md#configuring-musicbrainz-oauth

We should work out where the split of these files should be. The steps that are common to both vagrant and manual install could be either in readme, or in the second half of install. Then whatever file they are not in can just say "see here for next steps"

alastair: We talked about this on IRC. Setup information should be in README. INSTALL is only for manual server install, then it should refer to readme for setup.

README.md
```diff
@@ -67,6 +67,29 @@ commands:
...
+
+To use the dataset tools you need to configure OAuth with MusicBrainz.
```
gentlecat: I would generalize this to `features that require login`.

alastair:
I moved this around a bunch. Now manual install is in `INSTALL`, configuration, running, dumps are in `README`. I don't think we need to keep tweaking little things here, if it's generally OK let's merge it and make other changes as we think of them.

gentlecat:
Sure, that sounds fine to me.

PR #197, opened 2016-06-20
Clean up html on dataset list

alastair:
Continuation of feedback on https://github.com/metabrainz/acousticbrainz-server/pull/165

### DIFF
--- a/webserver/templates/macros.html
+++ b/webserver/templates/macros.html
@@ -12,39 +12,32 @@
   <div class="alert {{ alert_class }}">{{ message }}</div>
 {% endmacro %}
 
-{% macro pager(links, prevpage, current_page, last_page, nextpage) %}
-	<nav>
-	  <ul class="pagination">
-		{% if current_page == 1 %}
-		<li class="disabled">
+{% macro pager(links, prev_page, current_page, last_page, next_page) %}
+  <nav>
+    <ul class="pagination">
+      <li class="{{'disabled' if current_page == 1}}">
+      {% if current_page == 1 %}
         <span aria-hidden="true">&laquo;</span>
-		{% else %}
-		<li>
-          <a href="{{prevpage}}" aria-label="Previous">
-			<span aria-hidden="true">&laquo;</span>
-		  </a>
-		{% endif %}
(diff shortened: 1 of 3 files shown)


gentlecat:
We use 2 spaces for indentation there.

webserver/templates/macros.html
```diff
@@ -13,38 +13,37 @@
...
+                <a href="{{prevpage}}" aria-label="Previous">
+                    <span aria-hidden="true">&laquo;</span>
+                </a>
+            {% endif %}
+            </li>
+            {% for number, link in links %}
+                {% if current_page == number %}
+                <li class="active">
+                {% else %}
+                <li>
```
gentlecat: You can reduce these to

    <li {{ class=active if current_page == number }}>

webserver/templates/macros.html
```diff
@@ -13,38 +13,37 @@
 {% endmacro %}
 
 {% macro pager(links, prevpage, current_page, last_page, nextpage) %}
```
gentlecat: `prevpage` here could be renamed to `prev_page` to be consistent with other arguments.

gentlecat: `nextpage` too.

webserver/templates/macros.html
alastair: This is a bit funny. If I use
    <li {{'class=disabled' if current_page == 1}}>
then it works, since the browser correctly parses the class even though there are no `"` quotes around the class name.
Adding quotes:
    <li {{'class="disabled"' if current_page == 1}}>
results in jinja escaping them, so that the resulting source is 
    <li class=&#34;disabled&#34;>
which isn't really right either. 

My preference is for quotes around attributes when possible, and so I think I'll leave it as is. Adding a `|safe` filter seems just as messy.

gentlecat: Can also do

    <li class="{{ 'active' if current_page == number }}">

alastair: ohhhhhhhhhhhh :+1:

gentlecat:
LGTM

PR #198, opened 2016-06-29
Some small cleanup changes

alastair:
* Make sql update filenames consistent
* Adjust requirements.txt includes to fix rtd


gentlecat:
LGTM. I'll update RTD config to point to the right requirements file.

PR #199, opened 2016-06-30
[WIP] Dataset API part 2

gentlecat:
Some functions will also be useful for extending dataset editor UI if it's going to be based on that.


alastair:
What's happening with this?

gentlecat:
That's an outline of the rest of API functions that I planned to add. Don't know if I'll work on this again. Somebody else might have to pick this up.

These are needed to make "live editing" and, perhaps, pagination available on the website.

alastair:
#219

PR #200, opened 2016-07-07
API endpoint for getting pending jobs for a user to be evaluated remo…

kartikgupta0909:
…tely added

### DIFF
--- a/db/dataset_eval.py
+++ b/db/dataset_eval.py
@@ -296,6 +296,28 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
     return job_id
 
 
+def get_user_pending_jobs(user_api_key):
+    with db.engine.connect() as connection:
+        query = sqlalchemy.text("""
+                    SELECT dataset_eval_jobs.id
+                    FROM   dataset_eval_jobs, dataset_snapshot, dataset, api_key
+                    WHERE  dataset.id = dataset_snapshot.dataset_id
+                    AND    dataset_snapshot.id = dataset_eval_jobs.snapshot_id
+                    AND    api_key.value = :author_api_key
+                    AND    status = :status
+                    AND    eval_location = :eval_location
+                    AND    author = api_key.owner
+                    ORDER BY dataset_eval_jobs.created ASC
+                """)
+        result = connection.execute(query, {
+        "status": STATUS_PENDING,
+        "eval_location": EVAL_REMOTE,
+        "author_api_key": user_api_key
+        })
+        job_ids = result.fetchall()
+        return job_ids
+
+
 class IncompleteDatasetException(db.exceptions.DatabaseException):
     pass
 
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import
 
 import json
+import sys
 
 from flask import Blueprint, request, jsonify
 
 import db.data
+import db.dataset_eval
 import webserver.views.api.exceptions
 from db.data import submit_low_level_data, count_lowlevel
 from db.exceptions import NoDataFoundException, BadDataException
@@ -109,3 +111,20 @@ def _validate_offset(offset):
     else:
         offset = 0
     return offset
+
+@bp_core.route("/<user_api_key>/get_pending_jobs", methods=["GET"])
+def get_pending_jobs(user_api_key):
+    """ Return the pending jobs for a user.
+
+    Retrieve the Job Ids for the pending jobs for the user whose
+    user id is given as an argument and return that.
+    """
+    final_result = {}
+    try:
+        pending_jobs = db.dataset_eval.get_user_pending_jobs(user_api_key)
+        for i in range(len(pending_jobs)):
+            final_result[i+1] = str(pending_jobs[i][0])
+    except:
+        e = sys.exc_info()[0]
+        raise webserver.exceptions.APIBadRequest("%s" % e)
+    return jsonify(final_result)


alastair:
Remember the previous link which I sent about wording and formatting of commit messages. I see two things here:
 * Unless you have a very good reason, it's not normal to make messages more than 70 characters. See how github has formatted yours, and added the last few letters after a ...
 * We tend to prefer using the _imperative_ when writing messages (http://stackoverflow.com/a/3580764). This means that you should use something like "Add API Endpoint for getting pending jobs for a user", instead of your "... added"

alastair:
Please link the jira ticket and pull request when you submit the PR

db/dataset_eval.py
```diff
@@ -296,6 +296,28 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
     return job_id
 
 
+def get_user_pending_jobs(user_api_key):
+    with db.engine.connect() as connection:
+        query = sqlalchemy.text("""
+                    SELECT dataset_eval_jobs.id
```
alastair: double check the way that we format SQL statements. This should be consistent throughout the file

db/dataset_eval.py
```diff
@@ -296,6 +296,28 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
...
+                    FROM   dataset_eval_jobs, dataset_snapshot, dataset, api_key
```
alastair: In postgres there is no difference between the two, but in AB we use `JOIN` instead of `FROM`/`WHERE`, so you should do
```
 SELECT ...
   FROM dataset_eval_jobs
   JOIN dataset_snapshot
     ON dataset_snapshot.id = dataset_eval_jobs.snapshot_id
   JOIN dataset
     ON dataset.id = dataset_snapshot.dataset_id
   .. etc
```

db/dataset_eval.py
```diff
@@ -296,6 +296,28 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
     return job_id
 
 
+def get_user_pending_jobs(user_api_key):
```
alastair: I think this would make more sense to be a method which gets pending jobs for a _user_, not an ID, so you could call it `get_pending_jobs_for_user`

webserver/views/api/v1/core.py
```diff
@@ -109,3 +111,20 @@ def _validate_offset(offset):
     else:
         offset = 0
     return offset
+
+@bp_core.route("/<user_api_key>/get_pending_jobs", methods=["GET"])
```
alastair: We decided that we would use the `core` blueprint only for lowlevel submission and retrieval. We can make a new blueprint for dataset evaluation items, since eventually you will need a number of endpoints for this project.

kartikgupta0909:
The associated ticket : http://tickets.musicbrainz.org/browse/AB-134

webserver/views/api/v1/core.py
kartikgupta0909: Yes we can do that as we need quite a few endpoints

alastair: In the API we have a way of logging in using the api key as a token (https://github.com/metabrainz/acousticbrainz-server/blob/f62ffa81c82f81f8c91d2816dffcfeb9ca098790/webserver/login/__init__.py#L38)
Instead of adding the user key as a parameter here, you can make the API endpoint require a logged in user and then use the user id in the pending jobs query

db/dataset_eval.py
```diff
@@ -296,6 +296,28 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
...
+                    AND    author = api_key.owner
+                    ORDER BY dataset_eval_jobs.created ASC
+                """)
+        result = connection.execute(query, {
+        "status": STATUS_PENDING,
+        "eval_location": EVAL_REMOTE,
+        "author_api_key": user_api_key
+        })
+        job_ids = result.fetchall()
+        return job_ids
```
alastair: See in other methods in this file - we return dictionaries from database _get_ methods instead of directly returning the list

alastair: https://github.com/metabrainz/acousticbrainz-server/blob/78ef1a7051ed8e776b60786d69a51b29e095adad/db/dataset_eval.py#L172

webserver/views/api/v1/core.py
```diff
@@ -109,3 +111,20 @@ def _validate_offset(offset):
...
+
+    Retrieve the Job Ids for the pending jobs for the user whose
+    user id is given as an argument and return that.
+    """
+    final_result = {}
+    try:
+        pending_jobs = db.dataset_eval.get_user_pending_jobs(user_api_key)
+        for i in range(len(pending_jobs)):
+            final_result[i+1] = str(pending_jobs[i][0])
+    except:
```
alastair: In python it's almost never a good idea to catch all exceptions like this. If you're expecting a possible error (perhaps because the user id doesn't exist?) then you should throw that specific error in the database method, and specifically catch it here

webserver/views/api/v1/core.py
```diff
@@ -109,3 +111,20 @@ def _validate_offset(offset):
...
+        e = sys.exc_info()[0]
```
alastair: This is quite an unusual way of getting information about the exception. Can you just use `except MyException as e`?

kartikgupta0909: I think catching paticular exceptions like the one you suggested above be a better idea. But will that be fine to make an sql query just to check if the user id exists or not?

webserver/views/api/v1/core.py
```diff
@@ -109,3 +111,20 @@ def _validate_offset(offset):
...
+def get_pending_jobs(user_api_key):
+    """ Return the pending jobs for a user.
+
+    Retrieve the Job Ids for the pending jobs for the user whose
+    user id is given as an argument and return that.
+    """
+    final_result = {}
+    try:
+        pending_jobs = db.dataset_eval.get_user_pending_jobs(user_api_key)
+        for i in range(len(pending_jobs)):
```
alastair: Why do you want to return this as a dictionary? Is it because `jsonify` won't return a list? In this case, it would be better to wrap the data in a holder object instead. We do this in the dataset module: https://github.com/metabrainz/acousticbrainz-server/blob/78ef1a7051ed8e776b60786d69a51b29e095adad/webserver/views/api/v1/datasets.py#L80
Something like this:
```json
{
 "success": true,
 "datasets": [the,datasets,here]
}
```

alastair: However, if you wanted to do this for some other reason in the future, a good tool here is `enumerate`:
```python
>>> for i, item in enumerate(["a", "list", "here"], 1):
...     print i, item
... 
1 a
2 list
3 here
```
the second argument is the number to start counting from. by default it starts from 0

kartikgupta0909: Yes I did this becuase we couldnt return a list or even a python dict and Jsonify wouldnt take list.

webserver/views/api/v1/core.py
```diff
@@ -109,3 +111,20 @@ def _validate_offset(offset):
...
+            final_result[i+1] = str(pending_jobs[i][0])
```
alastair: instead of using `str` on a UUID object, you can `select id::text` in the database query, which is what we do in other places

alastair:
You should make sure that you contribute tests as well, so that we can verify the behaviour of these methods, including what happens when there is no or invalid data

alastair:
it's typical to name the branch after the self-contained feature that you are proposing. You have named this branch `Offline-Dataset-Evaluation-Tool` again, but you could make it much more specific to this task - perhaps something like `user-pending-jobs`

alastair:
In the ticket that you've linked, there is an item "The client gets the list and asks for the details for each job one by one from the server and runs them". Did you want to add the API endpoint to get a single evaluation job in this ticket?

kartikgupta0909:
No, I wish to have a seperate ticket for that.

webserver/views/api/v1/core.py
alastair: You can decide how you want to check this in the database query. If you are logged in then the user will exist. If you use the auth system then you don't need to check if the api key is valid. You should decide what you want the behaviour to be if the user has no datasets to evaluate.

gentlecat:
Don't forget to delete the branch if you no longer need it.

PR #201, opened 2016-07-07
Create a highlevel row for empty data

alastair:
this prevents the highlevel calculation from continually trying to compute files which are (for whatever reason) broken.

We keep offsets consistent - that is, a request to
 `mbid/high-level?offset=3`
should return the hl for the file that you get with a request to
 `mbid/low-level?offset=3`


gentlecat:
:shipit:

PR #202, opened 2016-07-11
Add an API for getting pending jobs for a user

kartikgupta0909:
This is associated to: http://tickets.musicbrainz.org/browse/AB-134

### DIFF
--- a/db/dataset_eval.py
+++ b/db/dataset_eval.py
@@ -300,6 +300,13 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
     return job_id
 
 def get_remote_pending_jobs_for_user(user_id):
+    """ Get all jobs for a user which have been created to
+        be evaluated on a remote server.
+
+    Args:
+      user_id: the id of the user to get the jobs for
+    """
+
     with db.engine.connect() as connection:
         query = sqlalchemy.text("""
                      SELECT dataset_eval_jobs.id::text,
@@ -320,14 +327,14 @@ def get_remote_pending_jobs_for_user(user_id):
             "eval_location": EVAL_REMOTE,
             "user_id": user_id
         })
-        jobs_list = []
+        jobs = []
         for row in result.fetchall():
-            jobs_list.append({
+            jobs.append({
                 "job_id": row[0],
                 "dataset_name": row[1],
-                "job_created": row[2]
+                "created": row[2]
                 })
-        return {"username": db.user.get(user_id)["musicbrainz_id"], "jobs": jobs_list}
+        return jobs
 
 
 class IncompleteDatasetException(db.exceptions.DatabaseException):
--- a/db/dataset_eval.py
+++ b/db/dataset_eval.py
@@ -2,6 +2,7 @@ import db
 import db.exceptions
 import db.dataset
 import db.data
+import db.user
 import json
 import sqlalchemy
 from sqlalchemy import text
@@ -297,6 +298,36 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
     })
     job_id = result.fetchone()[0]
     return job_id
+
+def get_remote_pending_jobs_for_user(user_id):
+    with db.engine.connect() as connection:
+        query = sqlalchemy.text("""
+                     SELECT dataset_eval_jobs.id::text,
+                            dataset.name,
+                            dataset_eval_jobs.created
+                       FROM dataset_eval_jobs
+                       JOIN dataset_snapshot
+                         ON dataset_snapshot.id = dataset_eval_jobs.snapshot_id
+                       JOIN dataset
+                         ON dataset.id = dataset_snapshot.dataset_id
+                      WHERE status = :status
+                        AND eval_location = :eval_location
+                        AND author = :user_id
+                   ORDER BY dataset_eval_jobs.created ASC
+            """)
+        result = connection.execute(query, {
+            "status": STATUS_PENDING,
+            "eval_location": EVAL_REMOTE,
+            "user_id": user_id
+        })
+        jobs_list = []
+        for row in result.fetchall():
+            jobs_list.append({
+                "job_id": row[0],
+                "dataset_name": row[1],
+                "job_created": row[2]
+                })
+        return {"username": db.user.get(user_id)["musicbrainz_id"], "jobs": jobs_list}
 
 
 class IncompleteDatasetException(db.exceptions.DatabaseException):
--- a/db/test/test_dataset_eval.py
+++ b/db/test/test_dataset_eval.py
@@ -162,3 +162,27 @@ class DatasetEvalTestCase(DatabaseTestCase):
         job2_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, dataset_eval.EVAL_LOCAL, None)
         job2 = dataset_eval.get_job(job2_id)
         self.assertEqual(job2["eval_location"], dataset_eval.EVAL_LOCAL)
+
+    def test_get_remote_pending_jobs_for_user(self):
+        job_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, dataset_eval.EVAL_REMOTE, None)
+        job_details = db.dataset_eval.get_job(job_id)
+
+        response = dataset_eval.get_remote_pending_jobs_for_user(self.test_user_id)
+        expected_response = {
+            'username': self.test_user_mb_name,
+            'jobs': [{
+                'job_id' : str(job_id),
+                'dataset_name' : self.test_data['name'],
+                'job_created' : job_details['created']
+                }]
+            }
+        self.assertEqual(response, expected_response)
+
+        db.dataset_eval.set_job_status(job_id, db.dataset_eval.STATUS_DONE)
+        response = dataset_eval.get_remote_pending_jobs_for_user(self.test_user_id)
+
+        expected_response = {
+            'username': self.test_user_mb_name,
+            'jobs': []
+            }
+        self.assertEqual(response, expected_response)
--- a/db/test/test_dataset_eval.py
+++ b/db/test/test_dataset_eval.py
@@ -163,26 +163,44 @@ class DatasetEvalTestCase(DatabaseTestCase):
         job2 = dataset_eval.get_job(job2_id)
         self.assertEqual(job2["eval_location"], dataset_eval.EVAL_LOCAL)
 
+
     def test_get_remote_pending_jobs_for_user(self):
+        """ Check that we get remote pending jobs for a user """
+
         job_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, dataset_eval.EVAL_REMOTE, None)
         job_details = db.dataset_eval.get_job(job_id)
 
         response = dataset_eval.get_remote_pending_jobs_for_user(self.test_user_id)
-        expected_response = {
-            'username': self.test_user_mb_name,
-            'jobs': [{
+        expected_response = [{
                 'job_id' : str(job_id),
                 'dataset_name' : self.test_data['name'],
-                'job_created' : job_details['created']
+                'created' : job_details['created']
                 }]
-            }
         self.assertEqual(response, expected_response)
 
-        db.dataset_eval.set_job_status(job_id, db.dataset_eval.STATUS_DONE)
+
+    def test_get_pending_jobs_for_user_local(self):
+        """ Check that a local eval dataset for this user doesn't show """
+        job_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, dataset_eval.EVAL_LOCAL, None)
+        job_details = db.dataset_eval.get_job(job_id)
+
         response = dataset_eval.get_remote_pending_jobs_for_user(self.test_user_id)
+        self.assertEqual(response, [])
 
-        expected_response = {
-            'username': self.test_user_mb_name,
-            'jobs': []
-            }
-        self.assertEqual(response, expected_response)
+    def test_get_pending_jobs_for_user_other_user(self):
+        """ Check that a remote eval job for another user doesn't show """
+
+        another_user_id = user.create("another user")
+        another_dataset_id = dataset.create_from_dict(self.test_data, author_id=another_user_id)
+        job_id = dataset_eval._create_job(self.conn, another_dataset_id, True, dataset_eval.EVAL_REMOTE, None)
+
+        response = dataset_eval.get_remote_pending_jobs_for_user(self.test_user_id)
+        self.assertEqual(response, [])
+
+    def test_get_pending_jobs_for_user_done(self):
+        """ Check that a remote eval job with a done status doesn't show """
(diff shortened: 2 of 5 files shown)


kartikgupta0909:
PR #200 can be closed now. This is a replacement for that.

db/dataset_eval.py
```diff
@@ -298,6 +298,27 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
...
+        query = sqlalchemy.text("""
+                     SELECT dataset_eval_jobs.id::text
+                       FROM dataset_eval_jobs
+                       JOIN dataset_snapshot
+                         ON dataset_snapshot.id = dataset_eval_jobs.snapshot_id
+                       JOIN dataset
+                         ON dataset.id = dataset_snapshot.dataset_id
+                      WHERE status = :status
+                        AND eval_location = :eval_location
+                        AND author = :author_id
```
alastair: I'd prefer to name this placeholder `:user_id`, so that it matches with the parameter name

db/test/test_dataset_eval.py
```diff
@@ -162,3 +162,8 @@ def test_eval_job_location(self):
         job2_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, dataset_eval.EVAL_LOCAL, None)
         job2 = dataset_eval.get_job(job2_id)
         self.assertEqual(job2["eval_location"], dataset_eval.EVAL_LOCAL)
+
+    def test_get_pending_jobs_for_user(self):
+        job_id = dataset_eval._create_job(self.conn, self.test_dataset_id, True, dataset_eval.EVAL_REMOTE, None)
```
alastair: This is a good start, but you should consider other situations which could happen. What about:

 - If there is a local evaluation, make sure it doesn't get returned
 - If there is a remote evaluation which has already been processed, make sure it doesn't get returned
 - If there is a remote evaluation by another user, make sure it doesn't get returned

db/test/test_dataset_eval.py
```diff
@@ -162,3 +162,8 @@ def test_eval_job_location(self):
...
+        jobs = dataset_eval.get_pending_jobs_for_user(self.test_user_id)
+        self.assertEqual(jobs[0]["id"],str(job_id))
```
alastair: Remember formatting here, something isn't right.

webserver/__init__.py
```diff
@@ -103,8 +103,10 @@ def register_api(app):
         v1_prefix = '/api/v1'
         from webserver.views.api.v1.core import bp_core
         from webserver.views.api.v1.datasets import bp_datasets
+        from webserver.views.api.v1.dataset_eval import bp_dataset_eval
         app.register_blueprint(bp_core, url_prefix=v1_prefix)
         app.register_blueprint(bp_datasets, url_prefix=v1_prefix + '/datasets')
+        app.register_blueprint(bp_dataset_eval, url_prefix=v1_prefix + '/dataset-eval')
```
alastair: `/datasets/evaluation`, perhaps. @gentlecat ?

webserver/views/api/v1/dataset_eval.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import absolute_import
+from flask import Blueprint, jsonify, request
+from flask_login import current_user
+from webserver.decorators import auth_required
+import db.dataset_eval
+
+bp_dataset_eval = Blueprint('api_v1_dataset_eval', __name__)
+
+@bp_dataset_eval.route("/pending-jobs-for-user", methods=["GET"])
```
alastair: If we have to log in, it's implicit that it's for a user. Let's just call it `pending-jobs`

webserver/views/api/v1/test/test_dataset_eval.py
```diff
@@ -0,0 +1,34 @@
...
+        self.test_user = db.user.get(self.test_user_id)
+
+
+    @mock.patch("db.dataset_eval.get_pending_jobs_for_user")
+    def test_get_pending_jobs_for_user(self, get_pending_jobs_for_user):
+        self.temporary_login(self.test_user_id)
+
+        get_pending_jobs_for_user.return_value = [{'id': "7804abe5-58be-4c9c-a787-22b91d031489"}]
+        submit = json.dumps({})
+        resp = self.client.get("/api/v1/dataset-eval/pending-jobs-for-user", data=submit, content_type='application/json')
```
alastair: I think you only need to set `data` if this is a `POST`. for the `GET` you should just need the url.

webserver/views/api/v1/dataset_eval.py
```diff
@@ -0,0 +1,21 @@
...
+@auth_required
+def pending_jobs_for_user():
+    """Return a list of pending jobs for a user to be evaluated remotely"""
+    jobs = db.dataset_eval.get_pending_jobs_for_user(current_user.id)
+    jobs_list = []
+    for job in jobs:
+        jobs_list.append(job["id"])
+    return jsonify(
+        success=True,
+        jobs=jobs_list
```
alastair: I think we should return the username too. Perhaps:

```json
{"success": true,
 "response": {"user": "username",
                     "jobs": [job, list]}
}
```

webserver/__init__.py
gentlecat: Yeah, it's probably OK to have a separate blueprint, but let's keep URL structure easier to understand.

webserver/views/api/v1/dataset_eval.py
alastair: If there is more than one job to process, how will you know what order to process them in? In the order that they are listed?
Perhaps we can also return metadata for each job: Dataset name, date submitted, job id.

gentlecat:
You also need to add new blueprint into Sphinx docs.

webserver/views/api/v1/test/test_dataset_eval.py
```diff
@@ -0,0 +1,34 @@
...
+        self.test_user_mb_name = "tester"
+        self.test_user_id = db.user.create(self.test_user_mb_name)
+        self.test_user = db.user.get(self.test_user_id)
+
+
+    @mock.patch("db.dataset_eval.get_pending_jobs_for_user")
+    def test_get_pending_jobs_for_user(self, get_pending_jobs_for_user):
+        self.temporary_login(self.test_user_id)
+
+        get_pending_jobs_for_user.return_value = [{'id': "7804abe5-58be-4c9c-a787-22b91d031489"}]
```
alastair: Don't mix quotes. It doesn't matter which ones you use, as long as it's consistent

db/dataset_eval.py
kartikgupta0909: Yes that seems a better option

db/test/test_dataset_eval.py
kartikgupta0909: For the second case mentioned: By processed do you mean that the status of the job is 'done'?

webserver/__init__.py
kartikgupta0909: I ll change it to datasets/evaluation as the prefix for this blueprint.

webserver/views/api/v1/dataset_eval.py
gentlecat: Don't forget to document all the endpoint properly. See existing ones for an example. In this case also mention that only pending jobs for authenticated user are returned.

webserver/views/api/v1/dataset_eval.py
gentlecat: Also, how is this endpoint is actually supposed to be used? I assume that's for remote evaluation.

kartikgupta0909: I think we could have the username in the config file along with the API key, but no harm in returning it here too. 
About the metadata as I am already sorting the database query result on the basis of their created timestamp the date submitted might not be usefull as long as we maintain the order. For this I think we can just enumerate the output of the database query before sending it to the api function.

db/dataset_eval.py
```diff
@@ -298,6 +298,27 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
     job_id = result.fetchone()[0]
     return job_id
 
+def get_pending_jobs_for_user(user_id):
```
gentlecat: From the name of this function it's not obvious that you are actually getting just jobs for remote evaluation.

webserver/views/api/v1/test/test_dataset_eval.py
kartikgupta0909: Oh okay, I followed the other tests but they were post. I ll change it.

webserver/views/api/v1/dataset_eval.py
kartikgupta0909: This endpoint will be used by the client to get the list of pending jobs for a user so that it can process them one by one. Soon I will upload the first version of the client using this endpoint.

gentlecat: Perhaps it would make more sense to make a generic endpoint for getting all evaluation jobs with some filtering options (user, evaluation location). Otherwise it will be hard to change in the future without breaking the API. What if we decide that we need to list all jobs, but this endpoint already returns only ones that are specific to logged in user?

gentlecat: Even better would be to have something like `/datasets/eval-jobs`. Again, with status filter.

gentlecat: This can actually be used by our web interface if there's a part that can get data from the API in JS.

db/test/test_dataset_eval.py
alastair: Sure - there are many statuses that a job can have. I recommend you have one in pending, and one in not pending - it doesn't matter which one.

kartikgupta0909: Ah okay. Will add that test case.

webserver/views/api/v1/dataset_eval.py
alastair: @gentlecat This is a good point. This is specifically for jobs which have been marked as being evaluated on a remote server, so if we used it for the website as well we would need parameters to be able to select all, local, or remote jobs.
My preference is to make the endpoint do a specific thing but make its name clear that this is its task.
What are the requirements that we have on the website? For now is it just to show all jobs that have been submitted?

alastair: @kartikgupta0909 If we asked the user to add their name to the config file then there is nothing stopping them from making a mistake. Since we can already get this data from the API, I think it makes more sense to return it there.
Returning additional metadata can help us during development/debugging of the remote application. For example, we can log this information when we retrieve it.

Following on from @gentlecat's ideas, it would make sense to make all related endpoints return the same data. So, we could have a single endpoint with parameters, which we also use for the website, or separate endpoints which return specific data. In both cases, it makes sense to be consistent in what we return.

kartikgupta0909: Oh okay I ll add the metadata.

kartikgupta0909:
Updated the commit with the suggested changes.

webserver/views/api/v1/dataset_eval.py
```diff
@@ -0,0 +1,20 @@
...
+from webserver.decorators import auth_required
+import db.dataset_eval
+
+bp_dataset_eval = Blueprint('api_v1_dataset_eval', __name__)
+
+@bp_dataset_eval.route("/pending-jobs", methods=["GET"])
+@auth_required
+def pending_jobs():
+    """Return a list of pending jobs for a user to be evaluated remotely.
+       This method is valid only for Authenticated users."""
```
gentlecat: This part still needs to have better documentation. Provide information about return values. See https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/views/api/v1/datasets.py#L25-L67 for example.

gentlecat: Also, instead of saying `This method is valid only for Authenticated users.` you can just specify that API key argument is required.

webserver/views/api/v1/test/test_dataset_eval.py
```diff
@@ -0,0 +1,50 @@
...
+class APIDatasetEvaluationViewsTestCase(ServerTestCase):
+
+    def setUp(self):
+        super(APIDatasetEvaluationViewsTestCase, self).setUp()
+
+        self.test_user_mb_name = 'tester'
+        self.test_user_id = db.user.create(self.test_user_mb_name)
+        self.test_user = db.user.get(self.test_user_id)
+
+
```
gentlecat: > Method definitions inside a class are surrounded by a single blank line.
> *- http://pep8.org/#blank-lines*

db/dataset_eval.py
```diff
@@ -298,6 +299,40 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
...
+        jobs_list = []
+        for row in result.fetchall():
+            job_id = row[0]
+            dataset_name = row[1]
+            job_created = row[2]
+            jobs_list.append({
+                'job_id':job_id,
+                'dataset_name': dataset_name,
+                'job_created': job_created
+                })
```
gentlecat: You don't need to assign temporary variables here. Just extract from `row` directly into a dictionary.

db/dataset_eval.py
```diff
@@ -298,6 +299,40 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
...
+                        AND eval_location = :eval_location
+                        AND author = :user_id
+                   ORDER BY dataset_eval_jobs.created ASC
+            """)
+        result = connection.execute(query, {
+            "status": STATUS_PENDING,
+            "eval_location": EVAL_REMOTE,
+            "user_id": user_id
+        })
+        username = db.user.get(user_id)["musicbrainz_id"]
```
gentlecat: Same with this one. It's used once, why not get it below in `return` statement?

webserver/views/api/v1/test/test_dataset_eval.py
kartikgupta0909: I followed what was done in the other test files. Also the standards say that two lines may be used between sets of different types of functions. I think two lines here are justified because this is a setup function and other functions will be the actual testing functions and hence might be termed as different types of functions.

kartikgupta0909:
I have updated the PR, with an improved documentation of the API and other suggestions.

alastair:
@gentlecat 👀 ?

gentlecat:
LGTM now. 👀 :shipit:

alastair:
@kartikgupta0909 take a look at these changes that I made before merging. The two main items are:
 * Separation of roles in `db` and `webserver`. You had the db method return both the job list and the username, when the username was already available to be webserver (so we actually save 1 database query). The db method we have is named `get_remote_pending_jobs_for_user`, so we expect that we should return jobs, and not a username. If you did a similar method and you needed another value (for example, if you wanted both a list of datasets, and a list of pending jobs), you should call the two functions separately from the view, and combine their responses
* More complete tests: I mentioned previously a few options of things that you can test, and you added some, but not all of the suggestions. Make sure that each time there is a difference in data or input parameters that you test each of these possibilities.

PR #203, opened 2016-07-13
Make navbar more compact by moving some items into dropdowns

gentlecat:
This also fixes [AB-140](http://tickets.musicbrainz.org/browse/AB-140).

### DIFF
--- a/webserver/templates/navbar.html
+++ b/webserver/templates/navbar.html
@@ -17,12 +17,27 @@
     {#- Collect the nav links, forms, and other content for toggling -#}
     <div class="collapse navbar-collapse" id="ab-navbar-collapse">
       <ul class="nav navbar-nav">
-        <li><a href="{{ url_for('index.contribute') }}">Contribute</a></li>
+        <li class="dropdown">
+          <a class="dropdown-toggle" data-toggle="dropdown" href="#">
+            About <span class="caret"></span>
+          </a>
+          <ul class="dropdown-menu" role="menu">
+            <li><a href="{{ url_for('index.goals') }}">Goals</a></li>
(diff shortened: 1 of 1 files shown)


webserver/templates/navbar.html
```diff
@@ -17,12 +17,27 @@
...
         <li><a href="{{ url_for('index.downloads') }}">Downloads</a></li>
-        <li><a href="{{ url_for('index.goals') }}">Goals</a></li>
         <li><a href="{{ url_for('data.data') }}">API/Data</a></li>
-        <li><a href="{{ url_for('index.faq') }}">FAQ</a></li>
-        <li><a href="{{ url_for('stats.graph') }}">Detailed Statistics</a></li>
+        <li class="dropdown">
+          <a class="dropdown-toggle" data-toggle="dropdown" href="#">
+            Datasets <span class="caret"></span>
+          </a>
+          <ul class="dropdown-menu" role="menu">
```
gentlecat: Challenges will go inside this one as well.

webserver/templates/navbar.html
```diff
@@ -17,12 +17,27 @@
...
+            <li><a href="{{ url_for('datasets.list_datasets') }}">Dataset list</a></li>
```
alastair: Can we be more explicit about what this is? It's a list of proposed datasets? pending? User-submitted? I want to make it explicit that they're not live.

gentlecat: I think it would be better to call all datasets as "datasets" and ones that are live "active datasets" or something similar. There can be an indicator next to live dataset, but I don't see much point in specifying that they are not live here.

Also, one important distinction is that datasets can't be live, models can be and there are snapshots of datasets that are associated with them (new ones anyway).

PR #205, opened 2016-07-14
Save workspace of dataset evaluation/training

alastair:
By keeping the workspace created during dataset training we
retain access to:
 - The svm parameters used in this training process
 - Results for any combination of parameters, allowing us to go back
   and look for patterns in the results of a particular parameter
 - Analyse this data in the future to see if we can remove some
   parameters (speeding up the training process)
 - The actual classifications of the cross-validation step, which
   allows us to list items in the results (e.g. "this specific
   recording was incorrectly classified as 'vocals' when in the
   training set it was 'non-vocals')

We delete the yaml files, as these take up space and can easily be
recreated from the database.

I tested this manually. It successfully trains the model, saves data, copies history file, updates database, and removes yaml files.


dataset_eval/gaia_wrapper.py
```diff
@@ -104,7 +105,7 @@ def train_svm_history(project, params, output_file_path):
     gt.className = "highlevel." + project["className"]
 
     history = train_svm(ds, gt, **params_model)  # doing the whole training
-    history.save(output_file_path)
+    history.save(str(output_file_path))
```
gentlecat: Is that actually necessary?

dataset_eval/evaluate.py
```diff
@@ -50,25 +51,27 @@ def evaluate_dataset(eval_job):
         db.dataset_eval.add_sets_to_job(eval_job["id"], train, test)
 
         logging.info("Generating filelist.yaml and copying low-level data for evaluation...")
-        filelist_path = os.path.join(temp_dir, "filelist.yaml")
-        filelist = dump_lowlevel_data(train.keys(), os.path.join(temp_dir, "data"))
+        filelist_path = os.path.join(eval_location, "filelist.yaml")
+        data_dir = os.path.join(eval_location, "data")
```
gentlecat: Maybe it's better to keep it in `/temp`? In case script fails and it doesn't get to cleanup point.

gentlecat:
LGTM apart from two minor things that I mentioned.

dataset_eval/evaluate.py
alastair: oh yeah, good idea

dataset_eval/gaia_wrapper.py
alastair: the `str`? yes - this is passed into gaia, and the bindings only accept strings as arguments, not unicode. The original code `tempfile.mkdtemp()` returned a string, but now we create something with the job id, which is unicode (from the database). On second thoughts, we should probably use `.encode("utf-8")` here instead.

gentlecat:
LGTM

PR #206, opened 2016-07-20
Tell the user the reason for the validation error when evaling a dataset

alastair:
http://tickets.musicbrainz.org/browse/AB-96


gentlecat:
LGTM

PR #207, opened 2016-07-25
API endpoint for getting details of a job

kartikgupta0909:
This is related to http://tickets.musicbrainz.org/browse/AB-144

### DIFF
--- a/webserver/views/api/v1/dataset_eval.py
+++ b/webserver/views/api/v1/dataset_eval.py
@@ -4,6 +4,7 @@ from flask_login import current_user
 from webserver.decorators import auth_required
 from webserver.views.api import exceptions
 import db.dataset_eval
+import db.dataset
 
 bp_dataset_eval = Blueprint('api_v1_dataset_eval', __name__)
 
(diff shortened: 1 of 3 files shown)


webserver/views/api/v1/dataset_eval.py
```diff
@@ -0,0 +1,72 @@
...
+               "testing_snapshot": null,
+               "training_snapshot": null,
+               "updated": "Tue, 07 Jun 2016 22:12:32 GMT"
+           },
+          "success": true
+    }
+    """
+    job = db.dataset_eval.get_job(job_id)
+    job['dataset_details'] = db.dataset.get(job_details['dataset_id'])
+    return jsonify({'success' : True, 'job_details' : job})
```
gentlecat: Perhaps it would be a good idea to combine `dataset_id` and `dataset_details` into one item called `dataset`.

gentlecat: I'm also not sure about having `success` here.

alastair: hmm. good point. I suggested we add it to the endpoint which returned a list, for consistency with the database endpoints, and because `jsonify` won't (wouldn't) serialise a top-level list. 
If we want to be consistent, what should we say here? `'status': 'ok'`?

gentlecat: I actually don't think that we need it. The data is returned, so request has been successful. We can return `{"error": "Bad things happened."}` with an appropriate HTTP status if there is an issue.

Apart from this, I think it's better to just have data on the top level without wrapping it into `job_details`.

webserver/views/api/v1/dataset_eval.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import absolute_import
+from flask import Blueprint, jsonify, request
+from flask_login import current_user
+from webserver.decorators import auth_required
+import db.dataset_eval
+import db.dataset
+
+bp_dataset_eval = Blueprint('api_v1_dataset_eval', __name__)
+
+@bp_dataset_eval.route("/<uuid:job_id>/job-details", methods=["GET"])
```
gentlecat: How about making a URL `/datasets/eval/jobs/<job_id>`?

webserver/views/api/v1/dataset_eval.py
gentlecat: That's how I've been doing it anyway: https://beta.acousticbrainz.org/api/v1/datasets/f1a33fc5-e2b3-46b0-9fed-1bb1b83e99cd.

webserver/views/api/v1/dataset_eval.py
kartikgupta0909: Yes I guess the job-id can be put in the end, I did this way to follow the datasets API.

gentlecat: I think it's better that way because if you, let's say, go to `.../jobs`, you should get some kind  of list of evaluation jobs. Then you can append an ID to get more details about a specific one.

kartikgupta0909:
Updated the PR, with the following changes.
1. The url is updated to have the Job Id in the end, such as http://localhost:8080/api/v1/datasets/evaluation/jobs/7804abe5-58be-4c9c-a787-22b91d031489
2. Success and job_details removed to have data directly in the json response and not in a field.
3. 'dataset_details' changed to 'dataset'. I have kept the 'dataset_id'  in the outer level as well just to avoid confusion.

alastair:
@gentlecat 👀 ?

webserver/views/api/v1/dataset_eval.py
```diff
@@ -4,6 +4,8 @@
 from webserver.decorators import auth_required
 from webserver.views.api import exceptions
 import db.dataset_eval
+import db.dataset
+from . import datasets
```
gentlecat: It might be better to do an absolute import there. http://programmers.stackexchange.com/a/159505/36419

webserver/views/api/v1/dataset_eval.py
```diff
@@ -49,3 +51,67 @@ def get_jobs():
...
+           "status_msg": null,
+           "testing_snapshot": null,
+           "training_snapshot": null,
+           "updated": "Tue, 07 Jun 2016 22:12:32 GMT"
+    }
+    """
+    job = db.dataset_eval.get_job(job_id)
+    if not job:
+        raise exceptions.APINotFound('No such job')
+    else:
```
gentlecat: `else` is not necessary here.

alastair:
changes squashed into my first commit, rebased against master, and merged

gentlecat:
💥 👍 💥

PR #208, opened 2016-07-26
Pylint updates + pycodestyle

gentlecat:
* Updated version
* Added config file with some rules to ignore


gentlecat:
I added it as a separate Jenkins project for now. I guess we can combine it with existing one after all issues are fixed (or not, if it's useful to keep it as a separate check).

alastair:
cool 👍

PR #209, opened 2016-07-31
Get multiple recordings

kartikgupta0909:
This is related to http://tickets.musicbrainz.org/browse/AB-147

### DIFF
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import json
+import uuid
 
 from flask import Blueprint, request, jsonify
 
@@ -109,3 +110,37 @@ def _validate_offset(offset):
     else:
         offset = 0
     return offset
+
+
+@bp_core.route("/recordings", methods=["GET"])
+def get_recordings():
+    """Returns the low level details for all the recording-ids
+       mentioned in the request with their offset. Offsets are
+       optional. If no offset is given then a default value of
+       0 is used.
+    """
+    recording_ids = request.args.get("ids")
+    recordings = recording_ids.split(";")
+    recording_details = {}
+    if len(recordings) > 200:
+        raise webserver.views.api.exceptions.APIBadRequest("More than 200 recordings not allowed per request")
+    for recording in recordings:
+        recording = str(recording).split(":")
+        recording_id = recording[0]
+        try:
+            uuid.UUID(recording_id)
+        except ValueError:
+            raise webserver.views.api.exceptions.APIBadRequest("One or more recording ids invalid")
+        offset = 0
+        if len(recording) == 2:
+            try:
+                offset = int(recording[1])
+            except ValueError:
+                offset = 0
+        if recording_details.has_key(recording_id):
+            recording_details[recording_id][offset] = db.data.load_low_level(recording_id, offset)
+        else:
+            recording_details[recording_id] = {}
+            recording_details[recording_id][offset] = db.data.load_low_level(recording_id, offset)
+    return jsonify(recording_details)
+
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -112,35 +112,88 @@ def _validate_offset(offset):
     return offset
 
 
-@bp_core.route("/recordings", methods=["GET"])
-def get_recordings():
-    """Returns the low level details for all the recording-ids
-       mentioned in the request with their offset. Offsets are
-       optional. If no offset is given then a default value of
-       0 is used.
+def _valididate_bulk_params(params):
+    """Validate a bulk query parameter string.
+
+    A valid parameter string takes the form
+      mbid[:offset];mbid[:offset];...
+    Where offset is a number >=0. Offsets are optional.
+    mbid must be a UUID.
+
+    If an offset is not specified or is invalid, 0 is used as a default.
+    If an mbid is not valid, an APIBadRequest Exception is
+    raised listing the bad MBID and no further processing is done.
+
+    Returns a list of tuples (mbid, offset)
     """
-    recording_ids = request.args.get("ids")
-    recordings = recording_ids.split(";")
-    recording_details = {}
-    if len(recordings) > 200:
-        raise webserver.views.api.exceptions.APIBadRequest("More than 200 recordings not allowed per request")
+
+    ret = []
+    recordings = params.split(";")
+
     for recording in recordings:
-        recording = str(recording).split(":")
-        recording_id = recording[0]
+        parts = str(recording).split(":")
+        recording_id = parts[0]
         try:
             uuid.UUID(recording_id)
         except ValueError:
-            raise webserver.views.api.exceptions.APIBadRequest("One or more recording ids invalid")
-        offset = 0
-        if len(recording) == 2:
+            raise webserver.views.api.exceptions.APIBadRequest("'%s' is not a valid UUID" % recording_id)
+        if len(parts) > 2:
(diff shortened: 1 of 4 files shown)


webserver/views/api/v1/dataset_eval.py
```diff
@@ -2,71 +2,72 @@
...
+        raise exceptions.APIBadRequest("parameter location must be 'remote' and status 'pending'")
+
+    jobs = db.dataset_eval.get_remote_pending_jobs_for_user(current_user.id)
+    return jsonify(
+        username=current_user.musicbrainz_id,
+        jobs=jobs
+    )
+
+@bp_dataset_eval.route("/bulk-recordings/<recording_ids>", methods=["GET"])
+def bulk_recordings(recording_ids):
```
gentlecat: I would suggest to reuse existing endpoint by providing an option to specify list of IDs as an argument. I don't like the trend of creating more and more endpoints that provide multiple ways to get the same data. There should be one place to get a particular entity.

Think about how people are going to use the API, not just a client that you are building. It's hard to change things in the API once they are there.

alastair: yeah, good idea. This is why I linked this ticket with http://tickets.musicbrainz.org/browse/AB-21 in jira. Let's add this functionality to the standard low-level and high-level endpoints in `core.py`.

webserver/views/api/v1/dataset_eval.py
```diff
@@ -2,71 +2,72 @@
...
+       0 is used.
     """
-    job = db.dataset_eval.get_job(job_id)
-    job['dataset_details'] = db.dataset.get(job_details['dataset_id'])
-    return jsonify({'success' : True, 'job_details' : job})
+    recordings = recording_ids.split(";")
+    recording_details = {}
+    if len(recordings) > 200:
+        return jsonify({
+            'error': 'Sorry you exceeded the no. of allowed recordings per request.'
```
alastair: You should say what the limit is in this error.
This should be an HTTP400 error, not 200. You can get the right behaviour by raising one of our ApiExceptions

webserver/views/api/v1/dataset_eval.py
```diff
@@ -2,71 +2,72 @@
...
+            })
+    for recording in recordings:
+        recording = str(recording).split(":")
+        recording_id = recording[0]
+        offset = 0
+        if len(recording) == 2:
+            offset = int(recording[1])
```
alastair: If this isn't a number we should also use offset=0, use try/catch ValueError

webserver/views/api/v1/dataset_eval.py
```diff
@@ -2,71 +2,72 @@
...
-    return jsonify({'success' : True, 'job_details' : job})
+    recordings = recording_ids.split(";")
+    recording_details = {}
+    if len(recordings) > 200:
+        return jsonify({
+            'error': 'Sorry you exceeded the no. of allowed recordings per request.'
+            })
+    for recording in recordings:
+        recording = str(recording).split(":")
+        recording_id = recording[0]
```
alastair: In the core endpoint we use `<uuid:mbid>` in the url, which lets us verify that the argument is a valid uuid (regardless of if it is actually in the database). We should check that this is a uuid here too. You can use
```python
try:
    uuid.UUID(recording_id)
except ValueError:
  # not a uuid
```

webserver/views/api/v1/dataset_eval.py
```diff
@@ -2,71 +2,72 @@
...
+        offset = 0
+        if len(recording) == 2:
+            offset = int(recording[1])
+        recording_details[recording_id] = db.data.load_low_level(recording_id, offset)
```
alastair: If I ask for recording1 offset0 and recording1 offset1 then I'll only get one of them, because your key is the same.
This comes back to a similar question about our lists and single items, and the `success` field. I'm wondering about the best way to do this:

```javascript
{"items": [ {"mbid": x, "offset": n, "item": {...}}, {"mbid": x, "offset": n, "item": {...}}}
```
The idea behind this is that we never return a list, due to the restriction in flask.

But what should this endpoint be if we only specify one mbid? Consistency dictates that we keep the surrounding data, or have a list of 1 item. However, I still think that our `/<mbid>/low-level` should return a single item. @gentlecat, thoughts?

webserver/views/api/v1/dataset_eval.py
gentlecat: So kind of like an existing endpoint, but instead of having ID as a part of URL just have `ids` argument.

webserver/views/api/v1/dataset_eval.py
```diff
@@ -2,71 +2,72 @@
...
+def bulk_recordings(recording_ids):
+    """Returns the low level details for all the recording-ids
+       mentioned in the request with their offset. Offsets are 
+       optional. If no offset is given then a default value of
+       0 is used.
     """
-    job = db.dataset_eval.get_job(job_id)
-    job['dataset_details'] = db.dataset.get(job_details['dataset_id'])
-    return jsonify({'success' : True, 'job_details' : job})
+    recordings = recording_ids.split(";")
```
gentlecat: I think it's more common to use commas as separators.

alastair: I suggested ; in http://tickets.musicbrainz.org/browse/AB-147, because we use it in MB: http://musicbrainz.org/doc/Development/XML_Web_Service/Version_2#collections

gentlecat: Ah, sorry, didn't see that.

webserver/views/api/v1/dataset_eval.py
kartikgupta0909: Okay will incorporate that into the core API endpoint. @alastair what are your views about putting the recording ids as a part of the url or as a parameter?

webserver/views/api/v1/dataset_eval.py
kartikgupta0909: I couldnt do this in the URL itself, because it was a list of more than one Ids and also had the offsets. But can use try/catch.

kartikgupta0909:
I have updated this PR with the changes suggested in the comments and the ones discussed on the irc.
1. Now we have a core api endpoint /api/v1/recordings?ids=mbid1:offset1;mbid2:offset2...
2. The response is of the folllowing format:
    {
        mbid1 : 
            {
                 offset1: {}
                 offset2_for_mbid1 : {}
             }
    }

webserver/views/api/v1/core.py
```diff
@@ -109,3 +110,36 @@ def _validate_offset(offset):
...
+       optional. If no offset is given then a default value of
+       0 is used.
+    """
+    recording_ids = request.args.get("ids")
+    recordings = recording_ids.split(";")
+    recording_details = {}
+    if len(recordings) > 200:
+        raise webserver.views.api.exceptions.APIBadRequest("More than 200 recordings not allowed per request")
+    for recording in recordings:
+        recording = str(recording).split(":")
```
gentlecat: So specifying an offset is mandatory? I don't think this is a good idea.

kartikgupta0909: No its not, Its optional. We are checking first if there are two parameters (id and an offset), if there is no offset we set a default offset of 0.

gentlecat: Ah, good. I see the comparison below now.

webserver/views/api/v1/dataset_eval.py
gentlecat: Is there any specific reason we need to follow this though if offsets are separated by `:`?

alastair:
This is merged and released!

PR #210, opened 2016-08-12
Schema change for user feedback feature

gentlecat:


gentlecat:
I'm going to merge this since I'm short on time and need to start testing feedback stuff on beta very soon. There are no changes to existing tables, this just adds another one.

But please take a look afterwards.

admin/sql/create_tables.sql
```diff
@@ -147,4 +147,11 @@ CREATE TABLE api_key (
   created   TIMESTAMP WITH TIME ZONE DEFAULT NOW()
 );
 
+CREATE TABLE feedback (
+  user_id            INTEGER, -- PK, FK to user
+  highlevel_model_id INTEGER, -- PK, FK to highlevel_model
+  correct            BOOLEAN NOT NULL,
```
alastair: This is OK for a 2-class problem, but what if it has more? Do we want a user to be able to select a specific correct answer?

admin/sql/create_tables.sql
```diff
@@ -147,4 +147,11 @@ CREATE TABLE api_key (
...
+  suggestion         TEXT
```
alastair: Is this freeform?

admin/sql/create_tables.sql
```diff
@@ -147,4 +147,11 @@ CREATE TABLE api_key (
   created   TIMESTAMP WITH TIME ZONE DEFAULT NOW()
 );
 
+CREATE TABLE feedback (
```
alastair: can we make the table name more specific? `highlevel_feedback` ?

admin/sql/create_tables.sql
gentlecat: This one is to know if the output is correct or not. Not to specify an expected result.

admin/sql/create_tables.sql
gentlecat: No, name of the expected class is supposed to go there. Perhaps I can rename it to `expected_class` to be more specific.

admin/sql/create_tables.sql
gentlecat: Sounds good.

admin/sql/create_tables.sql
alastair: 👍

admin/sql/create_tables.sql
alastair: 👍

admin/sql/create_tables.sql
alastair: 👍

PR #212, opened 2016-10-15
Dataset api part2 vibhor

vibhorbajpai:
Hi. This is a commit for the add/delete recordings. The files commited are: /webserver/views/api/v1/datasets.py
/webserver/views/api/v1/test/test_datasets.py
/db/dataset.py
/db/test/test_dataset.py
Thanks and regards,
Vibhor Bajpai

### DIFF
--- a/webserver/views/api/v1/datasets.py
+++ b/webserver/views/api/v1/datasets.py
@@ -97,6 +97,152 @@ def delete_dataset(dataset_id):
     )
 
 
+@bp_datasets.route("/<uuid:dataset_id>", methods=["PUT"])
+@auth_required
+def update_dataset_details(dataset_id):
+    """Update dataset details.
+
+    If one of the fields is not specified, it will not be updated.
+
+    **Example request**:
+
+    .. sourcecode:: json
+
+        {
+            "name": "Not Mood",
+            "description": "Dataset for mood misclassification.",
+            "public": true
+        }
+
+    :reqheader Content-Type: *application/json*
+    :<json string name: *Optional.* Name of the dataset.
+    :<json string description: *Optional.* Description of the dataset.
+    :<json boolean public: *Optional.* ``True`` to make dataset public, ``false`` to make it private.
+
+    :resheader Content-Type: *application/json*
+    """
+    raise NotImplementedError
+
+
+@bp_datasets.route("/<uuid:dataset_id>/classes", methods=["POST"])
+@auth_required
+def add_class(dataset_id):
+    """Add class into a dataset.
+
+    **Example request**:
+
+    .. sourcecode:: json
+
+        {
+            "name": "Not Mood",
+            "description": "Dataset for mood misclassification.",
+        }
+
+    :reqheader Content-Type: *application/json*
+    :<json string name: *Required.* Name of the class. Must be unique within a dataset.
+    :<json string description: *Optional.* Description of the class.
+    :<json array public: *Optional.* Array of recording MBIDs (``string``) to add into that class. For example:
+        ``["770cc467-8dde-4d22-bc4c-a42f91e"]``.
+
+
+    :resheader Content-Type: *application/json*
+    """
+    raise NotImplementedError
+
+
+@bp_datasets.route("/<uuid:dataset_id>/classes", methods=["PUT"])
+@auth_required
+def update_class(dataset_id):
+    """Update class in a dataset.
+
+    If one of the fields is not specified, it will not be updated.
+
+    **Example request**:
+
+    .. sourcecode:: json
+
+        {
+            "name": "Very happy",
+            "new_name": "Recordings that represent ultimate happiness."
+        }
+
+    :reqheader Content-Type: *application/json*
+    :<json string name: *Required.* Current name of the class.
+    :<json string new_name: *Optional.* New name of the class. Must be unique within a dataset.
+    :<json string description: *Optional.* Description of the class.
+
+    :resheader Content-Type: *application/json*
+    """
+    raise NotImplementedError
+
+
+@bp_datasets.route("/<uuid:dataset_id>/classes", methods=["DELETE"])
+@auth_required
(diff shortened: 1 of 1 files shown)


gentlecat:
You should preserve original authors of commits (in this case 6276440500296e4fb273f1761a59be50bd10e7a5). That's our only way of knowing who added what.

db/test/test_dataset.py
```diff
@@ -119,6 +123,22 @@ def test_get_by_user_id(self):
         datasets = dataset.get_by_user_id(self.test_user_id, public_only=False)
         self.assertEqual(len(datasets), 2)
 
+    def test_add_recordings(self):
+
+        id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        ds = dataset.get(id)
+
+        dataset.add_recordings(self.test_data_recordings)
```
alastair: The reason for adding tests is to perform some action and then to verify that this action has been successfully applied.
In this case you are calling a method but then not doing anything with it.

For me, a test for this method would do something like this:

1. Create a test dataset
2. Call the `add_recordings` method to add some new recording ids to a class
3. Get the dataset from the database again and check that it contains both the recordings which were originally in it plus the new ones.
4. An optional test, related to an earlier comment: Check that if there is a separate dataset with the same named class, that it was not affected.

Two other details in this method:
 * You call `ds = dataset.get(id)` but do nothing with this data. Make sure you don't have unused code in methods
 * We use dictionaries in `self` for data which is shared between many tests. In this case, `self.test_data` is used in many places and so we initialise it in the `setUp` method. In the case of the `test_data_recordings` dictionary, this is only used in this test, so you can define it in the method.

db/test/test_dataset.py
```diff
@@ -119,6 +123,22 @@ def test_get_by_user_id(self):
...
+
+    def test_delete_recordings(self):
+
+        id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        ds = dataset.get(id)
+
+        for cls in ds["classes"]:
+            dataset.delete_recordings(cls)
+
```
alastair: Similar to the add method, you should test that the delete call actually did something

webserver/views/api/v1/datasets.py
```diff
@@ -97,6 +97,165 @@ def delete_dataset(dataset_id):
...
+    :<json string class_name: *Required.* Name of the class.
+    :<json array recordings: *Required.* Array of recoding MBIDs (``string``) to add into that class.
+
+    :resheader Content-Type: *application/json*
+    """
+
+    #The json in this method's example request is given to be 'class_name' instead of 'name' which is in other json requests.
+
+    ds = get_dataset(dataset_id)
+    db.dataset.add_recordings(ds)
```
alastair: This isn't doing the correct task.

In URL endpoints we can get data from two places. From the URL, and from the request payload.
In this case with a url such as
`/api/v1/datasets/b86df890-32e0-4550-bc1b-e6e99414c135/recordings`
the `route` function will extract part of the url and put it in the `dataset_id` variable

The request payload is sent separately, and you access it with `request.get_json()`: https://github.com/vibhorbajpai/acousticbrainz-server/blob/6bd08b330251e0a1cf7a9d06280e84db66370bf6/webserver/views/api/v1/datasets.py#L68

You appear to have confused the request payload with the data which already exists in the database. In this case, you are getting the contents of a dataset and then using it to call the `db.dataset.add_recordings` method. These dictionaries are different, so I don't think that the `add_recordings` method will work.

The response of `get_dataset` is something like this
```json
{
  "author": 1,
  "classes": [
    {"description": null,
     "id": "106",
     "name": "classthree",
     "recordings": ["075a3b50-4650-4fb6-8c99-66e1af7bb07e"]
    }],
  "created": "Thu, 23 Jun 2016 13:36:01 GMT",
  "description": "",
  "id": "f67afc32-e417-4304-9ed9-8eb41a800dc8",
  ... other data removed ...
}
```

and the format which you expect has `recordings` and `class_name` fields: https://github.com/vibhorbajpai/acousticbrainz-server/blob/6bd08b330251e0a1cf7a9d06280e84db66370bf6/db/dataset.py#L65

This method should instead do something like get the dataset, get the request body, validate that the format of the body is correct, and call `add_recordings` passing in both the dataset id and the request body.

In these methods you have to check that the dataset exists and that the current user has permission to edit this dataset. Instead of `get_dataset` we have a helper which you can use, `get_check_dataset`

webserver/views/api/v1/test/test_datasets.py
```diff
@@ -79,3 +107,33 @@ def test_create_dataset(self, create_from_dict):
...
+        id = db.dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        ds = db.dataset.get(id)
+        self.temporary_login(self.test_user_id)
+        submit = json.dumps({"class_name": "Class1",
+            "recordings": ["1c085555-3805-428a-982f-e14e0a2b18e6"]
+        })
+        url = '/api/v1/datasets/%s/recordings' %(str(id))
+        resp = self.client.put(url, data=submit, content_type='application/json')
+        self.assertEqual(resp.status_code, 200)
+        expected = {"success": True, "message": "Recording(s) added."}
```
alastair: Here you are testing that the response from the client is correct, but like your db tests you are not checking the side effects of this call. In python if you are using mocks you can use the `assert_called_once_with` method: https://github.com/vibhorbajpai/acousticbrainz-server/blob/6bd08b330251e0a1cf7a9d06280e84db66370bf6/webserver/views/api/v1/test/test_datasets.py#L96

In the API tests we use mocks so that we don't access the database. This way we can have two separate types of tests: one to check that the database access methods correctly read and write to the database, and one to check that the API correctly calls the database methods. In the API tests we mock the database access methods. This means that you don't need to populate the database here (e.g. with `self.test_data`)

gentlecat:
@brainzbot, ok to test.

PR #213, opened 2016-10-28
Add/Delete recordings changes for dataset-api-part2

vibhorbajpai:
Hi,
Please find the relevant files for the add and delete recordings functionality.
Thanks and regards,
Vibhor Bajpai

### DIFF
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -321,21 +321,23 @@ def add_recordings(dataset, dataset_id):
     """Adds new recordings to a class in a dataset"""
 
     with db.engine.begin() as connection:
-        result = connection.execute("""SELECT id FROM dataset_class WHERE name = %s and dataset = %s""",
-                                    (dataset["class_name"], dataset_id))
+        result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
+                                    {"name": dataset["class_name"], "dataset_id": dataset_id})
+        if result.rowcount < 1:
+            raise exceptions.NoDataFoundException("No such class exists.")
         clsid = result.fetchone()
-
         for mbid in dataset["recordings"]:
-            connection.execute("""INSERT INTO dataset_class_member (class, mbid) VALUES (%s,%s)""", (clsid[0],mbid))
+            connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:clsid,:mbid)"""), {"clsid": clsid[0], "mbid": mbid})
 
 
 def delete_recordings(dataset, dataset_id):
     """Delete all recordings from the list of IDs"""
 
     with db.engine.begin() as connection:
-        result = connection.execute("""SELECT id FROM dataset_class WHERE name = %s and dataset = %s""",
-                                    (dataset["class_name"], dataset_id))
+        result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
+                                    {"name": dataset["class_name"], "dataset_id": dataset_id})
         clsid = result.fetchone()
-
+        if not clsid:
+            raise db.exceptions.NoDataFoundException("No such class exists.")
         for mbid in dataset["recordings"]:
             connection.execute("""DELETE FROM dataset_class_member WHERE class = %s and mbid = %s""", (clsid[0], mbid))
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -317,30 +317,47 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
     connection.execute(query, {"dataset_id": dataset_id})
 
 
-def add_recordings(dict, dataset_id):
+def add_recordings(dataset_id, class_name, recordings):
     """Adds new recordings to a class in a dataset"""
 
     with db.engine.begin() as connection:
-        result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
-                                    {"name": dict["class_name"], "dataset_id": dataset_id})
+        result = connection.execute(sqlalchemy.text("""
+          SELECT id FROM dataset_class
+              WHERE name = :name
+                  and dataset = :dataset_id
+                      """),
+                            {"name": class_name, "dataset_id": dataset_id})
         if result.rowcount < 1:
             raise exceptions.NoDataFoundException("No such class exists.")
         clsid = result.fetchone()
-        for mbid in dict["recordings"]:
-            connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:clsid, :mbid)"""), {"clsid": clsid[0], "mbid": mbid})
+        for mbid in recordings:
+            connection.execute(sqlalchemy.text("""
+              INSERT INTO dataset_class_member (class, mbid)
+                  VALUES (:clsid, :mbid)
+                      """),
+                            {"clsid": clsid[0], "mbid": mbid})
 
 
-def delete_recordings(dict, dataset_id):
+def delete_recordings(dataset_id, class_name, recordings):
     """Delete all recordings from the list of IDs"""
 
     with db.engine.begin() as connection:
-        result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
-                                    {"name": dict["class_name"], "dataset_id": dataset_id})
+        result = connection.execute(sqlalchemy.text("""
+                    SELECT id FROM dataset_class
+                        WHERE name = :name
+                            and dataset = :dataset_id
+                                """),
+                                    {"name": class_name, "dataset_id": dataset_id})
         if result.rowcount < 1:
             raise exceptions.NoDataFoundException("No such class exists.")
         clsid = result.fetchone()
-        for mbid in dict["recordings"]:
-            connection.execute(sqlalchemy.text("""DELETE FROM dataset_class_member WHERE class = :class_name and mbid = :mbid_num"""), {"class_name": clsid[0],"mbid_num": mbid})
+        for mbid in recordings:
+            connection.execute(sqlalchemy.text("""
+                DELETE FROM dataset_class_member
+                    WHERE class = :class_name
+                        and mbid = :mbid_num
+                            """),
+                               {"class_name": clsid[0],"mbid_num": mbid})
 
 
 def add_class(dict, dataset_id):
@@ -349,20 +366,24 @@ def add_class(dict, dataset_id):
     with db.engine.begin() as connection:
         if "description" not in dict:
             dict["description"] = None
-        result = connection.execute(sqlalchemy.text("""INSERT INTO dataset_class (name, description, dataset) VALUES (:name, :description, :datasetid) RETURNING id"""),
-                                    {"name": dict["name"], "description": dict["description"], "datasetid": dataset_id})
-        if result.rowcount < 1:
-            raise exceptions.NoDataFoundException("No such data exists.")
-        cls_id = result.fetchone()[0]
-        if "recordings" in dict:
-            for mbid in dict["recordings"]:
-                connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:class_name, :mbid_num)"""),
-                                   {"class_name": cls_id, "mbid_num": mbid})
+        result = connection.execute(sqlalchemy.text("""
+            INSERT INTO dataset_class (name, description, dataset)
+                SELECT :name, :description, :datasetid
+                    WHERE NOT EXISTS
+                        (select * from dataset_class d where d.name = :name and d.dataset = :datasetid)
+                    """),
+                        {"name": dict["name"], "description": dict["description"], "datasetid": dataset_id})
+    if "recordings" in dict:
+        add_recordings(dataset_id, dict["name"], dict["recordings"])
 
 
 def delete_class(dict, dataset_id):
     """Delete a class from a dataset"""
 
     with db.engine.begin() as connection:
-        connection.execute(sqlalchemy.text("""DELETE FROM dataset_class WHERE name = :class_name and dataset = :dataset_id"""),
-                                    {"class_name": dict["name"], "dataset_id": dataset_id})
+        connection.execute(sqlalchemy.text("""
+            DELETE FROM dataset_class
+                WHERE name = :class_name
+                    and dataset = :dataset_id
+                        """),
+                            {"class_name": dict["name"], "dataset_id": dataset_id})
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -336,8 +336,8 @@ def delete_recordings(dataset, dataset_id):
     with db.engine.begin() as connection:
         result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
                                     {"name": dataset["class_name"], "dataset_id": dataset_id})
+        if result.rowcount < 1:
+            raise exceptions.NoDataFoundException("No such class exists.")
         clsid = result.fetchone()
-        if not clsid:
-            raise db.exceptions.NoDataFoundException("No such class exists.")
         for mbid in dataset["recordings"]:
             connection.execute("""DELETE FROM dataset_class_member WHERE class = %s and mbid = %s""", (clsid[0], mbid))
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -315,3 +315,27 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
         DELETE FROM dataset_snapshot
               WHERE dataset_id = :dataset_id""")
     connection.execute(query, {"dataset_id": dataset_id})
+
+
+def add_recordings(dataset, dataset_id):
+    """Adds new recordings to a class in a dataset"""
+
+    with db.engine.begin() as connection:
+        result = connection.execute("""SELECT id FROM dataset_class WHERE name = %s and dataset = %s""",
+                                    (dataset["class_name"], dataset_id))
+        clsid = result.fetchone()
+
+        for mbid in dataset["recordings"]:
+            connection.execute("""INSERT INTO dataset_class_member (class, mbid) VALUES (%s,%s)""", (clsid[0],mbid))
+
+
+def delete_recordings(dataset, dataset_id):
+    """Delete all recordings from the list of IDs"""
+
+    with db.engine.begin() as connection:
+        result = connection.execute("""SELECT id FROM dataset_class WHERE name = %s and dataset = %s""",
+                                    (dataset["class_name"], dataset_id))
+        clsid = result.fetchone()
+
+        for mbid in dataset["recordings"]:
+            connection.execute("""DELETE FROM dataset_class_member WHERE class = %s and mbid = %s""", (clsid[0], mbid))
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -333,7 +333,9 @@ def add_recordings(dataset_id, class_name, recordings):
         for mbid in recordings:
             connection.execute(sqlalchemy.text("""
               INSERT INTO dataset_class_member (class, mbid)
-                  VALUES (:clsid, :mbid)
+                  SELECT :clsid, :mbid
+                      WHERE NOT EXISTS
+                          (select * from dataset_class_member d where d.class = :clsid and d.mbid = :mbid)
                       """),
                             {"clsid": clsid[0], "mbid": mbid})
 
@@ -366,7 +368,7 @@ def add_class(dict, dataset_id):
     with db.engine.begin() as connection:
         if "description" not in dict:
             dict["description"] = None
-        result = connection.execute(sqlalchemy.text("""
+        connection.execute(sqlalchemy.text("""
             INSERT INTO dataset_class (name, description, dataset)
                 SELECT :name, :description, :datasetid
                     WHERE NOT EXISTS
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -317,27 +317,52 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
     connection.execute(query, {"dataset_id": dataset_id})
 
 
-def add_recordings(dataset, dataset_id):
+def add_recordings(dict, dataset_id):
     """Adds new recordings to a class in a dataset"""
 
     with db.engine.begin() as connection:
         result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
-                                    {"name": dataset["class_name"], "dataset_id": dataset_id})
+                                    {"name": dict["class_name"], "dataset_id": dataset_id})
         if result.rowcount < 1:
             raise exceptions.NoDataFoundException("No such class exists.")
         clsid = result.fetchone()
-        for mbid in dataset["recordings"]:
-            connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:clsid,:mbid)"""), {"clsid": clsid[0], "mbid": mbid})
+        for mbid in dict["recordings"]:
+            connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:clsid, :mbid)"""), {"clsid": clsid[0], "mbid": mbid})
 
 
-def delete_recordings(dataset, dataset_id):
+def delete_recordings(dict, dataset_id):
     """Delete all recordings from the list of IDs"""
 
     with db.engine.begin() as connection:
         result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
-                                    {"name": dataset["class_name"], "dataset_id": dataset_id})
+                                    {"name": dict["class_name"], "dataset_id": dataset_id})
         if result.rowcount < 1:
             raise exceptions.NoDataFoundException("No such class exists.")
         clsid = result.fetchone()
-        for mbid in dataset["recordings"]:
-            connection.execute("""DELETE FROM dataset_class_member WHERE class = %s and mbid = %s""", (clsid[0], mbid))
+        for mbid in dict["recordings"]:
+            connection.execute(sqlalchemy.text("""DELETE FROM dataset_class_member WHERE class = :class_name and mbid = :mbid_num"""), {"class_name": clsid[0],"mbid_num": mbid})
+
+
+def add_class(dict, dataset_id):
+    """Add a class to a dataset"""
+
+    with db.engine.begin() as connection:
+        if "description" not in dict:
+            dict["description"] = None
+        result = connection.execute(sqlalchemy.text("""INSERT INTO dataset_class (name, description, dataset) VALUES (:name, :description, :datasetid) RETURNING id"""),
+                                    {"name": dict["name"], "description": dict["description"], "datasetid": dataset_id})
+        if result.rowcount < 1:
+            raise exceptions.NoDataFoundException("No such data exists.")
+        cls_id = result.fetchone()[0]
+        if "recordings" in dict:
+            for mbid in dict["recordings"]:
+                connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:class_name, :mbid_num)"""),
+                                   {"class_name": cls_id, "mbid_num": mbid})
+
+
+def delete_class(dict, dataset_id):
+    """Delete a class from a dataset"""
+
+    with db.engine.begin() as connection:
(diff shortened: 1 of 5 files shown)


db/dataset.py
```diff
@@ -315,3 +315,27 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
     connection.execute(query, {"dataset_id": dataset_id})
+
+
+def add_recordings(dataset, dataset_id):
+    """Adds new recordings to a class in a dataset"""
+
+    with db.engine.begin() as connection:
+        result = connection.execute("""SELECT id FROM dataset_class WHERE name = %s and dataset = %s""",
+                                    (dataset["class_name"], dataset_id))
+        clsid = result.fetchone()
```
alastair: If there is no class with this name, then `clsid` will be `None`. You need to check this here before you use it to insert the new recording ids

db/dataset.py
```diff
@@ -315,3 +315,27 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
               WHERE dataset_id = :dataset_id""")
     connection.execute(query, {"dataset_id": dataset_id})
+
+
+def add_recordings(dataset, dataset_id):
+    """Adds new recordings to a class in a dataset"""
+
+    with db.engine.begin() as connection:
+        result = connection.execute("""SELECT id FROM dataset_class WHERE name = %s and dataset = %s""",
+                                    (dataset["class_name"], dataset_id))
```
alastair: Please look at the other SQL queries in this file, and make sure that you use `sqlalchemy.text` and format the queries over multiple lines

db/dataset.py
```diff
@@ -315,3 +315,27 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+        clsid = result.fetchone()
+
+        for mbid in dataset["recordings"]:
```
alastair: If a recording id in this list is already in this class we should not add it. I think that we should check this in the view.

db/dataset.py
```diff
@@ -315,3 +315,27 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+            connection.execute("""INSERT INTO dataset_class_member (class, mbid) VALUES (%s,%s)""", (clsid[0],mbid))
+
+
+def delete_recordings(dataset, dataset_id):
```
alastair: The same points I made in the `add_recordings` method also apply in this method

db/test/test_dataset.py
```diff
@@ -169,6 +169,29 @@ def test_delete_shapshot(self):
...
+    def test_add_recordings(self):
+        id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        test_data_recordings = {
+            "class_name": "Class #1",
+            "recordings": ["1c085555-3805-428a-982f-e14e0a2b18e6"]
+        }
+        originaldataset = dataset.get(id)
+        dataset.add_recordings(test_data_recordings, id)
+        updateddataset = dataset.get(id)
+        self.assertDictEqual(originaldataset["classes"][1],updateddataset["classes"][1])
```
alastair: Be careful about formatting here, there should be a space after the comma

db/test/test_dataset.py
```diff
@@ -169,6 +169,29 @@ def test_delete_shapshot(self):
...
+        self.assertEqual(len(originaldataset["classes"][0]["recordings"])+1,len(updateddataset["classes"][0]["recordings"]))
```
alastair: This is a good test. Instead of checking the length of the datasets, you could explicitly check the contents of the recording list.

alastair: Can we also have a test for `add_recordings` which tries to add a class with the same name to a different dataset, to make sure that it only gets added to one?

webserver/views/api/v1/datasets.py
```diff
@@ -217,7 +217,18 @@ def add_recordings(dataset_id):
...
     :resheader Content-Type: *application/json*
     """
-    raise NotImplementedError
+    class_dict = request.get_json()
+    if not class_dict:
+        raise api_exceptions.APIBadRequest("Data must be submitted in JSON format.")
+    try:
+        db.dataset.add_recordings(class_dict, dataset_id)
+    except dataset_validator.ValidationException as e:
+        raise api_exceptions.APIBadRequest(e.message)
```
alastair: You are catching an exception here, but this exception is not raised in the dataset method.

webserver/views/api/v1/datasets.py
```diff
@@ -217,7 +217,18 @@ def add_recordings(dataset_id):
 
     :resheader Content-Type: *application/json*
     """
-    raise NotImplementedError
+    class_dict = request.get_json()
+    if not class_dict:
+        raise api_exceptions.APIBadRequest("Data must be submitted in JSON format.")
```
alastair: We should also check the contents of this dictionary before we do something with it. Does it have the required `class_name` and `recordings` fields? Are all of the recordings ids valid UUIDs?
If the class doesn't exist in the dataset (but the dataset does exist), should we automatically create the class, or is it an error?

webserver/views/api/v1/datasets.py
```diff
@@ -217,7 +217,18 @@ def add_recordings(dataset_id):
...
+    try:
+        db.dataset.add_recordings(class_dict, dataset_id)
```
alastair: This doesn't check who the owner of the dataset is, which means that anyone can modify someone else's dataset. You should use the `get_check_dataset` method to perform this check.

gentlecat:
@brainzbot, add to whitelist.

db/dataset.py
vibhorbajpai: If there are multiple recording ids in the json list and one of them is already there in the class  in database, do we raise an error without inserting any of the recording ids, or should we modify the json by removing the incorrect (duplicate) recording id and going ahead by adding rest of them? Also, is a scenario possible where the duplicate recording ids are present within the json list itself? 
For example, can we have a json like the following:
{
            "class_name": "Happy",
            "recordings": ["770cc467-8dde-4d22-bc4c-a42f91e", "770cc467-8dde-4d22-bc4c-a42f91e"]
        }
Do we need to put a check for this as well?

webserver/views/api/v1/datasets.py
vibhorbajpai: I have added a test for valid UUIDs, however, while creating a json request while testing, it automatically checks whether the recordings are a valid UUID. Does it need to be checked again in the view? The same goes for duplicate recordings in the JSON. While making the json request itself the program throws an error if there are duplicate recordings in the JSON.

db/test/test_dataset.py
vibhorbajpai: I am not sure what this test is about? Could you please elaborate?

vibhorbajpai:
Making the changes hence closing the request.

alastair:
Re-open original request

webserver/views/api/v1/datasets.py
```diff
@@ -136,18 +137,47 @@ def add_class(dataset_id):
...
-    :<json array public: *Optional.* Array of recording MBIDs (``string``) to add into that class. For example:
+    :<json array recordings: *Optional.* Array of recording MBIDs (``string``) to add into that class. For example:
         ``["770cc467-8dde-4d22-bc4c-a42f91e"]``.
 
 
     :resheader Content-Type: *application/json*
     """
-    raise NotImplementedError
+    ds = get_check_dataset(dataset_id)
+    if ds["author"] != current_user.id:
```
alastair: The `get_check_dataset` method checks read permissions, but we could add another method `get_check_dataset_for_write` which does the same thing but doesn't check if the dataset is public.

webserver/views/api/v1/datasets.py
```diff
@@ -136,18 +137,47 @@ def add_class(dataset_id):
...
+    if "recordings" in class_dict:
+        for mbid in class_dict["recordings"]:
+            try:
+                UUID(mbid, version=4)
+            except ValueError:
+                raise api_exceptions.APIBadRequest("MBID %s not a valid UUID" % (mbid,))
+        unique_mbids = list(set(class_dict["recordings"]))
+        class_dict["recordings"] = unique_mbids
+    try:
+        db.dataset.add_class(class_dict, dataset_id)
```
alastair: You are already checking if this dataset exists, using `get_check_dataset`. I think at this point you can assume that it exists, and that this `add_class` method won't fail to find the dataset.

webserver/views/api/v1/datasets.py
```diff
@@ -136,18 +137,47 @@ def add_class(dataset_id):
...
+    except db.exceptions.NoDataFoundException as e:
+        raise api_exceptions.APINotFound(e.message)
+    return jsonify(
+        success=True,
+        message="Class added."
```
alastair: we should be more explicit here with messages. Include the class name.

webserver/views/api/v1/datasets.py
```diff
@@ -136,18 +137,47 @@ def add_class(dataset_id):
...
-    raise NotImplementedError
+    ds = get_check_dataset(dataset_id)
+    if ds["author"] != current_user.id:
+        raise api_exceptions.APIUnauthorized("You can't create this class.")
+    class_dict = request.get_json()
+    if not class_dict:
+        raise api_exceptions.APIBadRequest("Data must be submitted in JSON format.")
+    if "name" not in class_dict:
+        raise api_exceptions.APIBadRequest("name key missing in JSON request.")
+    if class_dict["name"] in (classes["name"] for classes in ds["classes"]):
```
alastair: I'm not sure about this check. Perhaps @gentlecat can give some feedback too. I don't know if it should be an error if the class already exists. In this case we could:
 - treat it as an error
 - treat it as an update/add_recordings and add the recordings to the existing class

webserver/views/api/v1/datasets.py
```diff
@@ -194,7 +224,22 @@ def delete_class(dataset_id):
...
     :resheader Content-Type: *application/json*
     """
-    raise NotImplementedError
+    ds = get_check_dataset(dataset_id)
+    if ds["author"] != current_user.id:
+        raise api_exceptions.APIUnauthorized("You can't delete this class.")
+    class_dict = request.get_json()
+    if "name" not in class_dict:
+        raise api_exceptions.APIBadRequest("name key missing in JSON request.")
+    if class_dict["name"] not in (classes["name"] for classes in ds["classes"]):
```
alastair: This is similar to the check in the the `add_class` method. @gentlecat 

Here we could raise an error, as you do, or just silently skip the request. There is no HTTP 2xx code which says that something was removed, so we can't use that as an indicator.

webserver/views/api/v1/datasets.py
```diff
@@ -194,7 +224,22 @@ def delete_class(dataset_id):
...
+        raise api_exceptions.APIBadRequest("Class does not exists.")
+    try:
+        db.dataset.delete_class(class_dict, dataset_id)
```
alastair: This method doesn't raise `NoDataFoundException` so the try/except doesn't make sense

webserver/views/api/v1/datasets.py
```diff
@@ -217,7 +262,38 @@ def add_recordings(dataset_id):
...
+        raise api_exceptions.APIBadRequest("recordings key missing in JSON request.")
+    for mbid in class_dict["recordings"]:
+        try:
+            UUID(mbid, version=4)
+        except ValueError:
+            raise api_exceptions.APIBadRequest("MBID %s not a valid UUID" % (mbid, ))
+    for classes in ds["classes"]:
+        if classes["name"] == class_dict["class_name"]:
+            for mbid in class_dict["recordings"]:
+                if mbid in (classes["recordings"]):
```
alastair: You could do this check by turning the two lists of recordings into sets and taking the difference.

webserver/views/api/v1/datasets.py
```diff
@@ -240,7 +316,41 @@ def delete_recordings(dataset_id):
 
     :resheader Content-Type: *application/json*
     """
-    raise NotImplementedError
+    #ds = get_check_dataset(dataset_id)
```
alastair: Make sure you don't commit commented code

webserver/views/api/v1/datasets.py
```diff
@@ -136,18 +137,47 @@ def add_class(dataset_id):
...
     """
-    raise NotImplementedError
+    ds = get_check_dataset(dataset_id)
+    if ds["author"] != current_user.id:
+        raise api_exceptions.APIUnauthorized("You can't create this class.")
+    class_dict = request.get_json()
+    if not class_dict:
+        raise api_exceptions.APIBadRequest("Data must be submitted in JSON format.")
+    if "name" not in class_dict:
+        raise api_exceptions.APIBadRequest("name key missing in JSON request.")
```
alastair: We do validation of datasets in the `utils.dataset_validator` file. We can adapt this file to also do validation of input into these methods. You can use some of the existing helper methods, such as `_check_dict_structure`

db/dataset.py
```diff
@@ -315,3 +315,54 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+            raise exceptions.NoDataFoundException("No such data exists.")
+        cls_id = result.fetchone()[0]
+        if "recordings" in dict:
+            for mbid in dict["recordings"]:
+                connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:class_name, :mbid_num)"""),
+                                   {"class_name": cls_id, "mbid_num": mbid})
+
+
+def delete_class(dict, dataset_id):
+    """Delete a class from a dataset"""
```
alastair: Add another sentence to this documentation saying that class members will also be deleted because we have ON DELETE CASCADE set in the database

webserver/views/api/v1/datasets.py
gentlecat: Alternatively, this can be an argument to `get_check_dataset`.

webserver/views/api/v1/datasets.py
gentlecat: I think it would be easier to use with `UPSERT`-like behaviour. If class exists, then extend its list of recordings with new ones.

This should, obviously, be explicitly documented.

webserver/views/api/v1/datasets.py
```diff
@@ -194,7 +224,22 @@ def delete_class(dataset_id):
 
     :resheader Content-Type: *application/json*
     """
-    raise NotImplementedError
+    ds = get_check_dataset(dataset_id)
+    if ds["author"] != current_user.id:
+        raise api_exceptions.APIUnauthorized("You can't delete this class.")
```
gentlecat: I would just say "Only author can modify this dataset."

gentlecat: Same in other places.

webserver/views/api/v1/datasets.py
```diff
@@ -194,7 +224,22 @@ def delete_class(dataset_id):
...
+    class_dict = request.get_json()
+    if "name" not in class_dict:
+        raise api_exceptions.APIBadRequest("name key missing in JSON request.")
```
gentlecat: "Name of the class to be deleted must be specified." might be a better message.

webserver/views/api/v1/datasets.py
gentlecat: I think not raising an error in that case is a bad idea. 200 is the default response for successful request, `DELETE` too.

webserver/views/api/v1/datasets.py
```diff
@@ -194,7 +224,22 @@ def delete_class(dataset_id):
...
+    if class_dict["name"] not in (classes["name"] for classes in ds["classes"]):
+        raise api_exceptions.APIBadRequest("Class does not exists.")
+    try:
+        db.dataset.delete_class(class_dict, dataset_id)
+    except db.exceptions.NoDataFoundException as e:
+        raise api_exceptions.APINotFound(e.message)
+    return jsonify(
+        success=True,
+        message="Class deleted."
```
gentlecat: I'm not sure if there's actually any point in having these messages. This endpoint handles one task and the result can be determined from status code in the response.

db/dataset.py
```diff
@@ -315,3 +315,54 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
         DELETE FROM dataset_snapshot
               WHERE dataset_id = :dataset_id""")
     connection.execute(query, {"dataset_id": dataset_id})
+
+
+def add_recordings(dict, dataset_id):
```
gentlecat: What's supposed to be in the `dict`? This should be documented.

gentlecat: `add_recordings(dataset_id, class_name, recordings)` would be even better.

webserver/views/api/v1/datasets.py
alastair: Right, we could have 

    def get_check_dataset(dataset_id, write=False)

with the default we don't need to change existing read-only uses of the function, and we can use `write=True` for the API edit endpoints

webserver/views/api/v1/datasets.py
alastair: OK, I agree with this. @vibhorbajpai You can change this to add all recordings which are in the request but not yet in the class, if the class already exists.

webserver/views/api/v1/datasets.py
alastair: Agreed. If there is an error, return a message, but if no error return a blank response? with HTTP 200

db/dataset.py
alastair: This is because it's the input as given to the API endpoint. However, I agree. `(dataset_id, class_name, recordings)` is a good method signature

db/dataset.py
```diff
@@ -315,3 +315,54 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+    """Adds new recordings to a class in a dataset"""
+
+    with db.engine.begin() as connection:
+        result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
```
alastair: A reminder that we talked about formatting of SQL statements and making sure that lines are not too long

db/dataset.py
```diff
@@ -315,3 +315,54 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+
+def add_class(dict, dataset_id):
+    """Add a class to a dataset"""
+
+    with db.engine.begin() as connection:
+        if "description" not in dict:
+            dict["description"] = None
+        result = connection.execute(sqlalchemy.text("""INSERT INTO dataset_class (name, description, dataset) VALUES (:name, :description, :datasetid) RETURNING id"""),
+                                    {"name": dict["name"], "description": dict["description"], "datasetid": dataset_id})
+        if result.rowcount < 1:
```
alastair: Because we don't have INSERT ... ON CONFLICT, this command probably won't return less than 1 row. We dont need this check

db/dataset.py
```diff
@@ -315,3 +315,54 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+            raise exceptions.NoDataFoundException("No such data exists.")
+        cls_id = result.fetchone()[0]
+        if "recordings" in dict:
```
alastair: This is the same behaviour as the `add_recrodings` method and should be replaced with a call to that method

webserver/views/api/v1/datasets.py
alastair: OK, we can leave the error here in this case then (user tries to delete a class which doesn't exist in the dataset)

db/test/test_dataset.py
```diff
@@ -169,6 +169,61 @@ def test_delete_shapshot(self):
...
+        id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        test_data_recordings = {
+            "class_name": "Class #1",
+            "recordings": ["1c085555-3805-428a-982f-e14e0a2b18e6"]
+        }
+        originaldataset = dataset.get(id)
+        dataset.add_recordings(test_data_recordings, id)
+        updateddataset = dataset.get(id)
+        self.assertDictEqual(originaldataset["classes"][1], updateddataset["classes"][1])
+        self.assertEqual(len(originaldataset["classes"][0]["recordings"])+1, len(updateddataset["classes"][0]["recordings"]))
```
alastair: I said in the previous review that we should check the contents of the dataset, not the length. If you check the contents, you don't need to also check the length. 

For this test, it would be ok to be explicit about the recordings that you expect:

```python
expected_recordings = ["1c085555-3805-428a-982f-e14e0a2b18e6", "0dad432b-16cc-4bf0-8961-fd31d124b01b", "19e698e7-71df-48a9-930e-d4b1a2026c82"]
self.assertEqual(expected_recordings, updateddataset[0]["recordings"])
```

Because you know exactly what should be in the dataset, it's OK to list the recording ids in the test code.

db/test/test_dataset.py
```diff
@@ -169,6 +169,61 @@ def test_delete_shapshot(self):
...
+    def test_delete_recordings(self):
+        id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        test_data_recordings = {
+            "class_name": "Class #1",
+            "recordings": ["ed94c67d-bea8-4741-a3a6-593f20a22eb6"]
+        }
+        originaldataset = dataset.get(id)
+        dataset.add_recordings(test_data_recordings, id)
+        dataset.delete_recordings(test_data_recordings, id)
+        updateddataset = dataset.get(id)
```
alastair: Here, you call `add_recordings` and then `delete_recordings` directly one after another without checking the result. This test passes, but it would also pass if the add and delete methods didn't do anything.
A better test here would be to delete a single recording which already exists in `self.test_data` and then do a similar test to what I proposed in `test_add_recordings` to check that the list of recordings is smaller than the original dataset list.

db/test/test_dataset.py
```diff
@@ -169,6 +169,61 @@ def test_delete_shapshot(self):
...
+            "description": "This is description of class 3",
+            "recordings": ["1c085555-3805-428a-982f-e14e0a2b18e6"]
+
+        }
+        originaldataset = dataset.get(id)
+        dataset.add_class(test_data, id)
+        updateddataset = dataset.get(id)
+        self.assertDictEqual(originaldataset["classes"][0],updateddataset["classes"][0])
+        self.assertDictEqual(originaldataset["classes"][1], updateddataset["classes"][1])
+        self.assertEqual(len(originaldataset["classes"])+1, len(updateddataset["classes"]))
```
alastair: You can use a number here: 
```python
self.assertEqual(3, len(updateddataset["classes"]))
```

db/test/test_dataset.py
```diff
@@ -169,6 +169,61 @@ def test_delete_shapshot(self):
...
+        self.assertEqual(updateddataset["classes"][2]["name"],"Class #3")
```
alastair: You should also check the recordings in class3.  Check that you have a space after `,` in all places in this test

db/dataset.py
```diff
@@ -315,3 +315,54 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+        clsid = result.fetchone()
+        for mbid in dict["recordings"]:
+            connection.execute(sqlalchemy.text("""INSERT INTO dataset_class_member (class, mbid) VALUES (:clsid, :mbid)"""), {"clsid": clsid[0], "mbid": mbid})
+
+
+def delete_recordings(dict, dataset_id):
+    """Delete all recordings from the list of IDs"""
+
+    with db.engine.begin() as connection:
+        result = connection.execute(sqlalchemy.text("""SELECT id FROM dataset_class WHERE name = :name and dataset = :dataset_id"""),
```
alastair: You use this query in all of your methods, to get the class. You can move this into a method
```python
def _get_class_in_dataset(connection, dataset_id, class_name)
```
so that you don't duplicate code.  You can then test this method separately.
The test should check four things:
 - if the dataset doesn't exist, raises an error
 - if the dataset exists but the class doesn't, raise an error
 - if the dataset and class exist, return it
 - if there are two datasets and they both have a class called "Class #1", calling this method will get the class from the correct dataset

db/test/test_dataset.py
```diff
@@ -169,6 +169,61 @@ def test_delete_shapshot(self):
...
+    def test_delete_class(self):
+        id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        test_data = {
+            "name": "Class #2"
+        }
+        originaldataset = dataset.get(id)
+        dataset.delete_class(test_data, id)
+        updateddataset = dataset.get(id)
+        self.assertDictEqual(originaldataset["classes"][0],updateddataset["classes"][0])
+        self.assertEqual(len(originaldataset["classes"])-1,len(updateddataset["classes"]))
```
alastair: You can use a number here too instead of `len()-1`

webserver/views/api/v1/test/test_datasets.py
```diff
@@ -79,3 +79,648 @@ def test_create_dataset(self, create_from_dict):
         self.assertEqual(resp.status_code, 200)
         expected = {"success": True, "dataset_id": "6b6b9205-f9c8-4674-92f5-2ae17bcb3cb0"}
         self.assertEqual(resp.json, expected)
+
+
+    def test_add_recordings(self):
```
alastair: With tests in the webserver, we don't need to create data in the database and check that it has been updated. Instead, we can use mocks, which verify that we call a method with the correct arguments, but doesn't actually call the function. See some examples here: https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/views/api/v1/test/test_datasets.py#L48-L63

With mocks we can throw exceptions (using `.side_effect`) to make sure that the API returns these errors correctly, and we can also return data (`.return_value`) if we want to pass some data back to the API.

You can read more documentation on mock here: https://docs.python.org/dev/library/unittest.mock.html (don't worry that it says for python 3.3, we use a version of the same library for python 2)

webserver/views/api/v1/test/test_datasets.py
```diff
@@ -79,3 +79,648 @@ def test_create_dataset(self, create_from_dict):
...
+        id = db.dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        submit = json.dumps({
+            "class_name": "Class #1",
+            "recordings": ["1c085555-3805-428a-982f-e14e0a2b18e6",]
+        })
+        url = '/api/v1/datasets/%s/recordings' % (str(id))
+        resp = self.client.put(url, data=submit, content_type='application/json')
+        self.assertEqual(resp.status_code, 200)
+        expected = {"success": True, "message": "Recording(s) added."}
+        self.assertEqual(resp.json, expected)
```
alastair: Here are some examples of tests which I would add for `test_add_recordings`:

```python
@mock.patch("db.dataset.create_from_dict")
@mock.patch("utils.dataset_validator.validate_add_recordings")
def test_success_add_recordings(self, validate, create):
    submit = {
        "class_name": "Class #1",
        "recordings": ["1c085555-3805-428a-982f-e14e0a2b18e6",]
    }
    url = '/api/v1/datasets/%s/recordings' % (str(id))
    resp = self.client.put(url, json=submit, content_type='application/json')

    validate.assert_called_once_with(submit)
    create.assert_called_once_with(id, "Class #1", ["1c085555-3805-428a-982f-e14e0a2b18e6",])
    self.assertEqual(200, resp.code)

@mock.patch("db.dataset.create_from_dict")
@mock.patch("utils.dataset_validator.validate_add_recordings")
def test_success_add_recordings(self, validate, create):
    submit = {
        "class_name": "Class #1",
        "recordings": ["1c085555-3805-428a-982f-e14exxxx",]
    }
    url = '/api/v1/datasets/%s/recordings' % (str(id))

    validate.side_effect = ValidationError("1c085555-3805-428a-982f-e14exxxx is not a valid MBID")

    resp = self.client.put(url, json=submit, content_type='application/json')
    self.assertEqual(400, resp.code)
    self.assertEqual(resp.json, expected)
    validate.assert_called_once_with(submit)
    create.assert_not_called()
```

webserver/views/api/v1/datasets.py
gentlecat: Some APIs return data in similar cases, some don't. Examples:

* https://developer.spotify.com/web-api/unfollow-artists-users/
* https://stripe.com/docs/api#delete_coupon

And if we aren't returning anything it should probably be `204` instead. See http://stackoverflow.com/a/2342589/272770 and references there.

alastair:
Just two quick things before I continue with the review:

Make sure that your identity on git is configured correctly. You have not configured your email address correctly: https://help.github.com/articles/setting-your-email-in-git/

Try and make each commit represent a particular change. Instead of having 2 big commits which contain many changes, it's OK to have 10 or 15 commits with each of them only changing a single feature. The commit message should *always* describe the change that you are making, instead of a generic "work in progress" message, so that people looking at the history or reviewing the code can see at a glance what changes a commit makes. In the case of this PR, you could have one commit addressing each comment from my review. If there are a set of comments that are about similar things, you could join them in one commit.

alastair:
There are currently failing tests. When you make your changes, make sure that the tests still pass, updating them if necessary. This ties into the second comment I made above. A commit should be a single change if possible, which includes the change and also any required updates to tests.

I can't review this until the tests all pass, because I don't know how much is done and how much you are still working on.

alastair:
This is good. I've made some changes and to what you've done and merged it into #219. I'm closing this in favour of that PR.

PR #215, opened 2016-11-10
[WIP] Bigquery

alastair:
First attempt at a BigQuery streamer.

Still some issues with:
 - mfcc fields - lists of lists, we need to work out the best way to store this (perhaps not as records if we don't want to search it). There is much more of this kind of data in the frame documents.
 - metadata: do we submit rowid? mbid and offeset? submission date?
 - do we duplicate track metadata in ll and hl tables, or have a separate metadata table?

### DIFF
--- a/admin/bigquery/highlevel-schema.json
+++ b/admin/bigquery/highlevel-schema.json
@@ -1,0 +1,704 @@
+[
+  {
+    "name": "meta",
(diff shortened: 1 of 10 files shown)


alastair:
Not going to be finished for now, better to close this out of date work

PR #216, opened 2016-11-10
Inject pyopenssl for certificate validation

alastair:
Used in oauth login with musicbrainz, broken since
newhost setup. Also reported in AB-130 but not correctly
fixed the first time around

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,5 +20,6 @@ raven[flask] == 5.9.2
 setproctitle == 1.1.9
 six == 1.9.0
 SQLAlchemy == 1.0.9
+urllib3[secure]==1.19
 Werkzeug == 0.10.4
 -r docs/requirements.txt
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -20,5 +20,6 @@ raven[flask] == 5.9.2
 setproctitle == 1.1.9
 six == 1.9.0
 SQLAlchemy == 1.0.9
+urllib3[secure]==1.19
```
gentlecat: I also had these in CB before (not sure if needed):

```
ndg-httpsclient == 0.3.2
pyOpenSSL == 0.13
```

alastair: my understanding is that [secure] pulls these in

http://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification-in-python-2 "if you want to install the packages manually..."

alastair:
It seems a better solution is to just install `ndg-httpsclient==0.4.2`, this pulls in pyopenssl as a dependency and automatically injects itself into the right place so that the oauth client in AB is happy.

It seems that the above urllib3 thing is for `requests`, which we're not using.

PR #218, opened 2016-11-23
Dataset api part2 vibhor add and delete recordings

vibhorbajpai:
Hi,
Please find the relevant files for the add and delete recordings functionality. 
Thanks and regards,
Vibhor Bajpai


alastair:
We should be using the original request

PR #219, opened 2017-02-06
Dataset api clean

alastair:
Continuation of https://github.com/metabrainz/acousticbrainz-server/pull/213

Finish implementation of dataset create/edit/delete/update API. Change web tests to use mocks so that we don't call db methods.


db/dataset.py
```diff
@@ -36,27 +36,65 @@ def create_from_dict(dictionary, author_id):
...
+    updates = []
+    if "name" in meta:
+        updates.append("name = :name")
+        params["name"] = meta["name"]
+    if "description" in meta:
+        updates.append("description = :description")
+        params["description"] = meta["description"]
+    if "public" in meta:
+        updates.append("public = :public")
+        params["public"] = meta["public"]
```
gentlecat: This can be improved a bit by doing something like this:
```python
param = meta.pop("param", None)
if param is not None:
     updates.append("param = :param")
     params["param"] = meta["param"]
```
After that we can check if all keys in `meta` were valid:
```python
if meta:
    raise ValueError('Unexpected meta value: %r' % meta)
```

db/dataset.py
```diff
@@ -36,27 +36,65 @@ def create_from_dict(dictionary, author_id):
...
+    setstr = ", ".join(updates)
+    query = text("""UPDATE dataset
+                               SET %s
+                             WHERE id = :id
+
```
gentlecat: Query formatting looks a bit off.

db/dataset.py
```diff
@@ -315,3 +355,147 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+def add_recordings(dataset_id, class_name, recordings):
+    """Adds new recordings to a class in a dataset.
+
+    If any given recording already exist in this class it is not added again.
+
+    Args:
+        dataset_id: the uuid of the dataset
+        class_name: The class to add the recordings to
+        recordings: the recordings to add
```
gentlecat: "List of recordings (MBID strings) to add" might be a slightly better description.

db/dataset.py
```diff
@@ -315,3 +355,147 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+        for mbid in recordings:
+            connection.execute(sqlalchemy.text("""
+                DELETE FROM dataset_class_member
+                      WHERE class = :class_name
+                        AND mbid = :mbid_num
+                            """),
+                               {"class_name": clsid, "mbid_num": mbid})
+
+
+def add_class(dataset_id, dict):
```
gentlecat: I'd rather not use keywords as arguments or variable names. Might be better to rename `dict` into `class_data` or something like that.

db/dataset.py
```diff
@@ -315,3 +355,147 @@ def _delete_snapshots_for_dataset(connection, dataset_id):
...
+                           FROM dataset_class d
+                          WHERE d.name = :name
+                            AND d.dataset = :datasetid)
+                    """),
+                           {"name": dict["name"], "description": dict["description"], "datasetid": dataset_id})
+    if "recordings" in dict:
+        add_recordings(dataset_id, dict["name"], dict["recordings"])
+
+
+def delete_class(dataset_id, dict):
```
gentlecat: Again, better not use `dict`. In this case it can just be replaced with a `class_name`.

utils/dataset_validator.py
```diff
@@ -48,72 +48,190 @@ def validate(dataset):
...
 
-def _validate_class(cls, idx):
+def validate_class(cls, idx=None, recordings_required=True):
+    """ Validate the contents of a class dictionary
+
+    :param cls: the class to check
+    :param idx: the index of this class if it is part of a list of classes,
+                otherwise None if it's a standalone class
+    :param recordings_required: True if this class must have a `recordings` element
+    :raises: `ValidationException` if the class definition has a problem
```
gentlecat: We should use the same docstring formatting *(argument and return values documentation in this case)* throughout the project.

db/test/test_dataset.py
```diff
@@ -115,6 +115,14 @@ def test_update_metadata(self):
         self.assertNotEqual(existing_dataset["public"], new_dataset["public"])
         self.assertEqual(new_dataset["public"], False)
 
+    def test_update_dataset_meta_badmeta(self):
+        dsid = str(uuid.uuid4())
+	try:
+            dataset.update_dataset_meta(dsid, {"name": "new name", "badkey": "badvalue"})
+	    self.fail("Shouldn't get here")
+	except ValueError as e:
```
gentlecat: Tabs.

PR #220, opened 2017-02-07
Add crossdomain header for /api/v1/low-level endpoint (AB-295)

loujine:
Theoretically it should work... I haven't tested it yet though

Edit: tested, the modified test passes only with the change in core.py


alastair:
Thanks. Merged and released

PR #221, opened 2017-02-08
API endpoint for getting data count for multiple recordings (AB-294)

loujine:
First working draft

I see as possible improvements:
- refactoring some common code between get_many_lowlevel and get_many_count (core.py)
- add a NoDataFoundException in db/data.py
- more tests :) (with at least a count 2 and a count 0)


loujine:
Also, the JSON response format I chose {"mbid1": count1, "mbid2": count2} is not really consistent with the format of the MBID/count endpoint {"count": count, "mbid": mbid}

loujine:
Added tests
Waiting for feedback for the other points

db/data.py
```diff
@@ -423,6 +423,20 @@ def count_lowlevel(mbid):
         )
         return result.fetchone()[0]
 
+
+def count_many_lowlevel(mbids):
+    """Count number of stored low-level submissions for a specified set
+    of MBID."""
+    with db.engine.connect() as connection:
+        query = text("""
+            SELECT gid, COUNT(*)
```
alastair: Watch the formatting here - we right align SQL keywords

db/testing.py
```diff
@@ -77,3 +78,22 @@ def load_low_level_data(self, mbid):
         """
         with open(self.data_filename(mbid)) as json_file:
             db.data.submit_low_level_data(mbid, json.loads(json_file.read()), gid_types.GID_TYPE_MBID)
+
+    def load_fake_low_level_data(self, mbid):
```
alastair: This is a handy method. good idea.
We should probably make the method `submit_fake_low_level_data`, instead of `load_`
Can you add a small docstring explaining what it does?

I don't think it's important to have average loudness a random variable. Why did you make it like this?

webserver/views/api/v1/core.py
```diff
@@ -197,3 +197,39 @@ def get_many_lowlevel():
...
+        raise webserver.views.api.exceptions.APIBadRequest(
+            "Missing `recording_ids` parameter")
+
+    recordings = _parse_bulk_params(recording_ids)
+    if len(recordings) > 200:
+        raise webserver.views.api.exceptions.APIBadRequest(
+            "More than 200 recordings not allowed per request")
+
+    mbids = set(r[0] for r in recordings)
+    recording_count = {}.fromkeys(mbids, 0)
```
alastair: This is neat, I didn't know of this method before. Is the idea here to make sure that all MBIDs have a value in the returned dictionary even if it's 0?

This ties in to https://github.com/metabrainz/acousticbrainz-server/pull/223 a little bit too - we have to decide what to return for missing data.
I would be just as happy to return no data for mbids which are not in the database. In this case it's up to the caller to check if the keys exist.

webserver/views/api/v1/core.py
```diff
@@ -197,3 +197,39 @@ def get_many_lowlevel():
...
+    recording_count.update({str(mbid): int(count) for (mbid, count)
```
alastair: Can you make the db method return this dictionary, instead of formatting it in the view?

alastair:
Regarding format, what do you think about
```json
{"mbid": {"count": 1}, {"mbid": {"count":2}}
```

too complex?

db/data.py
loujine: fixed

db/testing.py
loujine: I changed the name and added a docstring

The random() allows me to submit several times the same mbid as a new separate entry, to test counts > 1. I tried to make that clear in the docstring, I'm not sure whether I managed.

webserver/views/api/v1/core.py
loujine: fixed

webserver/views/api/v1/core.py
loujine: Yes, all mbids returned by _parse_bulk_params (so I guess all valid musicbrainz mbids, whether they have low-level data associated or not) are initialized with count=0.
I guess invalid mbids (i.e. not existing in muscbrainz) are not returned by _parse_bulk_params? I didn"t check though

loujine:
I changed the response format, this looks fine for me (javascript object = easy to parse keys when receiving the response).

I thought you might want to have the same format as the MBID/count endpoint:
[ {"count": count1, "mbid": mbid1}, {"count": count2, "mbid": mbid2}, ... ]
but that make the response more annoying to parse for specific mbids on the client side

db/data.py
```diff
@@ -418,11 +418,29 @@ def count_lowlevel(mbid):
...
+    with db.engine.connect() as connection:
+        query = text(
+            """SELECT gid
+                    , COUNT(*)
+                 FROM lowlevel
+                WHERE gid
+                   IN :mbids
+             GROUP BY gid;""")
+        return {str(mbid): int(count) for mbid, count
+                in connection.execute(query, mbids=tuple(mbids))}
```
alastair: I missed this in the first review - we use a dict as the second argument to execute: `{"mbids": tuple(mbids)}`. IN fact, I didn't know that this other syntax was valid, but we should stay consistent.

db/data.py
```diff
@@ -418,11 +418,29 @@ def count_lowlevel(mbid):
...
+def count_many_lowlevel(mbids):
+    """Count number of stored low-level submissions for a specified set
+    of MBID."""
+    with db.engine.connect() as connection:
+        query = text(
+            """SELECT gid
+                    , COUNT(*)
+                 FROM lowlevel
+                WHERE gid
+                   IN :mbids
```
alastair: for consistency, we write this as `WHERE gid IN :mbids`, all on a single line

db/testing.py
alastair: Ah, I understand now. It would be more explicit to have a separate parameter to this function which you could manually change each time you want a different submission.

This might mean you need a bit more code when adding these items to the database (you won't be able to use a for loop unless you add some extra data), but I think it would make it clearer what the method does.

webserver/views/api/v1/core.py
```diff
@@ -197,3 +199,39 @@ def get_many_lowlevel():
...
+            "Missing `recording_ids` parameter")
+
+    recordings = _parse_bulk_params(recording_ids)
+    if len(recordings) > 200:
+        raise webserver.views.api.exceptions.APIBadRequest(
+            "More than 200 recordings not allowed per request")
+
+    mbids = set(r[0] for r in recordings)
+    recording_count = {}.fromkeys(mbids, {'count': 0})
+    for (mbid, count) in six.iteritems(db.data.count_many_lowlevel(mbids)):
```
alastair: We decided in #223 to not list an mbid in the response if it doesn't exist in the database. I think we should do the same here, this means we can just return `jsonify(db.data.count_many_lowlevel(mbids))`.
The documentation should say this - that if a mbid doesn't exist in the database then it is omitted from the results.

alastair:
I like the format better now. It fits in with the format of the bulk get method more. You're right, returning a list would be more annoying to consume by the client.

db/data.py
loujine: Of course. I introduced this mistake when rereading my code just before pushing... it was fine beforehand :)

db/data.py
loujine: fixed

webserver/views/api/v1/core.py
loujine: fixed

alastair:
:1st_place_medal:

PR #222, opened 2017-02-17
AB-138 Allow pasting URLs for recording MBIDs

samj1912:


jesus2099:
Hi @samj1912.
We should better ask the reviewer’s opinion than mine but I think we could get rid of the trailing `$` as we could copy the any recording URL including one of its tabs (Overview, Fingerprints, Aliases, Tags, Details, Edit, etc.).

Maybe a regular expression that could be tested would be something like:

    /\b[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\b/

Just a first thought…

alastair:
With this change I can no longer create or edit a dataset regardless of if I paste a URL or a mbid. The error in the server is
```
  File "/home/alastair/code/acousticbrainz-server/webserver/views/datasets.py", line 217, in create
    dataset_id = db.dataset.create_from_dict(dataset_dict, current_user.id)
  File "/home/alastair/code/acousticbrainz-server/db/dataset.py", line 31, in create_from_dict
    dataset_validator.validate(dictionary)
  File "/home/alastair/code/acousticbrainz-server/utils/dataset_validator.py", line 63, in validate
    _validate_classes(dataset["classes"])
  File "/home/alastair/code/acousticbrainz-server/utils/dataset_validator.py", line 74, in _validate_classes
    _validate_class(cls, idx)
  File "/home/alastair/code/acousticbrainz-server/utils/dataset_validator.py", line 106, in _validate_class
    _validate_recordings(cls["recordings"], cls["name"], idx)
  File "/home/alastair/code/acousticbrainz-server/utils/dataset_validator.py", line 114, in _validate_recordings
    if not UUID_RE.match(recording):
TypeError: expected string or buffer
```

samj1912:
Fixed :)

alastair:
This looks better now. Is it possible to share the uuid regex over the two places that it's used? It seems repetitive to have it twice.

My preference would be for the URL regex to specifically accept musicbrainz recording ids. Currently it will accept (anything)(uuid). This can have any text, but also artist, release, urls from musicbrainz too.
How about we accept `(https?://(beta.)?musicbrainz.org/recording/)?<mbid>`

jesus2099:
@alastair,
Regarding [I thought](https://github.com/metabrainz/acousticbrainz-server/pull/222#issuecomment-281603020) it was a pity to put a `$` after the MBID and [your suggestion](https://github.com/metabrainz/acousticbrainz-server/pull/222#issuecomment-288060288) of putting `(https?://(beta.)?musicbrainz.org/recording/)?<mbid>` is great although not followed.
Could you still set this great code after accepting PR?

alastair:
@jesus2099 I added `beta.` in another commit. I'm not a huge fan of allowing any recording URLs, because technically we are asking for an MBID (in its uuid or url form), of which `/overview`, etc isn't actually correct.
Can you explain a specific workflow that you might want to do that requires you to go to a detail page of a recording before deciding to add it?

I would prefer to spend our development time on improving the way that this page works, instead of small optimisations. For example, I'd like to see add-by-artist, add-by-release, add-by-tag search boxes, which should make this process much faster.

I read your Issue again and see that you suggested to submit the mbid as soon as it's valid. This is a pretty neat idea, I'll see if I can add that too.

jesus2099:
I only thought it would be safe to allow any URL by replacing the trailing `$` by `\b`.
The reason is just why not? and because
I might be on any sub-page when hitting F8, Ctrl+c to use this for pasting. :)
Anyway I am not that obsessed by it, I was just testing the site, back then, when some feedback was asked.

webserver/static/scripts/datasets/editor.js
```diff
@@ -460,39 +462,56 @@ var RecordingAddForm = React.createClass({
...
         this.props.onRecordingSubmit(mbid);
         this.refs.mbid.value = '';
     },
     handleChange: function () {
-        var mbid = this.refs.mbid.value;
-        var isValidUUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(mbid);
-        var isNotDuplicate = this.props.recordings.indexOf(mbid) == -1;
+        var mbidField = this.refs.mbid.value;
+        var isValidUUID = RECORDING_MBID_RE.test(mbidField);
+        var isNotDuplicate = false;
```
paramsingh: `isNotDuplicate` seems like a weird bool variable to read, maybe replacing it with `isDuplicate` would be a good idea.

Would also make reading this condition in the code easier:

`else if (this.state.length > 0 && !this.state.notDuplicate) {` 

vs

`else if (this.state.length > 0 && this.state.isDuplicate) {`

alastair: the biggest reason is because it was like that to begin with :) - it made sense when the only condition it was being used in was `validInput: isValidUUID && isNotDuplicate`, but you're right. Changing it to isDuplicate should make it easier to reason about in all of the places it's used

paramsingh: I fixed it (had the time) :P

alastair:
@jesus2099 you're right - there's no real reason to not include this functionality, especially since it's only a 1 character fix.

I've also updated it so that there's a checkbox which when selected will automatically add a valid input to the recording list as soon as it is typed or pasted

PR #223, opened 2017-02-20
Fix bulk low-level get request with non-existent MBID

saifulbkhan:
GET request for bulk low-level data was failing when any single
one of the MBIDs were not present in the database. These entries
are now put into an error field appended to the fetched data for
existing recordings.

Relevant Jira ticket: https://tickets.metabrainz.org/browse/AB-296

### DIFF
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -193,7 +193,12 @@ def get_many_lowlevel():
 
     # TODO: This should use a bulk get method for speed
     for recording_id, offset in recordings:
-        recording_details.setdefault(recording_id, {})[offset] = db.data.load_low_level(recording_id, offset)
+        try:
+            recording_details.setdefault(recording_id, {})[offset] = db.data.load_low_level(recording_id, offset)
+        except NoDataFoundException:
+            errors = recording_details.setdefault("errors", {})
+            mbid_entry = errors.setdefault(recording_id, {})
+            mbid_entry[offset] = {"message": "recording does not exist"}
 
     return jsonify(recording_details)
 
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -196,9 +196,7 @@ def get_many_lowlevel():
         try:
             recording_details.setdefault(recording_id, {})[offset] = db.data.load_low_level(recording_id, offset)
         except NoDataFoundException:
-            errors = recording_details.setdefault("errors", {})
-            mbid_entry = errors.setdefault(recording_id, {})
-            mbid_entry[offset] = {"message": "recording does not exist"}
+            pass
 
     return jsonify(recording_details)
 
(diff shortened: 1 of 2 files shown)


webserver/views/api/v1/core.py
```diff
@@ -193,7 +193,12 @@ def get_many_lowlevel():
 
     # TODO: This should use a bulk get method for speed
     for recording_id, offset in recordings:
-        recording_details.setdefault(recording_id, {})[offset] = db.data.load_low_level(recording_id, offset)
+        try:
+            recording_details.setdefault(recording_id, {})[offset] = db.data.load_low_level(recording_id, offset)
+        except NoDataFoundException:
+            errors = recording_details.setdefault("errors", {})
```
alastair: I've been thinking about this `errors` key. I don't think it's necessary. My opinion now is that if a recording doesn't exist, we just skip it.

webserver/views/api/v1/test/test_core.py
```diff
@@ -133,6 +133,31 @@ def test_get_bulk_ll(self, load_low_level):
...
+        expected_result = {
+            "c5f4909e-1d7b-4f15-a6f6-1af376bc01c9": {"0": rec_c5},
+            "405a5ff4-7ee2-436b-95c1-90ce8a83b359": {"2": rec_40_2},
+        }
+        self.assertEqual(resp.json, expected_result)
+
+        calls = [mock.call("c5f4909e-1d7b-4f15-a6f6-1af376bc01c9", 0),
+                 mock.call("7f27d7a9-27f0-4663-9d20-2c9c40200e6d", 3),
+                 mock.call("405a5ff4-7ee2-436b-95c1-90ce8a83b359", 2)]
+        load_low_level.assert_has_calls(calls)
```
alastair: these are good mocks!

alastair:
This is great, thanks. One thing I missed in the first review is that we need to add to the documentation to say that if data for a recording id doesn't exist, it is silently ignored and will not appear in the returned data.

webserver/views/api/v1/core.py
```diff
@@ -196,9 +196,7 @@ def get_many_lowlevel():
         try:
             recording_details.setdefault(recording_id, {})[offset] = db.data.load_low_level(recording_id, offset)
         except NoDataFoundException:
-            errors = recording_details.setdefault("errors", {})
-            mbid_entry = errors.setdefault(recording_id, {})
-            mbid_entry[offset] = {"message": "recording does not exist"}
+            pass
```
alastair: A small thing on this commit - because you're doing two distinct things, I would have made it as two commits, to make it easier to follow what changes are being made.

saifulbkhan: RIght! I'll seperate the test from this change.

alastair: You didn't have to make the change again - I was just mentioning it for future work. Thanks for going back and doing it 👍

alastair:
:tada:

PR #224, opened 2017-03-01
Add detailed contributing documentation

alastair:
An attempt at starting detailed contributing documentation so that we don't have to keep repeating ourselves. Based a little bit on discussion in IRC, on the Forums, and things that I've noticed over the last few years.

My preference is to get this merged as quickly as possible, and iterate on changes as issues come up; rather than spending a long time discussing updates in this PR.
I'm open to corrections of things which are obviously wrong and updates of things that we currently do.

We should also consider if this is a good place to have these guidelines. CB has a nice set of documentation online (http://critiquebrainz.readthedocs.io/contributing.html). We could also replicate this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,136 @@
-## Contributing
+# AcousticBrainz contribution guidelines
 
-### How to contribute code
-
-1. Fork this repository.
-2. Create a new branch.
-3. Make some changes and [write good commit messages](http://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message).
-4. Create [a new pull request](https://github.com/metabrainz/acousticbrainz-server/compare).
-5. :shipit:
-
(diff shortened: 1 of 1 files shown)


gentlecat:
It would be nice if you could move some of the common stuff to https://github.com/metabrainz/guidelines and put a link to that page. And then expand on AcousticBrainz-specific suggestions/requirements.

alastair:
:+1: I didn't know about this repo

alastair:
pointed to metabrainz guidelines!

CONTRIBUTING.md
```diff
@@ -1,18 +1,61 @@
-## Contributing
+# AcousticBrainz contributing guidelines
```
gentlecat: Would `contribution guidelines` be more correct? Otherwise to me it sounds like guidelines are contributing to something. :)

PR #225, opened 2017-03-02
AB-92: Switch to brainzutils-python for caching purposes

saifulbkhan:
Give up memcache in favour of brainzutils-python which now uses redis.
This also helps with the issue of cache namespace key being stored
in the cache itself and prone to removal by any LRU algorithm.

Ticket: https://tickets.metabrainz.org/browse/AB-92


db/stats.py
```diff
@@ -16,9 +16,9 @@
 STATS_CACHE_TIMEOUT = 60 * 10  # 10 minutes
 LAST_MBIDS_CACHE_TIMEOUT = 60  # 1 minute (this query is cheap)
 
-STATS_MEMCACHE_KEY = "recent-stats"
-STATS_MEMCACHE_LAST_UPDATE_KEY = "recent-stats-last-updated"
-STATS_MEMCACHE_NAMESPACE = "statistics"
+STATS_REDIS_KEY = "recent-stats"
+STATS_REDIS_LAST_UPDATE_KEY = "recent-stats-last-updated"
+STATS_REDIS_NAMESPACE = "statistics"
```
gentlecat: How about `SOMETHING_CACHE_KEY` instead?

saifulbkhan: So then something like:

`STATS_CACHE_KEY`
`STATS_CACHE_LAST_UPDATE_KEY`
`STATS_CACHE_NAMESPACE`

^this?

fabfile.py
```diff
@@ -36,17 +36,17 @@ def build_static():
     local("./node_modules/.bin/gulp")
 
 
-def clear_memcached():
+def clear_redis_cache():
     try:
-        cache.init(config.MEMCACHED_SERVERS)
+        cache.init(config.REDIS_HOST, config.REDIS_PORT, config.REDIS_NAMESPACE)
         cache.flush_all()
```
alastair: When I run this I get an error
```
redis.exceptions.ResponseError: invalid DB index
```

db/stats.py
gentlecat: Right.

fabfile.py
gentlecat: Might be a bug in the lib.

fabfile.py
```diff
@@ -36,17 +36,17 @@ def build_static():
     local("./node_modules/.bin/gulp")
 
 
-def clear_memcached():
+def clear_redis_cache():
     try:
-        cache.init(config.MEMCACHED_SERVERS)
+        cache.init(config.REDIS_HOST, config.REDIS_PORT, config.REDIS_NAMESPACE)
```
saifulbkhan: This is hard to argue with. I'll add a `ns_versions_loc` parameter as well. What do you think should be the directory path/name for namespace?

fabfile.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import with_statement
 from fabric.api import local
 from fabric.colors import green, yellow, red
-from db import cache
+#from brainzutils import cache
```
saifulbkhan: Sorry about that. Will correct it.

admin/install_web_server.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 curl -sL https://deb.nodesource.com/setup_5.x | sudo -E bash -
-apt-get -y install memcached python-virtualenv python-dev ipython pxz nodejs libffi-dev
+apt-get -y install python-virtualenv python-dev ipython pxz nodejs libffi-dev redis-server
```
saifulbkhan: Right!

fabfile.py
saifulbkhan: I am not running a Vagrant VM, so maybe I missed this. I'll try to look into it if I can.

saifulbkhan: No bug in the lib. Just a silly error on my part. Cache should be initialized as:

`cache.init(host='127.0.0.1', port=6379, namespace='AB')`
or
`cache.init('127.0.0.1', 6379, 0, 'AB')`

to avoid setting the 'AB' string for `db_number` argument in the init function. I should change this for other cache inits as well.

fabfile.py
alastair: Good question. For now just set it up in the config file as `os.path.join(os.path.dirname(os.path.realpath(__file__)), 'cache_namespaces')`. In the near future with docker this will have to go in a mounted volume somewhere.

paramsingh:
Is there anything left in this PR? If yes, I'd be happy to finish it.

alastair:
@paramsingh thanks for the offer, I'm merging this along with #236, I don't think it needs anything else

alastair:
I rebased this against my CustomFlask changes from #236. @paramsingh can you have a glance at it? I've not actually tested it yet, will do this tomorrow

admin/install_web_server.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
-apt-get -y install memcached python-virtualenv python-dev ipython pxz nodejs libffi-dev libssl-dev git
+curl -sL https://deb.nodesource.com/setup_5.x | sudo -E bash -
```
alastair: badly merged, this should be 6.x

webserver/__init__.py
```diff
@@ -35,12 +35,22 @@ def create_app(debug=None, config_path=None):
...
-        from db import cache
-        cache.init(app.config['MEMCACHED_SERVERS'],
-                   app.config['MEMCACHED_NAMESPACE'],
-                   debug=1 if app.debug else 0)
+    # Cache 
+    if 'REDIS_HOST' in app.config and\
+       'REDIS_PORT' in app.config and\
+       'REDIS_NAMESPACE' in app.config and\
+       'REDIS_NS_VERSIONS_LOCATION' in app.config:
+        if not os.path.exists(app.config['REDIS_NS_VERSIONS_LOCATION']):
```
alastair: should be try/except instead of exists. We're still on python 2 so can't use the `exist_ok` flag to makedirs

webserver/__init__.py
```diff
@@ -35,12 +35,22 @@ def create_app(debug=None, config_path=None):
...
+            os.makedirs(app.config['REDIS_NS_VERSIONS_LOCATION'])
+
+        from brainzutils import cache
+        cache.init(
+            host=app.config['REDIS_HOST'],
+            port=app.config['REDIS_PORT'],
+            namespace=app.config['REDIS_NAMESPACE'],
+            ns_versions_loc=app.config['REDIS_NS_VERSIONS_LOCATION'])
+    else:
+        raise Exception('One or more redis cache configuration options are missing from config.py')
```
alastair: custom_config.py

manage.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import print_function
+
 import db
-import db.cache
```
alastair: need to remove `import db.cache` line from dataset_eval/artistfilter.py

PR #228, opened 2017-03-08
AB-71 Make back button work in dataset editor + AB-95 Browsing a dataset messes with browsers' back button

nupurbaghel:
Fixed below issues : 
Browser back button not working in Class details of dataset Viewer and Editor
Job status url uses # as separator instead of / in Evaluation Tab of dataset Editor

Ticket ids: AB-71 and AB-95


webserver/static/scripts/datasets/eval-jobs-viewer.js
```diff
@@ -311,6 +313,12 @@ var JobDetails = React.createClass({
         result: React.PropTypes.object,
         onReturn: React.PropTypes.func.isRequired
     },
+    componentDidMount() { 
+        window.history.pushState(null,'job-details',this.props.id);
```
alastair: Be careful here about formatting - we put whitespace after commas

webserver/static/scripts/datasets/eval-jobs-viewer.js
```diff
@@ -61,14 +61,16 @@ var EvaluationJobsViewer = React.createClass({
...
+
     },
     componentWillUnmount: function() {
         window.removeEventListener('hashchange', this.handleHashChange);
     },
     handleHashChange: function(e) {
         // Hash is used to store ID of the currently viewed job.
         if (this.state.jobs) {
-            var hash = window.location.hash.substr(1);
+            //var hash = window.location.hash.substr(1);
```
alastair: We don't need comments here. Just remove it

webserver/static/scripts/datasets/eval-jobs-viewer.js
```diff
@@ -61,14 +61,16 @@ var EvaluationJobsViewer = React.createClass({
...
+            var hash = this.state.active_job_index;
```
alastair: This variable name made sense when we were reading the url hash (the part after the #). Perhaps now we should call it `job_id`

nupurbaghel:
All issues fixed other than the forward button. Added 2 new end points in datasets.py to enable direct viewing of class and job details. Hash related code removed from eval-jobs-viewer.js.

webserver/views/datasets.py
```diff
@@ -103,6 +113,15 @@ def eval_info(dataset_id):
         author=db.user.get(ds["author"]),
     )
 
+@datasets_bp.route("/<uuid:dataset_id>/evaluation/<uuid:job_id>")
```
alastair: Can we use the existing view and just make the job id optional? (http://stackoverflow.com/a/16569475). This prevents us from having to maintain the same code in two different places.

nupurbaghel:
Fixed forward button.
Also now we use the existing view with optional job id / classname.Other comments also taken care of.

paramsingh:
What happened here? 😅

paramsingh:
@brainzbot add to whitelist

alastair:
This still had significant issues when I last tested it, so I'm closing it. Hopefully we can come up with a more robust solution now that the react community is maturing more.

PR #232, opened 2017-03-14
Provide custom error handlers for API blueprints

saifulbkhan:
Some errors, such as 404 and 500, were still leading to an HTML page
(eg. when an invalid UUID was used in the url). These cases will now
be handled by the provided app_errorhandlers since flask does not
support blueprint level error-handling for 404 and 500 code errors.

### DIFF
--- a/webserver/errors.py
+++ b/webserver/errors.py
@@ -1,4 +1,4 @@
-from flask import render_template, jsonify
+from flask import render_template, jsonify, request
 from webserver.views.api import exceptions as api_exceptions
 
 
@@ -12,18 +12,31 @@ def init_error_handlers(app):
     def bad_request(error):
         return render_template('errors/400.html', error=error), 400
 
+    @app.errorhandler(401)
+    def unauthorized(error):
+        #always returning JSON, until we need a template for 401 error
+        return jsonify_error(error)
+    
     @app.errorhandler(403)
     def forbidden(error):
         return render_template('errors/403.html', error=error), 403
 
     @app.errorhandler(404)
     def not_found(error):
+        if request.path.startswith('/api/v1/'):
+            return jsonify_error(error)
         return render_template('errors/404.html', error=error), 404
 
     @app.errorhandler(500)
     def internal_server_error(error):
+        if request.path.startswith('/api/v1/'):
+            return jsonify_error(error)
         return render_template('errors/500.html', error=error), 500
 
     @app.errorhandler(503)
     def service_unavailable(error):
         return render_template('errors/503.html', error=error), 503
+
+def jsonify_error(error):
+    api_error = api_exceptions.APIError(error.description, error.code)
+    return jsonify(api_error.to_dict()), api_error.status_code
--- a/webserver/errors.py
+++ b/webserver/errors.py
@@ -1,4 +1,5 @@
 from flask import render_template, jsonify, request
+import webserver
 from webserver.views.api import exceptions as api_exceptions
 
 
@@ -10,11 +11,13 @@ def init_error_handlers(app):
 
     @app.errorhandler(400)
     def bad_request(error):
+        if request.path.startswith(webserver.API_PREFIX):
+            return jsonify_error(error)
         return render_template('errors/400.html', error=error), 400
 
     @app.errorhandler(401)
     def unauthorized(error):
-        #always returning JSON, until we need a template for 401 error
+        # always returning JSON because this is only raised from API endpoints
         return jsonify_error(error)
     
     @app.errorhandler(403)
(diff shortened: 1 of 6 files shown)


saifulbkhan:
Please follow these links explaining use of `@bp.app_errorhandler` decorator:
[flask issue 1935](https://github.com/pallets/flask/issues/1935)
[flask.Blueprint.errorhandler](http://flask.pocoo.org/docs/0.10/api/#flask.Blueprint.errorhandler)
[flask.Blueprint.app_errorhandler](http://flask.pocoo.org/docs/0.10/api/#flask.Blueprint.app_errorhandler)

A few things we could discuss:

- Do we need bp level errorhandlers for other error codes (400, 401, etc) as well? I don't see them being raised unless someone manually calls for them. I could be wrong though.
- If we add error handlers for other status codes maybe we should discard the custom API exceptions and use a seperate set of handlers for each blueprint. I made no change to the existing exception system for this initial commit. In fact I'm using it to create objects for convenience.
- I'm setting the error message for 404 and 500 statuses to the corresponding error.description. Let me know if this needs to change.

One last nitpick. The files involved import the `webserver.views.api.v1.exceptions` module in an inconsistent manner. This makes the changes non-uniform as well. Maybe we could do some formatting and grouping at the top. But that's a separate issue.

alastair:
Thanks for looking into this :+1:. Now I understand why trying `@blueprint.errorhandler()` wasn't working when I did it.
What happens with an API view decorated with `@auth_required`? This should result in a 401 if the user is not logged in, so we need to make sure that it returns json too.
You're right, 400 is only triggered once we are in a view, so we are using our custom exception handler.

I don't like the duplication which we have with the error handlers in each file. I think it would be easy for someone to create a new file/blueprint and forget to install the error handler.
For the setup of the app we install normal error handlers here:
https://github.com/saifulbkhan/acousticbrainz-server/blob/eeb2ea54750a4783820acffa010eaa3edb6c0f6f/webserver/__init__.py#L49-L50
and set up the blueprints here:
https://github.com/saifulbkhan/acousticbrainz-server/blob/eeb2ea54750a4783820acffa010eaa3edb6c0f6f/webserver/__init__.py#L103-L104

Could we do this:
```python
from webserver.errors import init_blueprint_error_handlers
    def register_api(app):
        v1_prefix = '/api/v1'
        from webserver.views.api.v1.core import bp_core
        from webserver.views.api.v1.datasets import bp_datasets
        from webserver.views.api.v1.dataset_eval import bp_dataset_eval
        app.register_blueprint(bp_core, url_prefix=v1_prefix)

        init_blueprint_error_handlers(bp_core)
        init_blueprint_error_handlers(bp_datasets)
```
and in `errors.py`:
```python
def init_blueprint_error_handlers(blueprint):
    @blueprint.app_errorhandler(404)
    def not_found(error):
        api_error = webserver.views.api.exceptions.APINotFound(error.description)
        return jsonify(api_error.to_dict()), api_error.status_code

    @blueprint.app_errorhandler(500)
    def internal_server_error(error):
        api_error = webserver.views.api.exceptions.APIError(error.description, 500)
        return jsonify(api_error.to_dict()), api_error.status_code
```

This way when someone adds a new blueprint to `register_api`, they can see that they also have to call the error handler init.

alastair:
I was looking at the documentation for `app_errorhandler`. It says

> Like Flask.errorhandler() but for a blueprint. This handler is used for all requests, even if outside of the blueprint.

Does this mean that it overwrites our existing error handlers? What's the behaviour of installing multiple error handlers for a code?

I wonder if we're going to have to do some more complex work here - some options that come to mind are
 * Check the URL in the error handler and return json if we are under `/api`
 * Run two distinct apps in 2 separate processes (we can do this easily with uwsgi) and have a specific web error handler and an api error handler. Will this correctly share sessions? (I think so, since they have the same secret?)

alastair:
It looks like my first option is listed as a recommendation in the ticket which you linked to

saifulbkhan:
oops! It is almost what you suggested. Let me tinker around a bit with these solutions. Thanks for reviewing :)

saifulbkhan:
I decided to go for the first option on the account that its much simpler than having another app just for handling API errors. Let me know what you think, and if we need tests or any documentation for this.

webserver/errors.py
```diff
@@ -10,6 +10,8 @@ def api_error(error):
 
     @app.errorhandler(400)
     def bad_request(error):
+        if request.path.startswith('/api/v1/'):
```
alastair: This should be a parameter shared with https://github.com/saifulbkhan/acousticbrainz-server/blob/61bae8d874be62d6ce2c2d5928056d2edb78b35f/webserver/__init__.py#L103

saifulbkhan:
We can also add a few `assert_template_used` in the tests for views to verify that they're being loaded when a non-API call error occurs. Let me know if you need that.

alastair:
@paramsingh can you take a look at this last commit? If you OK it I will fix the conflict and merge.

alastair:
Weird, I don't see those tests fail for me. Are you running with `pytest`? I wonder if it's something to do with fixtures or the database. I'll try and recreate my environment.

alastair:
Merged, thanks!

PR #233, opened 2017-03-15
AB-290 : Viewing the recording information of a dataset or a snapshot requires being logged in

nupurbaghel:
Removed the criteria of being logged in for viewing the recording info from datasets.

### DIFF
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -511,25 +511,25 @@ def update_class(dataset_id, class_name, class_data):
             connection.execute(update_query, params)
 
 
-def check_recording_in_dataset(dataset_id,mbid):
-    """Check whether mbid belongs to given dataset.
+def check_recording_in_dataset(dataset_id, mbid):
+    """Check whether an MBID is in a given dataset.
     Args:
-        dataset_id(string/uuid): ID of a dataset.
-        mbid(string/uuid): ID which is to be verified 
+        dataset_id: ID of a dataset
+        mbid: MBID to be checked
     Returns:
-        True/False depending on whether given mbid is a part of dataset 
+        True if an MBID appears anywhere in a dataset, False otherwise
     """
     with db.engine.connect() as connection:
         result = connection.execute(sqlalchemy.text("""
-            SELECT a.id from dataset_class as a
-             JOIN dataset_class_member as b
-             ON b.class = a.id 
-             WHERE a.dataset = :dataset_id and 
-             b.mbid = :mbid
+            SELECT dataset_class.id
+              FROM dataset_class
+              JOIN dataset_class_member
+                ON dataset_class_member.class = dataset_class.id 
+             WHERE dataset_class.dataset = :dataset_id
+               AND dataset_class_member.mbid = :mbid
         """), {"dataset_id": dataset_id, "mbid": mbid})
         
-        row = result.fetchone()
-        if not row:
-            return False
-        else:
+        if result.rowcount > 0:
             return True
+        else:
+            return False
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -528,8 +528,5 @@ def check_recording_in_dataset(dataset_id, mbid):
              WHERE dataset_class.dataset = :dataset_id
                AND dataset_class_member.mbid = :mbid
         """), {"dataset_id": dataset_id, "mbid": mbid})
-        
-        if result.rowcount > 0:
-            return True
-        else:
-            return False
+
+        return result.rowcount > 0
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -509,3 +509,27 @@ def update_class(dataset_id, class_name, class_data):
             params["id"] = clsid
 
             connection.execute(update_query, params)
+
+
+def check_recording_in_dataset(dataset_id,mbid):
+    """Check whether mbid belongs to given dataset.
+    Args:
+        dataset_id(string/uuid): ID of a dataset.
+        mbid(string/uuid): ID which is to be verified 
+    Returns:
+        True/False depending on whether given mbid is a part of dataset 
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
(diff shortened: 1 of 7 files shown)


alastair:
Thanks for the patch. As we mention in our getting started guide and on the wiki, understanding the problem before starting work on it is an important part of contributing to these projects.
We had a small discussion about this task when the PR came up. We think that removing the `login_required` decorator is not a good fix for this problem for a number of reasons:

1. This endpoint was originally developed for use in the dataset editor, where the user has to be logged in
2. AB-290 is related to viewing datasets when not logged in
3. Our general policy in MetaBrainz is to be as open as possible, so requiring users to log in to just view a dataset isn't necessarily a good idea
4. Having an open endpoint like this isn't a good idea either - it's possible (though unlikely) that a person may find this endpoint and start using it to get metadata from musicbrainz, bypassing the ratelimiting in place on the main musicbrainz API

There are probably two different ways that we could look at solving this problem

1. https://tickets.metabrainz.org/browse/AB-115 is a solution for having a copy of the musicbrainz database closer to the acousticbrainz server. This would make it easier to send metadata with the structure of the dataset in the API query, although it requires more up-front work (we have to finish AB-115 first)
2. This endpoint could stay open, but we could require that the client send both the dataset id and the musicbrainz recording id as parameters. We can do a check to see if the recording id is in the dataset, and only return metadata if this is the case. This will prevent large-scale scraping of musicbrainz metadata through acousticbrainz.

If you're interested in continuing with this could you move discussion back to Jira and discuss how you could implement option 2?

nupurbaghel:
Thanks a lot for appreciating! I have completed the cleanups suggested.

alastair:
@paramsingh ready for review

webserver/views/datasets.py
```diff
@@ -335,6 +332,22 @@ def recording_info(mbid):
...
+    """Endpoint for getting information about recordings (title and artist)."""
+    return _get_recording_info_for_mbid(mbid)
+
+
+@datasets_bp.route("/metadata/dataset/<uuid:dataset_id>/<uuid:mbid>")
+def recording_info_in_dataset(dataset_id, mbid):
+    """Endpoint for getting information about recordings (title and artist), for the
+     case when user is not logged in"""
+    if not db.dataset.check_recording_in_dataset(dataset_id, mbid):
+        return jsonify(error="Recording not found in the dataset"), 404
```
paramsingh: Would `raise NotFound("Recording not found in the dataset")` be better here?

db/dataset.py
```diff
@@ -509,3 +509,27 @@ def update_class(dataset_id, class_name, class_data):
...
+        result = connection.execute(sqlalchemy.text("""
+            SELECT dataset_class.id
+              FROM dataset_class
+              JOIN dataset_class_member
+                ON dataset_class_member.class = dataset_class.id 
+             WHERE dataset_class.dataset = :dataset_id
+               AND dataset_class_member.mbid = :mbid
+        """), {"dataset_id": dataset_id, "mbid": mbid})
+        
+        if result.rowcount > 0:
```
paramsingh: Just a `return result.rowcount > 0` would look better here, imo :)

webserver/views/datasets.py
```diff
@@ -335,6 +332,22 @@ def recording_info(mbid):
...
 
+@datasets_bp.route("/metadata/recording/<uuid:mbid>")
+@login_required
+def recording_info(mbid):
+    """Endpoint for getting information about recordings (title and artist)."""
+    return _get_recording_info_for_mbid(mbid)
+
+
+@datasets_bp.route("/metadata/dataset/<uuid:dataset_id>/<uuid:mbid>")
+def recording_info_in_dataset(dataset_id, mbid):
```
paramsingh: I think adding the expected behaviour of not returning data for recordings not in the dataset to the docstring here would be a good idea.

webserver/views/test/test_datasets.py
```diff
@@ -205,6 +211,24 @@ def test_recording_info(self):
         resp = self.client.get(url_for("datasets.recording_info", mbid=self.test_uuid))
         self.assert404(resp)
 
+    @mock.patch('webserver.external.musicbrainz.get_recording_by_id')
+    def test_recording_info_in_dataset(self, get_recording_by_id):
+        # self.test_mbid_1, is in the dataset
+        recording_mbid = "e8afe383-1478-497e-90b1-7885c7f37f6e"
```
paramsingh: why not just use `self.test_mbid_1` here instead of `recording_mbid`?

db/dataset.py
alastair: lucky you didn't see what I had before:
```
count = result.rowcount
if count > 0.....
```
in general I prefer it a bit more explicit, but I'm OK either way

webserver/views/datasets.py
alastair: I did originally suggest this, but there's a reason I didn't make the change (which I can't completely remember)...

We can use our `API*` exception classes to return json errors. We still have an issue with these endpoints that a call to them with bad uuids (for example) will return a flask 404 which will be html - same issue as https://github.com/metabrainz/acousticbrainz-server/pull/232 except we don't have a url namespace to filter on.
Two options would be:
 * put api endpoints which are used only by the interface under `/api` somwhere
 * see if it's possible to make a decorator that you can say "return any exception from here in html"

webserver/views/test/test_datasets.py
alastair: mostly because it was originally like this and I didn't change it. Good idea

webserver/views/datasets.py
paramsingh: I think making these new views API endpoints (first option) would be a better idea.

paramsingh: Maybe this should be done in a new patch, this has been standing for too long anyways.

paramsingh: [AB-319](https://tickets.metabrainz.org/browse/AB-319)

PR #234, opened 2017-09-09
AB-297: Add docker support for development

ferbncode:
Task remaining: 

- [x] Updating documentation (`README.md`)
- [x] Removing some redundant dependencies and packages installed in some containers dataset_evaluator` and `hl_extractor` (and some code being copied).
- [x] Running tests
- [x] Removing vagrant related files

PR #235, opened 2017-09-20
Upgrade javascript dependencies

alastair:
based on output of `npm outdated`. I left react, jquery, lodash, and highcharts
because they were a major version upgrade. Tested by running gulp
and checking the dataset editor/viewer pages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,30 +7,30 @@
     "url": "https://github.com/metabrainz/acousticbrainz-server.git"
   },
   "dependencies": {
-    "babel-core": "6.2.1",
-    "babel-plugin-transform-react-constant-elements": "6.1.20",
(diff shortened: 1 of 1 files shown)


paramsingh:
I'll download and build tomorrow or the day after, but do we not have Jenkins or Travis for AB?

alastair:
We used to have Jenkins on ci.metabrainz.org, but I don't know if this is correctly connected to github any more. I previously had a habit of not using travis because of the github permissions that it used to require, but I believe that this is now fixed (and we use it for LB anyway). Feel free to propose a patch to enable it and work with @zas to get it enabled.

paramsingh:
I'd like to do the travis setup after the docker PR #234 gets merged, I'll build this and test manually.

PR #236, opened 2017-09-21
Use MetaBrainz CustomFlask class for application setup

alastair:
This brings us in line with other MetaBrainz flask apps, for consistency


webserver/testing.py
```diff
@@ -1,14 +1,7 @@
...
-from webserver import create_app
 
 
-class ServerTestCase(flask_testing.TestCase, DatabaseTestCase):
-
-    def create_app(self):
-        app = create_app()
-        app.config['TESTING'] = True
-        return app
+class ServerTestCase(DatabaseTestCase):
```
alastair: I'm not sure about this file any more... The only thing that this gives us is the `temporary_login` method. We could just have a single base test class which we use for both database and application tests.

paramsingh: I think application tests tend to use the database a lot anyways, so I agree. I was thinking of something like a `ListenBrainzTestCase` (or `AcousticBrainzTestCase` here) which would create a webapp and init the db etc. I'll open a ticket.

paramsingh: [AB-311](https://tickets.metabrainz.org/browse/AB-311)

paramsingh:
I think documentation changes would be required too, with addition of custom_config.py ?

db/dump_manage.py
```diff
@@ -101,7 +104,8 @@ def _json_highlevel(location, rotate):
 @click.option("--id", type=int)
 @click.option("--threads", "-t", type=int)
 def incremental(location, id, threads):
-    db.init_db_engine(config.SQLALCHEMY_DATABASE_URI)
+    # inits the db engine
+    create_app()
```
paramsingh: Does it feel a bit weird to you to create the full web app just to init the db engine?

alastair: The idea here was to have only one place in the codebase that calls `db.init_db_engine`. The advantage of having it as part of the flask init is that we use flask's `app.config` inside `create_app`, which allows us to do things like override the database uri with the test config file. To do the same in `manage.py` we would have to specifically know which config file to import in this file.
Note that critiquebrainz does something similar (but not exactly the same) - https://github.com/metabrainz/critiquebrainz/blob/c7033487c5ce1beb45703193d47b90f4ec357ad9/manage.py#L98

alastair:
I'll update the documentation for the new config files, thanks for the reminder (will probably cause a conflict with #234, I'll look into the best way to not duplicate work there)

db/dump_manage.py
paramsingh: Ah okay, makes sense.

paramsingh:
hl_calc.py uses config which has been removed in this PR: https://github.com/metabrainz/acousticbrainz-server/blob/master/hl_extractor/hl_calc.py#L20

paramsingh:
```
param@krypton:~/meb/acousticbrainz-server$ git grep "import config"
admin/clean-tags.py:import config
admin/updates/20160616-snapshots-2.py:import config
dataset_eval/artistfilter.py:import config
dataset_eval/evaluate.py:import config
db/dump_manage.py:import config
db/testing.py:import config
fabfile.py:import config
hl_extractor/hl_calc.py:import config
hl_extractor/hl_clear_errors.py:import config
hl_extractor/job_calc.py:import config
manage.py:import config
webserver/__init__.py:    import config
webserver/upgrade_data_sha256.py:import config
webserver/views/api/v1/test/test_datasets.py:import config
```

PR #243, opened 2017-12-20
AB-317: Use Flask CLI for manage.py commands

paramsingh:
This stops us from having to create an app everytime db
needs to be initiated or something similar.

Note that we first have to `export FLASK_APP=manage.py` to get the cli to work.

```
$ export FLASK_APP=manage.py
$ flask runserver
* Running on http://0.0.0.0:8080/ (Press CTRL+C to quit)
 * Restarting with stat
```
and so on.


paramsingh:
I'm not satisfied with the environment variable that needs to be exported, tbh. Should try to find a better way...

alastair:
note that in docker we basically get this for free by setting an env variable in docker-compose.yml.

It looks like with the dev version of flask, if we have wsgi.py in the root it will automatically use this: 
http://flask.pocoo.org/docs/dev/cli/

alastair:
Can you also add new management commands for other utils (dataset_eval, hl_extractor)?

alastair:
as per https://tickets.metabrainz.org/projects/AB/issues/AB-321, we need to update the documentation to say how to run the helper utils.

manage.py
```diff
@@ -15,23 +15,19 @@
...
-@cli.command()
+@app.cli.command()
 @click.option("--host", "-h", default="0.0.0.0", show_default=True)
 @click.option("--port", "-p", default=8080, show_default=True)
-@click.option("--debug", "-d", type=bool,
-              help="Turns debugging mode on or off. If specified, overrides "
-                   "'DEBUG' value in the config file.")
-def runserver(host, port, debug):
-    app = create_app(debug=debug)
+def runserver(host, port):
```
alastair: flask comes with its own `run` command. Can we override that one here so that we don't have 2 with differing behaviour? (runserver has a different host/port and the extra_files arg, run doesn't)

hl_extractor/hl_calc.py
```diff
@@ -217,9 +221,11 @@ def main(num_threads):
             else:
                 break
 
-if __name__ == "__main__":
+
+@app.cli.command()
+@click.option("--threads", "-t", default=DEFAULT_NUM_THREADS, show_default=True, help="Number of threads to start")
+def hl_calc(threads):
```
alastair: What about making commands part of the top level manage.py? That way we only need to set one environment variable in order to run all commands.
Here's a proposal for a command hierarchy, I'm open to any other suggestions:

* flask run
  * flask ab init_db
  * flask ab compute_stats
  * flask ab add_admin   (all utils in manage.py go under some common commandclass)
    * flask ab dump full_db   (db command class goes under `ab`, or as `abdump`?)
    * flask ab util hl_cal
    * flask ab util dataset_eval  (not sure if additional apps should go under their own commandclass, or at toplevel `ab`)

alastair:
rebased against master with docker patches, and force pushed

PR #245, opened 2018-01-02
Fix AB-321: Added directory to the Python's module search path

RashiSah:
For Issue:
https://tickets.metabrainz.org/browse/AB-321

Initially, The errors were- "ImportError: No module named db" and "can not import webserver"  while running hl_calc.py in /hl_extractor and evaluate.py in /dataset_eval. 
Have added the path for the directory to be inside python's package search and added a variable in default_config.py

### DIFF
--- a/dataset_eval/evaluate.py
+++ b/dataset_eval/evaluate.py
@@ -8,7 +8,8 @@ import tempfile
 import time
 
 import gaia2.fastyaml as yaml
-
+import sys
+sys.path.insert(0,os.path.abspath('..'))
 import db
 import db.data
 import db.dataset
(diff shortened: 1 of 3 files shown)


alastair:
Thanks for the patch!
We explicitly started removing code which edits `sys.path` because it has the potential to cause weird and difficult to understand errors (https://github.com/metabrainz/acousticbrainz-server/commit/95e5e0294a6a7425b30ba7371238ca49aa869430)
To fix this issue we should just fix the documentation that explains how to run these tools.

paramsingh:
@alastair a little unrelated, but do you think it would be a good idea to create an `acousticbrainz` module that contains `webserver`, `db` and all these scripts like in LB, CB and MsB?

alastair:
@paramsingh I'm not against the idea (although we've already reorganised AB once...). I've opened https://tickets.metabrainz.org/browse/AB-322

PR #247, opened 2018-01-29
Update commands to run the tools

RashiSah:


paramsingh:
This isn't needed as it's already been fixed during the dockerization PR #234

alastair:
Thanks, Rashi, as @paramsingh said this patch is no longer needed!

PR #249, opened 2018-01-30
Use FlaskGroup to create the CLI interface

alastair:
This will automatically create the app and wrap commands in a flask context, which will connect to any required services (e.g., database, cache)

### DIFF
--- a/worker_manage.py
+++ b/worker_manage.py
@@ -10,14 +10,16 @@ import webserver
 cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app)
 
 
-@cli.command('hl_extractor', help='High-level extractor tool')
+@cli.command('hl_extractor')
 @click.option('--threads', '-t', default=1, type=int)
 def command_hl_extractor(threads=1):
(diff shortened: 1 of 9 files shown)


alastair:
@paramsingh ready

worker_manage.py
```diff
@@ -0,0 +1,25 @@
...
+
+import dataset_eval.evaluate
+import hl_extractor.hl_calc
+import webserver
+
+cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app)
+
+
+@cli.command('hl_extractor', help='High-level extractor tool')
+@click.option('-t', default=1, type=int)
```
paramsingh: Docstrings for this and the other command would be good to have.

paramsingh: Also, maybe using `threads` instead of `t` would be better.

alastair: I moved the help arguments to a docstring (click uses these as help if not explicitly set) and also cleaned up the docs in the main manage.py.

Added --threads as a long argument format.

Removed all create_app that wasn't needed

PR #250, opened 2018-02-02
manage.py with proper exit codes

RashiSah:

PR #252, opened 2018-02-08
manage.py with proper exit codes 

RashiSah:
Previous PR for this file: https://github.com/metabrainz/acousticbrainz-server/pull/250
I mistakenly pushed the last PR for this file directly from the master and it became difficult to handle other commits.
Kindly, reconsider this PR.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -133,6 +133,7 @@ def add_admin(username, force=False):
         click.echo("Made %s an admin." % username)
     except db.exceptions.DatabaseException as e:
         click.echo("Error: %s" % e, err=True)
+        sys.exit(-1)
 
 
 @cli.command()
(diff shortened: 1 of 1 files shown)


alastair:
Thanks for the patch :tada:. I made a small change to return `1` instead of `-1` and merged it  (a176bde43efdb716e226fde3ccf2958919123679),
 it's more common to use positive numbers for error codes. `1` is a standard "generic error" code which suits our purpose well.

I noticed that there was an error when running this script - you didn't import `sys`, which makes me think that you didn't actually test this change after making it. Please make sure that you test all changes that you make in the future before submitting the request!

RashiSah:
@alastair 
Thank you! :)
I will take care of it from now on.

PR #254, opened 2018-02-09
AB-330: Class name range check

rsh7:
Changed the range check of higher value from MIN to MAX.

### DIFF
--- a/utils/dataset_validator.py
+++ b/utils/dataset_validator.py
@@ -142,7 +142,7 @@ def validate_class(cls, idx=None, recordings_required=True):
     if not (CLASS_NAME_LEN_MIN <= len(cls["name"]) <= CLASS_NAME_LEN_MAX):
         raise ValidationException("Length of the `name` field in class%s doesn't fit the limits. "
                                   "Class name must be between %s and %s characters" %
-                                  (class_number_text, CLASS_NAME_LEN_MIN, CLASS_NAME_LEN_MIN))
+                                  (class_number_text, CLASS_NAME_LEN_MIN, CLASS_NAME_LEN_MAX))
 
     # Description (optional)
     if "description" in cls and cls["description"] is not None:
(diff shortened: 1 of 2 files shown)


alastair:
Thanks for this patch. Also, good catch with the bad variables in the error message! :clap: 
If you find two things to fix that are related, you can add them both to the same pull request. You don't need to open a separate one.

Please make sure you read the [CONTRIBUTING guidelines](https://github.com/metabrainz/acousticbrainz-server/blob/master/CONTRIBUTING.md), especially the section about writing good commit messages. Make sure your commit messages and pull request descriptions clearly describe the change that you have made.

utils/dataset_validator.py
```diff
@@ -139,10 +139,10 @@ def validate_class(cls, idx=None, recordings_required=True):
     # Name
```
alastair: We have some tests for the dataset validator at https://github.com/RashiSah/acousticbrainz-server/blob/b00f1f849ef2ba6041c308755c7de4dbcb90f699/utils/dataset_validator.py. It would be good to make sure that we have tests for these three methods that you changed so that we can be sure that they work as expected.

RashiSah: @alastair 
Can you please explain me a bit more about the tests you are talking about here: https://github.com/RashiSah/acousticbrainz-server/blob/b00f1f849ef2ba6041c308755c7de4dbcb90f699/utils/dataset_validator.py ?

alastair: Whoops, I meant to say https://github.com/RashiSah/acousticbrainz-server/blob/b00f1f849ef2ba6041c308755c7de4dbcb90f699/utils/test/test_dataset_validator.py

It looks like we have tests for most things regarding the structure of the dataset dictionary, but we don't have any tests to check that an error is shown if the names of classes or datasets are longer or shorter than the given values. We should add 3 more tests that provide names that are too short, and too long, and check that the correct exception is raised.

rsh7: Added tests for the dataset or class names that are too short or too long than the specified ranges.

alastair:
Great, thanks!

Just as a note, you can make long strings in python by doing something like 
```py
"a"*100
```
which results in a string that is just as long (but maybe a little more boring than yours :) )

PR #255, opened 2018-02-09
getMonth() returns the month (from 0-11) for a particular date

RashiSah:
For January, d.getMonth() returns value 0. That's why it displayed 00 as the month value on the main page.
Now it will return proper month values for every month.

### DIFF
--- a/webserver/static/scripts/homepage.js
+++ b/webserver/static/scripts/homepage.js
@@ -4,7 +4,7 @@ $(function() {
     if (lc_time > 0) {
         // lc_time of 0 means never
         var d = new Date(lc_time);
-        var collected_str = d.getFullYear() + "-" + ('0'+d.getMonth()).slice(-2) + "-" +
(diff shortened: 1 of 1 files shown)


alastair:
nice :100:. I can't even remember seeing the wrong month when I first wrote this code... I must have been too focused on making sure the timezone changed correctly!

I added a small comment to the JS file to remind us why we have this odd `+1` in the code (4251a1e137eacdeb1080e4ca6388e2453a96b798).

A reminder about the value of good commit messages. The message you have here explains how something works, but doesn't explain how the specific change that you made affects the AcousticBrainz server. As a suggestion, I would write something like this:

```
Increment the value of getMonth() by 1 when displaying

In Javascript, Date.getMonth() returns a 0-indexed value (i.e. January is 0) This means that
if we want to show it on screen we must increment it by 1
```
In this case we are saying *what* change the commit makes (in the first section), and *why* (in the second section). This is different from your original message which describes the behaviour of the API without any context to the change that was made.

Thanks for the contributions!

PR #257, opened 2018-02-16
AB-98: Add a feature to select SVM parameters

rsh7:
This is a...
- [ ]   Bug Fix
- [x]          Feature addition
- [ ]          Refactoring
- [ ]          Other

## **Present Case:**
When a dataset has been created, the user can submit it for evaluation.
before
This screen has 2 preferences (filtering and normalization) for now. We don't have any options to select which SVM parameters the user want to use so that the search space can be reduced.

## **Addition** 
I have added more additional preferences of `C` , `gamma` and `Preprocessing values` as well so that the user can be more specific on his/her requirements of the SVM parameters. And if the user doesn't provide any input, then the default values will be taken.
Now, the evaluation screen on `acousticbrainz server` will look like:
after

The additional preferences taken by the user are getting saved into database and then retrieved while model training stage. I am writing these values into the custom project file which is being generated during the evaluation stage before running the tests.
Thus, this feature will help the users to easily select which parameters they want to use.


paramsingh:
Hey @rsh7, in general whenever we do a schema change, we add a update script too. See examples here: https://github.com/metabrainz/acousticbrainz-server/blob/master/admin/updates/20160812-user-feedback.sql

paramsingh:
Also, tests seem to be failing, please run test.sh and fix them.

rsh7:
@paramsingh, Fixed the failing tests and added an update script as well.

admin/sql/create_tables.sql
```diff
@@ -104,17 +104,20 @@ CREATE TABLE dataset_snapshot (
...
+  status               eval_job_status          NOT NULL DEFAULT 'pending',
+  status_msg           VARCHAR,
+  options              JSONB,
+  training_snapshot    INT,                     -- FK to dataset_eval_sets
+  testing_snapshot     INT,                     -- FK to dataset_eval_sets
+  created              TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  updated              TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  result               JSONB,
+  eval_location        eval_location_type       NOT NULL DEFAULT 'local',
+  c_value              VARCHAR,
```
paramsingh: Should we consider having default values for these new columns?

dataset_eval/evaluate.py
```diff
@@ -59,12 +59,17 @@ def evaluate_dataset(eval_job, dataset_dir, storage_dir):
         groundtruth_path = os.path.join(eval_location, "groundtruth.yaml")
         with open(groundtruth_path, "w") as f:
             yaml.dump(create_groundtruth_dict(snapshot["data"]["name"], train), f)
-
+        """
```
paramsingh: In general, `"""` is used for [docstrings](https://www.python.org/dev/peps/pep-0257/#id15) and `#` for comments that explain what the code is doing. You should change this comment to `# Passing more user preferences to train the model`.

dataset_eval/evaluate.py
```diff
@@ -59,12 +59,17 @@ def evaluate_dataset(eval_job, dataset_dir, storage_dir):
...
+        Passing more user preferences to train the model
+        """
         logging.info("Training model...")
         results = gaia_wrapper.train_model(
             project_dir=eval_location,
             groundtruth_file=groundtruth_path,
             filelist_file=filelist_path,
+            c_value = eval_job["c_value"],
```
paramsingh: just a small nitpick, but no space before and after operators in function calls. see the line above this.

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +46,29 @@ def train_model(project_dir, groundtruth_file, filelist_file):
         datasets_dir=os.path.join(project_dir, "datasets"),
         results_dir=os.path.join(project_dir, "results"),
     )
+    """
```
paramsingh: should be a `#` comment

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +46,29 @@ def train_model(project_dir, groundtruth_file, filelist_file):
...
+    Converting the preferences from string type to list
+    """
+    c_value = [int(value) for value in c_value.split(',')]
+    gamma_value = [int(s) for s in gamma_value.split(',')]
+    preprocessing_values = [str(value) for value in preprocessing_values.split()]
+    """
```
paramsingh: should be a `#` comment

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +46,29 @@ def train_model(project_dir, groundtruth_file, filelist_file):
...
+    Updating C, Gamma, and preprocessing values in the project file which were
+    taken as input by the user
+    """
+    update_parameters(project_file, c_value, gamma_value, preprocessing_values)
     run_tests(project_file)
     return select_best_model(project_dir)
 
+def update_parameters(project_file, c_value, gamma_value, preprocessing_values):
```
paramsingh: We should have a docstring explaining what this function does and what values its arguments take. See https://github.com/metabrainz/guidelines/blob/master/Python.md#docstrings

db/dataset_eval.py
```diff
@@ -273,7 +283,9 @@ def add_dataset_eval_set(connection, data):
     return snapshot_id
 
 
-def _create_job(connection, dataset_id, normalize, eval_location, filter_type=None):
+def _create_job(connection, dataset_id, normalize, eval_location, c_value, gamma_value,
```
paramsingh: maybe we should consider using default values for the arguments `c_value`, `gamma_value` and `preprocessing_values`, that would mean that we don't have to change all the tests.

db/dataset_eval.py
```diff
@@ -283,8 +295,12 @@ def _create_job(connection, dataset_id, normalize, eval_location, filter_type=No
         raise ValueError("Incorrect 'eval_location'. Must be one of %s" % VALID_EVAL_LOCATION)
     snapshot_id = db.dataset.create_snapshot(dataset_id)
     query = sqlalchemy.text("""
-                INSERT INTO dataset_eval_jobs (id, snapshot_id, status, options, eval_location)
-                     VALUES (uuid_generate_v4(), :snapshot_id, :status, :options, :eval_location)
+                INSERT INTO dataset_eval_jobs (id, snapshot_id, status, options, eval_location,
```
paramsingh: This query could be formatted better: https://github.com/metabrainz/guidelines/blob/master/Python.md#sql


I don't think we mind lines of ~130 characters so this could work like

```
INSERT INTO dataset_eval_jobs (id, snapshot_id, status, options, eval_location, _value, gamma_value, preprocessing_values)
     VALUES (uuid_generate_v4(), :snapshot_id, :status, :options, :eval_location, :c_value, :gamma_value, :preprocessing_values)
  RETURNING id
```

Example: https://github.com/metabrainz/critiquebrainz/blob/master/critiquebrainz/db/review.py#L295

webserver/forms.py
```diff
@@ -14,6 +15,16 @@
 DATASET_DONE = "done"
 DATASET_ALL = "all"
 
+DATASET_C_VALUE = ' -5, -3, -1, 1, 3, 5, 7, 9, 11 '
```
paramsingh: why the trailing and leading whitespace?

webserver/forms.py
```diff
@@ -14,6 +15,16 @@
...
+DATASET_GAMMA_VALUE = '3, 1, -1, -3, -5, -7, -9, -11'
+PREPROCESSING_VALUES = [ 'basic', 'lowlevel', 'nobands', 'normalized', 'gaussianized' ]
+PREPROCESSING_VALUES = [(value, value) for value in PREPROCESSING_VALUES]
```
paramsingh: This could be done in one line with a comment to clarify why we need a list of pairs here.

`PREPROCESSING_VALUES = [('basic', 'basic'), ...]`

webserver/forms.py
```diff
@@ -34,4 +45,15 @@ class DatasetEvaluationForm(Form):
         default = DATASET_EVAL_LOCAL
     )
     normalize = BooleanField("Normalize classes")
-
+    """
```
paramsingh: should be a `#` comment

admin/sql/create_tables.sql
alastair: we already have the `options` field which stores some user-provided settings for an evaluation. I think that we should store these items in there instead of having new parameters.

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +46,29 @@ def train_model(project_dir, groundtruth_file, filelist_file):
         datasets_dir=os.path.join(project_dir, "datasets"),
         results_dir=os.path.join(project_dir, "results"),
     )
+    """
+    Converting the preferences from string type to list
+    """
+    c_value = [int(value) for value in c_value.split(',')]
```
alastair: We need error checking here - if a person enters a string then the `int()` cast will fail. I think it's OK to silently skip non-numerical values here

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +46,29 @@ def train_model(project_dir, groundtruth_file, filelist_file):
...
+    """
+    update_parameters(project_file, c_value, gamma_value, preprocessing_values)
     run_tests(project_file)
     return select_best_model(project_dir)
 
+def update_parameters(project_file, c_value, gamma_value, preprocessing_values):
+    with open(project_file, "r") as pfile:
+        project = yaml.load(pfile)
+    for pref in project['classifiers']['svm']:
+        pref['gamma'], pref['C'], pref['preprocessing'] = gamma_value, c_value, \
```
alastair: I don't think there's a reason that this has to be a multi-assignment statement. Three separate statements is a bit easier to follow

dataset_eval/gaia_wrapper.py
alastair: In addition to @paramsingh's note about the syntax of the comment, consider the purpose of adding them. In this case I can see that you're converting a string to a list, because I can read the code. A good rule of thumb is to say *why* you are doing something, not *what* you are doing. In this case, I'd consider saying

    # skip invalid non-numerical c & gamma values

(with the updated code to actually skip these values)

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +46,29 @@ def train_model(project_dir, groundtruth_file, filelist_file):
...
         results_dir=os.path.join(project_dir, "results"),
     )
+    """
+    Converting the preferences from string type to list
+    """
+    c_value = [int(value) for value in c_value.split(',')]
+    gamma_value = [int(s) for s in gamma_value.split(',')]
+    preprocessing_values = [str(value) for value in preprocessing_values.split()]
+    """
+    Updating C, Gamma, and preprocessing values in the project file which were
```
alastair: A very small thing here - the comment is good, but we typically write comments in the imperative mood (git commit messages too). In this case, I would say
> Update the project file with user-provided C, gamma, and preprocessing values

(note the use of 'update', not 'updating')

rsh7:
I have added a feature flag like https://github.com/metabrainz/acousticbrainz-server/blob/fbb9a89774ea57535c8592aa73299abee3015f9c/default_config.py#L62 to only show these preferences option on the acousticbrainz site when it is enabled.
So, if the value is set to False, it will display:

new

admin/sql/create_tables.sql
rsh7: As we also have a parameter of `evaluation_location` here: https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/forms.py#L31, which is enabled to display on the acousticbrainz site when https://github.com/metabrainz/acousticbrainz-server/blob/master/default_config.py#L62 is turned `True`.
Similarly, the c, gamma and preprocessing parameters are added separately.
Should I change it? :)

dataset_eval/gaia_wrapper.py
rsh7: c_value is getting stored in the database as `string` type. for example: '1 2 3 4 5'
I am converting it to a list of integers similar to what is saved as default value in https://github.com/MTG/gaia/blob/1d593b89a65cbf0a6b3f1e70cb34e8d067dc6fd5/src/bindings/pygaia/scripts/classification/classification_project_template.yaml#L99

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +44,35 @@ def train_model(project_dir, groundtruth_file, filelist_file):
...
+    preprocessing_values = [str(value) for value in preprocessing_values.split()]
+    # Update the project file with user-provided C, gamma, and preprocessing values
+    update_parameters(project_file, c_value, gamma_value, preprocessing_values)
     run_tests(project_file)
     return select_best_model(project_dir)
 
+def update_parameters(project_file, c_value, gamma_value, preprocessing_values):
+    """Update the project file.
+
+    An optional, more detailed description of this function if needed.
```
paramsingh: Should replace this line with a description :)

admin/sql/create_tables.sql
paramsingh: Storing in the `options` field would remove the need for a schema change. I'd prefer it but let us wait for @alastair to confirm.

paramsingh: @rsh7 I think we should go with moving the values in the new columns to options.

dataset_eval/evaluate.py
```diff
@@ -60,11 +60,15 @@ def evaluate_dataset(eval_job, dataset_dir, storage_dir):
         with open(groundtruth_path, "w") as f:
             yaml.dump(create_groundtruth_dict(snapshot["data"]["name"], train), f)
 
+        # Passing more user preferences to train the model.
```
paramsingh: I don't think this comment is needed here really, but if it is, `Pass user preferences to train the model` would read better. Note alastair's comment about tense in comments.

dataset_eval/gaia_wrapper.py
```diff
@@ -44,9 +44,31 @@ def train_model(project_dir, groundtruth_file, filelist_file):
...
+    # Update the project file with user-provided C, gamma, and preprocessing values
+    update_parameters(project_file, c_value, gamma_value, preprocessing_values)
     run_tests(project_file)
     return select_best_model(project_dir)
 
+def update_parameters(project_file, c_value, gamma_value, preprocessing_values):
+    """Update the project file with user-provided preferences 
+
+    Args:
+        project_file: The file to be updated.
```
paramsingh: Please mention the type of these args in the docstrings too.

db/dataset_eval.py
```diff
@@ -24,8 +24,13 @@
 # Filter types are defined in `eval_filter_type` type. See schema definition.
 FILTER_ARTIST = "artist"
 
+# Default values for parameters
+preprocessing = ['basic', 'lowlevel', 'nobands', 'normalized', 'gaussianized']
```
paramsingh: these default values are repeated three times in this PR, once here, once in [forms.py](https://github.com/metabrainz/acousticbrainz-server/pull/257/files#diff-d426557fa42c5459f53956bcfdc1e0aaR18) and once in [views/datasets.py](https://github.com/metabrainz/acousticbrainz-server/pull/257/files#diff-b1772f3502d312f3d84deae191687459R19). In general, it is a good idea to avoid duplication of code and constants as much as possible. Please define these in `dataset_eval/__init__.py` and import them, reducing duplication. 

https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

webserver/views/datasets.py
```diff
@@ -167,10 +171,23 @@ def evaluate(dataset_id):
...
             db.dataset_eval.evaluate_dataset(
                 dataset_id=ds["id"],
                 normalize=form.normalize.data,
                 eval_location=form.evaluation_location.data,
+                c_value=c_value,
+                gamma_value=gamma_value,
+                preprocessing_values=form.preprocessing_values.data,
                 filter_type=form.filter_type.data,
             )
             flash.info("Dataset %s has been added into evaluation queue." % ds["id"])
```
paramsingh: I think flashing a message here if the input values were bad would be a good idea too.

webserver/templates/datasets/evaluate.html
```diff
@@ -39,6 +39,22 @@ <h2 class="page-title">Evaluate dataset "{{ dataset['name'] }}"</h2>
       </div>
       {% endif %}
     </fieldset>
+    <fieldset>
+      {% if config.get('FEATURE_MODEL_SELECTION') %}
+      <div class="form-group">
```
paramsingh: This should be indented two spaces.

webserver/views/datasets.py
```diff
@@ -167,10 +171,23 @@ def evaluate(dataset_id):
         try:
             if form.filter_type.data == forms.DATASET_EVAL_NO_FILTER:
                 form.filter_type.data = None
+            # take default value
+            # if invalid non-numerical c and gamma
+            try:
+                c_value = [int(value) for value in form.c_value.data.split(',')]
+            except:
```
paramsingh: Instead of catching all exceptions, mentioning which error you're looking for here (`ValueError`) would be a good idea.

rsh7:
Rebased to master!

alastair:
🎉

PR #258, opened 2018-02-17
AB-139: Enabled sign in option on error pages

rsh7:
This is a...
- [x]   Bug Fix
- [ ]          Feature addition
- [ ]          Refactoring
- [ ]          Other

**Problem:** 
Error pages are missing the Sign in button.

**Solution:**
I've enabled the `Sign in` button on HTTPS500 error pages, and `Sign in` button and corresponding drop-down with `Your profile` and sign out button for all other error pages as well.

### DIFF
--- a/webserver/errors.py
+++ b/webserver/errors.py
@@ -1,4 +1,5 @@
-from flask import render_template, jsonify, request
+from flask import render_template, jsonify, request, has_request_context, _request_ctx_stack, current_app
+
 import webserver
 from webserver.views.api import exceptions as api_exceptions
 
@@ -13,7 +14,7 @@ def init_error_handlers(app):
     def bad_request(error):
         if request.path.startswith(webserver.API_PREFIX):
             return jsonify_error(error)
-        return render_template('errors/400.html', error=error, code = error.code), 400
+        return render_template('errors/400.html', error=error), 400
 
     @app.errorhandler(401)
     def unauthorized(error):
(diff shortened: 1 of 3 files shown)


paramsingh:
We compromised and showed the sign-in button only when the user is not signed-in on CB error pages. (https://github.com/metabrainz/critiquebrainz/pull/178) This kinda addressed gentlecat's original concerns of page not rendering if user isn't loaded while showing a sign-in button. Maybe we'd wanna do something similar here?

alastair:
As @paramsingh said, we can't unconditionally show this menu on all error pages. See also [AB-74](https://tickets.metabrainz.org/projects/AB/issues/AB-74), which shows that we can have problems when we try and show an error page and the error was due to a problem with the database.

The possibility of errors with the database should only happen on the HTTP500 page. An alternative could be to show the user menu normally on 40x pages, but only show 'Sign in' if it's a 500 page. @paramsingh what do you think?

paramsingh:
Sounds logical to me, other than being a little weird from a UI perspective, I guess.

alastair:
You mean that going from logged in to 'sign in' on a 500 page is weird? I'm not sure what else we could do though, unless we store the username in a session/cookie...

rsh7:
This change enables the `sign in` button on error pages when a user is not signed in already.
Should this be reflected only on a 500 page, or is it okay with other error pages as well for now?

rsh7:
Accordingly as @alastair mentioned above, now the user menu will be shown normally on 40x pages, but only the 'Sign in' button if it's a 500 page.

rsh7:
Hi @alastair , I saw your work there and was also thinking to work similarly on AB. I will make a commit today.

alastair:
Thanks for the update. I've merged this to master - I fixed the test errors just before you did.

Remember that tests are an important part of the AcousticBrainz codebase. It's normal to make sure that tests pass before you push to github - it often does not make sense to make a code change in one commit, and then fix tests in a separate commit. Common changes should be grouped in one commit.

rsh7:
Okay, I will take care of it from now on. :)

PR #259, opened 2018-02-21
AB-333: Use 'FlaskGroup' for creating CLI interface in 'dump_manage.py'

ferbncode:
https://tickets.metabrainz.org/browse/AB-333

### DIFF
--- a/db/dump_manage.py
+++ b/db/dump_manage.py
@@ -8,7 +8,7 @@ import os
 import webserver
 
 
-cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app)
+cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app_flaskgroup)
 
 
 @cli.command()
--- a/db/dump_manage.py
+++ b/db/dump_manage.py
@@ -1,12 +1,14 @@
 from __future__ import print_function
+from flask.cli import FlaskGroup
 from db import dump
 import shutil
 import click
 import re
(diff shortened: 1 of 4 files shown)


paramsingh:
@brainzbot add to whitelist

db/dump_manage.py
```diff
@@ -1,12 +1,14 @@
...
 from db import dump
 import shutil
 import click
 import re
 import os
+import webserver
 
 
-cli = click.Group()
+cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app)
```
alastair: I don't think this is correct - http://flask.pocoo.org/docs/0.12/cli/#custom-scripts says 

> The next thing we do is defining a function that is invoked with the script info object (ScriptInfo) from Flask and its purpose is to fully import and create the application.

In this case, the `FlaskGroup` will call `create_app` with script_info as the first parameter. We need to have a separate factory function that accepts script_info, and calls to our create_app.

paramsingh: Weird thing is that it works without an extra factory function too. https://github.com/metabrainz/acousticbrainz-server/blob/master/manage.py#L22

alastair: That's a bug too, then :)
Our factory function is able to take a parameter, so perhaps it's incorrectly trying to load the script_info as a config file???!?

alastair:
nice 👍

PR #261, opened 2018-02-26
AB-21: Get high-level data for multiple recordings with one request

rsh7:
This is a...
- [ ]   Bug Fix
- [x]          Feature addition
- [ ]          Refactoring
- [ ]          Other

**Present Case:** 
We don't have any feature which provides high-level data with one request for many `recording ids`.

**Solution:**
I have added a feature through which we can fetch JSON documents for a list of MusicBrainz recording IDs for high-level data.
The user can construct the endpoint (Suppose for a list of 3 MBIDs) for high-level data by:
 GET https://acousticbrainz.org/api/v1/high-level?recording_ids=mbid;mbid;mbid

### DIFF
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -205,6 +205,55 @@ def get_many_lowlevel():
     return jsonify(recording_details)
 
 
+@bp_core.route("/high-level", methods=["GET"])
+@crossdomain()
+def get_many_highlevel():
+    """Get high-level data for many recordings at once.
+
+    **Example response**:
+
+    .. sourcecode:: json
+
+       {"mbid1": {"offset1": {document},
+                  "offset2": {document}},
+        "mbid2": {"offset1": {document}}
+       }
+
+    Offset keys are returned as strings, as per JSON encoding rules.
+    If an offset was not specified in the request for an mbid, the offset
+    will be 0.
+
+    If the list of MBIDs in the query string has a recording which is not
+    present in the database, then it is silently ignored and will not appear
+    in the returned data.
+
+    :query recording_ids: *Required.* A list of recording MBIDs to retrieve
+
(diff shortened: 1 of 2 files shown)


paramsingh:
Also, should probably add a test for the new endpoint.

webserver/views/api/v1/core.py
```diff
@@ -176,35 +189,45 @@ def get_many_lowlevel():
...
         except NoDataFoundException:
             pass
 
     return jsonify(recording_details)
 
 
+@bp_core.route("/low-level", methods=["GET"])
+@crossdomain()
+def get_many_lowlevel():
+    """Get low-level data for many recordings at once.
```
paramsingh: The previous docstring that was here was used to automatically generate documentation on readthedocs for the endpoint. We should keep it and add a similar docstring for the new endpoint added.

paramsingh: https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/views/api/v1/core.py# and https://acousticbrainz.readthedocs.io/api.html#get--api-v1-low-level

webserver/views/api/v1/core.py
```diff
@@ -155,6 +155,37 @@ def _parse_bulk_params(params):
     return [x for x in ret if not (x in seen or seen.add(x))]
 
 
+def check_bad_request_for_multiple_recordings():
+    """Checking whether the recording ids throw exceptions
```
paramsingh: You should mention exactly what exceptions will be thrown and when. Also, what happens if the data is correct.

rsh7:
@paramsingh, does this PR need more changes?

alastair:
good patch, thanks so much. sorry for taking so long to merge it 😱

rsh7:
Thank you for the merge!

PR #262, opened 2018-03-04
AB-334: MB database image setup in AB and fetch data for recording entity

rsh7:
Added a container of the [image](https://github.com/metabrainz/musicbrainz-server/blob/master/docker/templates/Dockerfile.test-database.m4) in AB in order to test using MB data at the scale of AcousticBrainz. The volume for the container is specified to be `acousticbrainz-server/data/mbdata.` Access the MB database directly for `recording` entity without use of web service.

Instructions to import the data:
To import the database dumps in a single command, use: docker-compose -f docker/docker-compose.dev.yml run musicbrainz_database


paramsingh:
I'm not sure if we want to merge this into master or if we want to create new branches for each of the two methods of direct db access we're considering. @alastair ?

paramsingh:
Also, you should fix tests for this, I guess we'd have to write new tests for the musicbrainz_db module?

rsh7:
@paramsingh , Yeah I would add tests to the musicbrainz_database module.

rsh7:
There should be separate PR's for image setup and use of MB database in AB. And we would use Brainzutils to access database for different entities. 
Closing this.

PR #266, opened 2018-03-22
Upgrade JS dependencies

alastair:
Upgrade JS dependences as listed by `npm outdated`
Tested that gulp still generates the bundle as expected, and js parts of the site work.

Leaving React for now, as it will require code changes to do the upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,24 +13,24 @@
     "babel-plugin-transform-react-jsx": "6.24.1",
     "babel-preset-es2015": "6.24.1",
     "babel-preset-stage-2": "6.24.1",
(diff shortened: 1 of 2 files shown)


alastair:
The only other package (other than bootstrap and react) waiting for an upgrade was highcharts - I went ahead and upgraded that too (one small code change). The graphs have changed a bit (in the way the legend shows), but I don't think we need to alter them at all:

highcharts 4:
screenshot from 2018-03-27 17-27-41

highcharts 6

screenshot from 2018-03-27 17-27-35

paramsingh:
Looks better, tbh!

PR #274, opened 2018-06-16
 AB-344: Add a force option to manage.py init_mb_db

rsh7:
Add a force option to init_mb_db:
* That drops the musicbrainz schema if exists, along with all the tables and keys.
* And then recreates the schema structure.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -99,7 +99,7 @@ def init_db(archive, force, skip_create_db=False):
 
 
 @cli.command()
-@click.option("--force", "-f", is_flag=True, help="Drop existing musicbrainz schema and tables.")
+@click.option("--force", "-f", is_flag=True, help="Drop existing MusicBrainz schema and tables.")
 def init_mb_db(force):
     """Initialize the MusicBrainz database.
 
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -99,7 +99,8 @@ def init_db(archive, force, skip_create_db=False):
 
 
 @cli.command()
-def init_mb_db():
+@click.option("--force", "-f", is_flag=True, help="Drop existing musicbrainz schema and tables.")
```
paramsingh: Since this is a user-facing message, the help should say `MusicBrainz` instead of `musicbrainz.

manage.py
```diff
@@ -111,6 +112,11 @@ def init_mb_db():
 
     musicbrainz_db.init_db_engine(current_app.config['MB_DATABASE_URI'])
 
+    if force:
+        res = db.run_sql_script_without_transaction(os.path.join(ADMIN_SQL_DIR, 'drop_musicbrainz_schema.sql'))
```
paramsingh: You should print a message saying `Dropping MusicBrainz schema...` here.

rsh7:
Updated!

PR #278, opened 2018-06-20
AB-346: Import MusicBrainz data for every new recording added to AB db

rsh7:
I have wrote a script called mb_import.py which runs continously with the time gap of 30 seconds. Every 30 seconds it searches for a new recording if added to low-level table in AcousticBrainz db. If it finds any new recording, MusicBrainz data for the corresponding recording gets imported and thus keeps the musicbrainz schema in AcousticBrainz updated.


db/mb_import.py
```diff
@@ -0,0 +1,33 @@
...
+import db.import_mb_data
+import logging
+from sqlalchemy import text
+import time
+
+SLEEP_DURATION = 30  # number of seconds to wait between runs
+
+
+def main():
+    logging.info("Checking if any import is required...")
```
paramsingh: This message should be `musicbrainz_importer` started.

db/mb_import.py
```diff
@@ -0,0 +1,33 @@
+import db
```
paramsingh: let's rename this file to musicbrainz_importer and move it to `acousticbrainz-server/musicbrainz_importer/musicbrainz_importer.py`

db/mb_import.py
```diff
@@ -0,0 +1,33 @@
...
+            logging.info("Updating AcousticBrainz database...")
+            db.import_mb_data.fetch_and_insert_musicbrainz_data(gids_in_AB)
+        else:
+            logging.info("No new recording found. Sleeping %s seconds." % SLEEP_DURATION)
+            time.sleep(SLEEP_DURATION)
+
+
+def get_new_recordings_from_AB():
+    with db.engine.begin() as connection:
+        query = text("""SELECT lowlevel.gid
```
paramsingh: You should apply a limit to the number of recordings you import in a batch here. The value should be taken from the config files.

rsh7: I thought the same initially. The time gap of checking and importing the data for newly added recordings is 30 seconds only. Should we still add a limit here?

rsh7: I thought the same initially. As the time gap of checking the newly added recordings for import is 30 seconds only, so should we still add a limit here?

db/mb_import.py
rsh7: :+1:

db/mb_import.py
rsh7: Added a new musicbrainz_importer folder.

db/mb_import.py
paramsingh: Good point. I was thinking that what we'd do in prod is just run this script (without running `python manage.py mb_import`) and ideally it would just import all recordings over time. That would require batches here.

rsh7: Got it. Done! :baby:

musicbrainz_importer/musicbrainz_importer.py
```diff
@@ -0,0 +1,43 @@
...
+        offset = 0
+        rows_to_fetch = current_app.config['RECORDINGS_FETCHED_PER_BATCH']
+
+        query = text("""SELECT lowlevel.gid
+                          FROM lowlevel
+                     LEFT JOIN musicbrainz.recording
+                            ON lowlevel.gid = musicbrainz.recording.gid
+                         WHERE musicbrainz.recording.gid is NULL
+                      ORDER BY lowlevel.id
+                        OFFSET :offset
```
paramsingh: Why do you need offset here? The query would work fine without it.

musicbrainz_importer/musicbrainz_importer.py
```diff
@@ -0,0 +1,43 @@
...
+        if gids_in_AB:
+            logging.info("Importing MusicBrainz data...")
+            logging.info('Inserting data for %d recordings...' % (rows_to_fetch))
+            db.import_mb_data.fetch_and_insert_musicbrainz_data(gids_in_AB)
+        else:
+            logging.info("No new recording found. Sleeping %s seconds." % SLEEP_DURATION)
+            time.sleep(SLEEP_DURATION)
+
+
+def get_new_recordings_from_AB():
```
paramsingh: Because this is a Sql query, I think we should move this to the db module. I should have mentioned this in the previous review, sorry.

musicbrainz_importer/musicbrainz_importer.py
```diff
@@ -0,0 +1,43 @@
...
+                      ORDER BY lowlevel.id
+                        OFFSET :offset
+                         LIMIT :rows_to_fetch
+        """)
+        gids = connection.execute(query, {"offset": offset, "rows_to_fetch": rows_to_fetch})
+        gids = gids.fetchall()
+        gids_in_AB = [value[0] for value in gids]
+        offset = offset + rows_to_fetch
+
+        return gids_in_AB, rows_to_fetch
```
paramsingh: Is there a need to return rows_to_fetch?

rsh7: I am returning it to print a message of - Inserting 'rows_to_fetch' no. of recordings.

paramsingh: Shouldn't `len(gids)` be printed instead? It isn't necessary that you'll have `rows_to_fetch` recordings.

musicbrainz_importer/musicbrainz_importer.py
rsh7: Yeah, point. Removing offset.

db/get_lowlevel_recordings.py
```diff
@@ -0,0 +1,21 @@
+import db
```
paramsingh: oh, maybe I was unclear, this should go in db/data.py

paramsingh: and we should add a test for this too.

PR #281, opened 2018-06-28
AB-348: Make the musicbrainz import more efficient

rsh7:


db/import_mb_data.py
```diff
@@ -48,12 +66,12 @@ def load_artist_credit(connection, gids_in_AB, MB_release_data):
     return MB_artist_credit_data
 
 
-def load_artist_type(connection, gids_in_AB):
+def load_artist_type(connection, gids_in_AB, artist_credit_from_recording):
```
paramsingh: Is the var `gids_in_AB` still needed here?

rsh7: No, not needed and I've removed it.

rsh7:
Data import was pretty fast, but data insertion was taking much time as there were a lot of duplications in the data fetched from many tables. So, I have again included the DISTINCT keyword in the queries. And also improved some queries to get all the data from MusicBrainz db as the number of rows in some tables are less than 20.

paramsingh:
Rebase this once, please.

rsh7:
Rebased, and add a commit for some changes in docstrings of all the functions.

db/import_mb_data.py
```diff
@@ -248,117 +179,53 @@ def load_end_area_type(connection, gids_in_AB):
...
+    """)
+    result = connection.execute(release_status_query)
     MB_release_status_data = result.fetchall()
 
     return MB_release_status_data
 
 
-def load_release_group_primary_type(connection, gids_in_AB, MB_release_group_data):
+def load_release_group_primary_type(connection):
     """Fetch release_group_primary_type table data from MusicBrainz database for the
```
paramsingh: now that these functions are fetching entire tables, we should change docs to reflect that.

db/import_mb_data.py
```diff
@@ -80,123 +81,53 @@ def load_artist_credit(connection, gids_in_AB, MB_release_data, MB_release_group
...
-    INNER JOIN artist_credit
-            ON artist.id = artist_credit.id
-    INNER JOIN recording
-             ON artist_credit.id = recording.artist_credit
-             {filterstr}
-    """.format(filterstr=filterstr)
-    )
-    result = connection.execute(area_type_query, filter_data)
+    area_type_query = text("""
+        SELECT * FROM area_type
```
paramsingh: this should be


```
  SELECT column1,
         column2
    FROM area_type
ORDER BY id
```

rsh7:
Changes made.

PR #284, opened 2018-07-13
Add consul_config template and use it for docker deployment environments

paramsingh:
Also, this PR moves back to the single config.py model.

We encountered many problems when using multiple config files in LB ranging from different config values in different parts of the code to race conditions and so on. We decided that moving to a single config file would be better overall.


paramsingh:
@brainzbot retest this please

paramsingh:
Jenkins is kinda fragile, I've been wrestling for a few hours with this, taking a break, I'll fix soon.

paramsingh:
@brainzbot retest this please

paramsingh:
@brainzbot retest this please

consul_config.py.ctmpl
```diff
@@ -0,0 +1,51 @@
...
+        "{{.}}",
+    {{- end -}}
+{{- end -}}
+
+import os
+
+DEBUG = False
+SECRET_KEY = '''{{template "KEY" "secret_key"}}'''
+
+{{if service "pgbouncer-master"}}
```
paramsingh: this will need to change to the new postgres service

paramsingh:
See https://github.com/metabrainz/docker-server-configs/pull/46 also

consul_config.py.ctmpl
```diff
@@ -0,0 +1,51 @@
...
+
+{{if service "acousticbrainz-pgbouncer"}}
+{{with index (service "pgbouncer-master") 0}}
+SQLALCHEMY_DATABASE_URI = "postgresql://acousticbrainz:acousticbrainz@{{.Address}}:{{.Port}}/acousticbrainz"
+POSTGRES_ADMIN_URI="postgresql://postgres@{{.Address}}:{{.Port}}/template1"
+POSTGRES_ADMIN_AB_URI="postgresql://postgres@{{.Address}}:{{.Port}}/acousticbrainz"
+{{end}}
+{{end}}
+
+{{if service "acousticbrainz-redis"}}
```
mayhem: AB's use of redis is not very great -- let's not duplicate this and use the main metabrainz redis since that is on track for no longer being a single point of failure.

webserver/__init__.py
```diff
@@ -20,24 +31,30 @@ def create_app_with_configuration(config_path=None):
...
-    app.config.from_pyfile(os.path.join(
-        root_path,
-        'default_config.py'
-    ))
-    app.config.from_pyfile(os.path.join(
-        root_path,
-        'custom_config.py'
-    ), silent=True)
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "config.py")
+    print(config_file)
```
mayhem: Not sure if we need this now since there is only one config file.

mayhem: For the MeB setup, I actually generate the consul config to consul_config.py and refuse to even look at config.py should it miraculously appear in the repo. (I've also added config.py to .dockerignore) This makes it extra difficult to get the wrong config file... Perhaps even impossible.

webserver/__init__.py
```diff
@@ -47,7 +64,7 @@ def create_app(debug=None, config_path=None):
 
     # Logging
     from webserver.loggers import init_loggers
-    init_loggers(app)
+    init_loggers(app) # TODO: use CustomFlask.init_loggers here instead
```
mayhem: At some point we should add support for setting the log level via the config file.

consul_config.py.ctmpl
```diff
@@ -0,0 +1,51 @@
...
+MUSICBRAINZ_USERAGENT = '''{{template "KEY" "musicbrainz_useragent"}}'''
+MUSICBRAINZ_HOSTNAME = '''{{template "KEY" "musicbrainz_hostname"}}'''
+
+MUSICBRAINZ_CLIENT_ID = '''{{template "KEY" "musicbrainz/client_id"}}'''
+MUSICBRAINZ_CLIENT_SECRET = '''{{template "KEY" "musicbrainz/client_secret"}}'''
+
+# Set to True if Less should be compiled in browser. Set to False if styling is pre-compiled.
+COMPILE_LESS = False
+
+LOG_FILE_ENABLED = True
```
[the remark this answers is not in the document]
mayhem: That could also be cargo culting. :)

webserver/__init__.py
```diff
@@ -20,24 +31,30 @@ def create_app_with_configuration(config_path=None):
...
-    ), silent=True)
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "config.py")
+    print(config_file)
+    if deploy_env and not config_path:
+        print("Checking if consul generated config file exists: $s" % config_file)
+        for _ in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
+            if not os.path.exists(config_file):
+                time.sleep(1)
+
+        if not os.path.exists(config_file):
```
alastair: as I understand, this loop is to wait for the config file to be created by consul, right? Out of interest, is there a reason why we're not using something like dockerize to wait for the file before launching flask? (https://github.com/jwilder/dockerize#waiting-for-other-dependencies)
The advantage of this is that we don't have this weird deploy-level logic inside a configuration/setup method

webserver/__init__.py
```diff
@@ -20,24 +31,30 @@ def create_app_with_configuration(config_path=None):
...
-    root_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
-    app.config.from_pyfile(os.path.join(
-        root_path,
-        'default_config.py'
-    ))
-    app.config.from_pyfile(os.path.join(
-        root_path,
-        'custom_config.py'
-    ), silent=True)
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "config.py")
```
alastair: what's the config file in production, and how is it loaded? I don't see any `app.config.from_*` that loads consul_config.py, which I understand is the name of the production config file.

alastair: got it, the consul config file reads from `consul_config.py.ctmpl` and writes to `config.py`

consul_config.py.ctmpl
paramsingh: By metabrainz redis, do you mean this one: https://github.com/metabrainz/docker-server-configs/blob/master/scripts/nodes/hip.sh#L10

or one of the many started by musicbrainz on williams: https://github.com/metabrainz/docker-server-configs/blob/master/scripts/nodes/williams.sh#L18

mayhem: LOL. The same one that LB uses, should be fine.

webserver/__init__.py
paramsingh: [AB-355](https://tickets.metabrainz.org/browse/AB-355)

README.md
```diff
@@ -13,7 +13,7 @@ Please report issues here: http://tickets.musicbrainz.org/browse/AB
 You can use [docker](https://www.docker.com/) and [docker-compose](https://docs.docker.com/compose/) to run the AcousticBrainz server. Make sure docker and docker-compose are installed.
 Then copy two config files:
 
-1. `custom_config.py.example` to `custom_config.py` *(you don't need to modify this file)*
+1. `config.py.example` to `config.py` *(you don't need to modify this file)*
```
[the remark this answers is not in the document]
paramsingh: Well, to be honest, we had to modify it earlier also. I'll fix it.

consul_config.py.ctmpl
```diff
@@ -0,0 +1,51 @@
...
+
+{{if service "acousticbrainz-pgbouncer"}}
+{{with index (service "pgbouncer-master") 0}}
+SQLALCHEMY_DATABASE_URI = "postgresql://acousticbrainz:acousticbrainz@{{.Address}}:{{.Port}}/acousticbrainz"
+POSTGRES_ADMIN_URI="postgresql://postgres@{{.Address}}:{{.Port}}/template1"
+POSTGRES_ADMIN_AB_URI="postgresql://postgres@{{.Address}}:{{.Port}}/acousticbrainz"
+{{end}}
+{{end}}
+
+{{if service "listenbrainz-redis"}}
```
paramsingh: should probably consider renaming this then :P

webserver/__init__.py
mayhem: Upon sleeping on it, I prefer to keep this as python code, complete with access for easy Sentry reporting in case shit goes wrong. Dockerize is good for a few things, namely ensuring the correct start-up order of containers. But I feel that this situation needs more nuance and more reporting.

consul_config.py.ctmpl
mayhem: Ah, yes. Well, it used to be standalone until we moved to production. We should've renamed it then...

docker/consul-template.conf
```diff
@@ -0,0 +1,5 @@
+template {
+    source = "/code/acousticbrainz/consul_config.py.ctmpl"
+    destination = "/code/acousticbrainz/consul_config.py"
```
mayhem: 👍

webserver/__init__.py
```diff
@@ -20,24 +31,35 @@ def create_app_with_configuration(config_path=None):
...
+
+    # if deploying in a production docker environment, load consul_config.py
+    if deploy_env and not config_path:
+        print("Checking if consul generated config file exists: $s" % config_file)
+        consul_config = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "consul_config.py")
+        for _ in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
+            if not os.path.exists(config_file):
+                time.sleep(1)
+
+        if not os.path.exists(config_file):
```
mayhem: This should be consul_config, no?

webserver/__init__.py
```diff
@@ -20,24 +31,35 @@ def create_app_with_configuration(config_path=None):
...
+            print("No config file generated. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT)
+            sys.exit(-1)
+
+        app.config.from_pyfile(consul_config)
+        return app
+
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "config.py")
```
mayhem: I think for completeness sake we should add another print statement logging which config file was read as part of this process.

paramsingh:
don't merge yet, I'll do it, there's something going wrong still.

mayhem:
These things need to be carefully merged, so I try to leave them to you to merge at the right time.

webserver/__init__.py
```diff
@@ -20,24 +31,37 @@ def create_app_with_configuration(config_path=None):
...
+        app.config.from_pyfile(consul_config)
+        return app
+
+    print('Loading config.py...')
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "config.py")
+    app.config.from_pyfile(config_file)
 
     if config_path:
-        app.config.from_pyfile(config_path)
+        app.config.from_pyfile(config_path, silent=True)
```
alastair: This has some awkward behaviour which is causing me some problems when running tests:

* the `config_path` argument was to specify a separate config file for testing. If we no longer call this method with the parameter then we should remove it
* In ListenBrainz you rename `consul_config.py.tmpl` to `config.py` (https://github.com/metabrainz/listenbrainz-server/blob/8d12052fc15c6a0a9349b51a9f53e7f1020b4485/docker/consul-template.conf#L3), but in AcousticBrainz you rename it to `consul_config.py`, which is inconsistent
* config.py is loaded unconditionally, which will probably cause a server startup error on production, as the config file is consul_config.py
* If config.py exists, this will overwrite the production config
* Unconditionally prints unneeded messages, especially during development and testing. In tests, I see the entire config printed to screen every time I create an app (each test), instead of the normal pytest `.....F..E.`
* It tires to pretty-print the config (for readability?), but it doesn't work - when running tests I see the config all on one line: `<Config {'JSON_AS_ASCII': True, 'FEATURE_EVAL_LOCATION': False, 'SESSION_REFRESH_EACH_REQUEST': True, 'DATASET_DIR': '/data/datasets', 'SENTRY_DSN': '' ... [lots more chopped]...>`

alastair: Action points:
 - [ ] remove `config_path` argument
 - [ ] standardaise on consul_config.py _or_ config.py in both apps
 - [ ] only load config.py during development
 - [ ] only use `print` in production
 - [ ] correctly pretty-print config in production

paramsingh: Commenting for posterity: https://github.com/metabrainz/acousticbrainz-server/pull/294, was reading this PR and this conversation looked unresolved.

PR #286, opened 2018-07-14
Add production dockerfiles, uwsgi config and push.sh script.

paramsingh:
This is based off of #284 because it needed the `consul-template.conf` file added there.


docker/prod/uwsgi/uwsgi.service
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
mayhem: I think it would be good to also setup the beta env here. I know it is extra work, but I really wish I would've done that in the first place with LB. This way we can initially deploy to hetzner with beta env and then later promote it to prod -- during this process we shake out the issues of running two setups on the same servers.

Dockerfile.gaia.prod
```diff
@@ -0,0 +1,87 @@
...
+                       libtag1-dev \
+                       libqt4-dev \
+                       swig2.0 \
+                       pkg-config \
+                       libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# PostgreSQL client
+RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8
+ENV PG_MAJOR 9.5
```
mayhem: Does it make sense to upgrade to a more recent version of postgres while we're doing all of this? A reload of the data is a big deal, might as well make it count, no?

mayhem: As per our discussion in IRC with @alastair, a move to PG 10 would be good.

docker/prod/uwsgi/uwsgi.ini
```diff
@@ -0,0 +1,11 @@
+[uwsgi]
+uid = www-data
+gid = www-data
+master = true
+socket = 0.0.0.0:3031
+module = acousticbrainz.server
```
alastair: this doesn't look right

PR #289, opened 2018-07-25
Add consul template for hl_extractor profile

paramsingh:
As it says on the tin.


paramsingh:
Sister commit here: https://github.com/metabrainz/docker-server-configs/pull/46/commits/95ac63b8060c163463278ce0cce7d2b8546bd853

@alastair using `v2.1_beta1` in production, same value as the value in spike.

PR #290, opened 2018-07-26
Add a downtime status page and view

paramsingh:
For the future migration to hetzner, I've added a configurable downtime page we can use in the future too.


webserver/views/index.py
```diff
@@ -62,3 +62,7 @@ def goals():
 @index_bp.route("/faq")
 def faq():
     return render_template("index/faq.html")
+
+@index_bp.route("/down")
```
alastair: Maybe it'd be better to do this directly with nginx. You can use something like `try_files` (http://nginx.org/en/docs/http/ngx_http_core_module.html#try_files) to see if a file exists (is_down.html), and serve that, otherwise forward to the uwsgi server. We use this workflow in some other projects to put the app into maintenance mode while we upgrade/restart the app server.

paramsingh:
@brainzbot retest this please.

alastair:
The html file can't be called is_down.html, otherwise nginx will always see that AB is down. Instead, we should call it is_down.html.disabled, or similar.

I'd like to see some instructions somewhere on how to take this up/down. It could be a management command, or a docker exec command to rename the file.

paramsingh:
I guess after the discussion on IRC, the way forward is to merge this, deploy during downtime on spike, then use normal hetzner procedures to keep AB down after DNS switch? Then remove the is_down file after migration.

PR #238, opened 2017-10-22
AB-117: Add functions for creating dataset dumps

ferbncode:
Added functions `dump_dataset_tables`  and `import_datasets_dumps` for exporting and importing dataset dumps. 

To create dumps, one can use:
`python manage.py dump full_dataset_dump`
To import dataset dump, one can use:
`python manage.py import_data --is-dataset-dump export/<dump.tar.xz>`

PR #293, opened 2018-08-18
 Limit size of files while dumping lowlevel_json 

paramsingh:
Instead of dumping the entire lowlevel_json and highlevel_model tables into a single file, I've split it into multiple files. 

This is very similar to how we do it in LB. We keep dumping more rows into a file until the size of the file becomes more than 10 GB.  

We then add these files to the archive and remove them from temporary storage.

### DIFF
--- a/db/dump.py
+++ b/db/dump.py
@@ -119,7 +119,7 @@ _DATASET_TABLES = {
         "snapshot_id",
         "status",
         "status_msg",
-        "options"
+        "options",
         "training_snapshot",
         "testing_snapshot",
         "created",
@@ -200,8 +200,51 @@ def _copy_table(cursor, location, table_name, query):
         cursor.copy_to(f, query)
 
 
-def _copy_tables(location, dataset_dump, start_time=None, end_time=None):
-    """Copies all tables into separate files within a specified location (directory).
+def _copy_dataset_tables(location, start_time=None, end_time=None):
+    """ Copies datasets tables into seperate files withing a specified location (directory).
+    """
+    connection = db.engine.raw_connection()
+    try:
+        cursor = connection.cursor()
+        # dataset
+        _copy_table(cursor, location, "dataset", "(SELECT %s FROM dataset)" %
+                    (", ".join(_DATASET_TABLES["dataset"])))
+
+        # dataset_class
+        _copy_table(cursor, location, "dataset_class", "(SELECT %s FROM dataset_class)" %
+                    (", ".join(_DATASET_TABLES["dataset_class"])))
+
+        # dataset_class_member
+        _copy_table(cursor, location, "dataset_class_member", "(SELECT %s FROM dataset_class_member)" %
+                    (", ".join(_DATASET_TABLES["dataset_class_member"])))
+
+        # dataset_snapshot
+        _copy_table(cursor, location, "dataset_snapshot", "(SELECT %s FROM dataset_snapshot)" %
+                    (", ".join(_DATASET_TABLES["dataset_snapshot"])))
+
+        # dataset_eval_jobs
+        _copy_table(cursor, location, "dataset_eval_jobs", "(SELECT %s FROM dataset_eval_jobs)" %
+                    (", ".join(_DATASET_TABLES["dataset_eval_jobs"])))
+
+        # dataset_eval_sets
+        _copy_table(cursor, location, "dataset_eval_sets", "(SELECT %s FROM dataset_eval_sets)" %
+                    (", ".join(_DATASET_TABLES["dataset_eval_sets"])))
+
+        # challenge
+        _copy_table(cursor, location, "challenge", "(SELECT %s FROM challenge)" %
+                    (", ".join(_DATASET_TABLES["challenge"])))
+
+        # dataset_eval_challenge
+        _copy_table(cursor, location, "dataset_eval_challenge", "(SELECT %s FROM dataset_eval_challenge)" %
+                    (", ".join(_DATASET_TABLES["dataset_eval_challenge"])))
+    finally:
+        connection.close()
+
+
+def _copy_tables(location, start_time=None, end_time=None):
+    """Copies all core tables into separate files within a specified location (directory).
+
+    NOTE: only copies tables in the variable _TABLES
 
     You can also define time frame that will be used during data selection.
     Files in a specified directory will only contain rows that have timestamps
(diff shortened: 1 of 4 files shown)


db/dump.py
```diff
@@ -112,45 +176,69 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
...
+    offset = 0
+    while True:
+        file_name = str(uuid.uuid4())
+        path = os.path.join(location, table_name, file_name[0], file_name[0:2])
+        utils.path.create_path(path)
+        with open(os.path.join(path, file_name), "a") as f:
+            logging.info(" - Copying table {table_name} to {file_name}...".format(table_name=table_name, file_name=file_name))
+            print(" - Copying table {table_name} to {file_name}...".format(table_name=table_name, file_name=file_name))
+            while True:
+                size = f.tell()
```
alastair: do you know if this operation takes longer as the files get larger? As we know that all of the low level files (and therefore rows) are of a similar size, do you think that it might be more efficient to just copy a set number of rows to each file?

paramsingh: definitely would make the code simpler. I'm not sure if it'll change much in terms of efficiency though.

db/dump.py
```diff
@@ -112,45 +175,63 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
...
+        tar: the TarFile object to which the table dumps must be added
+        archive_name: the name of the archive
+    """
+    location = tempfile.mkdtemp()
+    done = False
+    offset = 0
+    while True:
+        file_name = str(uuid.uuid4())
+        path = os.path.join(location, table_name, file_name[0], file_name[0:2])
+        utils.path.create_path(path)
```
alastair: Here, you create a folder called `table_name` and then inside it, a number of files with uuids for filenames? I don't like the uuids, because:
 * it's opaque as to what the file is if you look at just the filename part
 * you lose the ordering of these parts - if you uncompress them, the last segment could be first because of the alphabetical ordering of the filenames (it's true that they'll be in the right order in the tarfile)

could we instead name them as `abdump/table/table_n` where n is an increasing number?

db/dump.py
```diff
@@ -112,45 +175,63 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
...
+        with open(os.path.join(path, file_name), "a") as f:
+            logging.info(" - Copying table {table_name} to {file_name}...".format(table_name=table_name, file_name=file_name))
+            cursor.copy_to(f, query.format(limit=ROWS_PER_FILE, offset=offset))
+            offset += ROWS_PER_FILE
+            if f.tell() == 0:
+                done = True
```
alastair: I think this would be more explicit if it was called something like `no_more_rows` - I think `f.tell()` will be 0 when the query results in 0 rows because the offset is higher than the number of rows in the table?

db/dump.py
paramsingh: I chose the directory structure so that we didn't have too many files in the one directory. But okay, I'll change it to `abdump/table/table_n`. :+1:

db/dump.py
paramsingh: on it.

db/dump.py
```diff
@@ -264,7 +387,13 @@ def import_db_dump(archive_path):
                         logging.info("Schema version verified.")
 
                 else:
-                    if file_name in table_names:
+                    if _is_partitioned_table_dump_file(file_name):
```
alastair: If the member is the lowlevel directory then this will pass and try and import it. Maybe we can only do the check if member.is_file() is true

db/dump.py
```diff
@@ -264,7 +387,13 @@ def import_db_dump(archive_path):
...
+                        table_name = member.name.split("/")[2]
```
alastair: I think this is a little opaque and could be made more explicit. it looks like you're doing this:
 - If filename is lowlevel-1 then it's partitioned
 - member name is archivename/abdump/lowlevel/lowlevel-1, so the table name is member.split('/')[2] (lowlevel)

It'd be nice to either have a comment here explaining that partitioned tables are in a folder named by the table, and this is how we get it. This could be as simple as 
`archive, _, tablename, filename = member.name.split('/')`, so that it's explicit that the name should have 4 components, and what each part is.

db/dump.py
```diff
@@ -264,7 +387,13 @@ def import_db_dump(archive_path):
...
+                        logging.info(" - Importing data from file %s into %s table..." % (file_name, table_name))
+                        cursor.copy_from(tar.extractfile(member), '"%s"' % table_name,
```
alastair: We're streaming each file out of the archive file in order, which means we need to be absolutely sure that we apply multi-part files in the right order (lowlevel-1, lowlevel-2, lowlevel-3, etc). I think that the order when uncompressing will be the same as when we compress, but I can imagine that for some reason we uncompress/recompress a dump file and then find that the files are in some other order.
Can we have a basic assert that the suffix numbers apply in increasing order?

db/dump.py
```diff
@@ -112,45 +186,70 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
...
+        archive_name: the name of the archive
+    """
+    location = tempfile.mkdtemp()
+    offset = 0
+    file_count = 0
+    utils.path.create_path(os.path.join(location, table_name))
+    while True:
+        more_rows_added = False
+        file_count += 1
+        file_name = '{table_name}-{file_number}'.format(table_name=table_name, file_number=file_count)
```
alastair: 0 padding

manage.py
```diff
@@ -102,6 +102,17 @@ def import_data(archive):
...
     db.dump.import_db_dump(archive)
+    print('Done!')
+
+
+@cli.command()
+@click.argument("archive", type=click.Path(exists=True))
+def import_dataset_data(archive):
+    """Imports dataset dump into the database."""
+
+    print('Importing dataset data...')
```
alastair: print/logging?

db/dump.py
```diff
@@ -555,3 +684,101 @@ def _get_incremental_dump_timestamp(dump_id=None):
...
+    pxz_command = ["pxz", "--decompress", "--stdout", archive_path]
+    pxz = subprocess.Popen(pxz_command, stdout=subprocess.PIPE)
+
+    table_names = _DATASET_TABLES.keys()
+    connection = db.engine.raw_connection()
+    try:
+        cursor = connection.cursor()
+        with tarfile.open(fileobj=pxz.stdout, mode="r|") as tar:
+            temp_dir = tempfile.mkdtemp()
+            tar.extractall(temp_dir)
```
alastair: This is different to the data file dumps, where we stream the data out a file at a time instead of all at once. Is there a reason for the difference?

paramsingh: addressed in #238

db/dump.py
```diff
@@ -86,6 +88,78 @@
     ),
 }
 
+_DATASET_TABLES = {
```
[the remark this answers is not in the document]
paramsingh: Yes.

db/dump.py
```diff
@@ -176,67 +275,91 @@ def generate_where(row_name, start_t=start_time, end_t=end_time):
...
-        with open(os.path.join(location, "lowlevel"), "w") as f:
-            logging.info(" - Copying table lowlevel...")
-            cursor.copy_to(f, "(SELECT %s FROM lowlevel %s)" %
-                           (", ".join(_TABLES["lowlevel"]), generate_where("submitted")))
-
-        # lowlevel_json
-        with open(os.path.join(location, "lowlevel_json"), "w") as f:
-            logging.info(" - Copying table lowlevel_json...")
-            query = "SELECT %s FROM lowlevel_json WHERE id IN (SELECT id FROM lowlevel %s)" \
+        if dataset_dump:
```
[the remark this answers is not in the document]
paramsingh: addressed in #238

db/dump.py
```diff
@@ -112,45 +186,70 @@ def dump_db(location, threads=None, incremental=False, dump_id=None):
...
+        cursor: a psycopg2 cursor
+        location: the directory where the table should be copied.
+        table_name: the name of the table to be copied.
+        query: the select query for getting data from the table.
+    """
+    with open(os.path.join(location, table_name), "w") as f:
+        logging.info(" - Copying table {table_name}...".format(table_name=table_name))
+        print(" - Copying table {table_name}...".format(table_name=table_name))
+        copy_query = 'COPY ({query}) TO STDOUT'.format(query=query)
+        cursor.copy_expert(copy_query, f)
```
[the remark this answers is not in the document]
paramsingh: I think it was undocumented behaviour. 

copy_to constructs the copy query here and it could possibly work with the queries we passed. https://github.com/psycopg/psycopg2/blob/153b0983c18cf6f5536c4d276af2b63d8a5f0568/psycopg/cursor_type.c#L1524

paramsingh:
@alastair ready for review again.

PR #294, opened 2018-08-27
Fix Config loading code

paramsingh:
Action points:
 - [x] remove `config_path` argument
 - [x] Remove `create_app_with_configuration` method, which is no longer used independently
 - [x] standardaise on consul_config.py _or_ config.py in both apps
 - [x] only use `print` in production
 - [x] correctly pretty-print config in production

### DIFF
--- a/webserver/__init__.py
+++ b/webserver/__init__.py
@@ -29,8 +29,8 @@ def load_config(app):
             if not os.path.exists(config_file):
                 time.sleep(1)
 
-    if not os.path.exists(config_file):
-        print("No config file generated. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT)
+        if not os.path.exists(config_file):
+            print("No config file generated. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT)
 
     app.config.from_pyfile(config_file)
 
--- a/webserver/__init__.py
+++ b/webserver/__init__.py
@@ -3,9 +3,9 @@ from __future__ import print_function
 from brainzutils.flask import CustomFlask
 from flask import request, url_for, redirect
 from flask_login import current_user
+from pprint import pprint
 
 import os
-import pprint
 import sys
 import time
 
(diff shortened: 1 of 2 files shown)


webserver/__init__.py
```diff
@@ -22,46 +22,32 @@ def create_app_flaskgroup(script_info):
...
-    if config_path:
-        app.config.from_pyfile(config_path, silent=True)
-
-    return app
+    if deploy_env:
+        print('Config file loaded!')
+        pprint(dict(app.config))
 
 
 def create_app(debug=None, config_path=None):
```
alastair: I don't think we have any code that calls `create_app(config_path=...)`. Can we remove this argument?

webserver/__init__.py
```diff
@@ -22,46 +22,32 @@ def create_app_flaskgroup(script_info):
...
+    for _ in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
+        if not os.path.exists(config_file):
+            time.sleep(1)
 
-        print('Loading consul_config.py...')
-        app.config.from_pyfile(consul_config)
-        return app
+    if not os.path.exists(config_file):
+        print("No config file generated. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT)
+        sys.exit(-1)
```
alastair: `app.config.from_pyfile` will fail anyway if the file doesn't exist. Does that make the `sys.exit` unneeded?

webserver/__init__.py
```diff
@@ -22,46 +22,32 @@ def create_app_flaskgroup(script_info):
...
-            if not os.path.exists(consul_config):
-                time.sleep(1)
-
-        if not os.path.exists(consul_config):
-            print("No config file generated. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT)
-            sys.exit(-1)
+def load_config(app):
+    """Load configuration file for specified Flask app"""
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "config.py")
+    for _ in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
```
alastair: in development I'd like it much more to have a failure message immediately if I forget to copy config.py.example to config.py rather than have to wait 10 seconds for the message to appear. Is there a way that we can support both of these cases?

webserver/__init__.py
```diff
@@ -25,13 +24,13 @@ def create_app_flaskgroup(script_info):
...
-    for _ in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
-        if not os.path.exists(config_file):
-            time.sleep(1)
+    if deploy_env:
+        for _ in range(CONSUL_CONFIG_FILE_RETRY_COUNT):
+            if not os.path.exists(config_file):
+                time.sleep(1)
 
     if not os.path.exists(config_file):
         print("No config file generated. Retried %d times, exiting." % CONSUL_CONFIG_FILE_RETRY_COUNT)
```
alastair: This sentence is not strictly true in development any more. Do we care?

alastair: Perhaps we could fix it by indenting one level?

paramsingh: on it.

PR #296, opened 2018-09-05
AB-93: Update sequences in the db after an import of some data dump

paramsingh:
Also, add a command to manage.py to update_sequences, if needed.

### DIFF
--- a/db/dump.py
+++ b/db/dump.py
@@ -238,6 +238,48 @@ def _copy_tables(location, start_time=None, end_time=None):
         connection.close()
 
 
+def update_sequence(seq_name, table_name):
+    """ Update the specified sequence's value to the maximum value of ID in the table.
+
+    Args:
+        seq_name (str): the name of the sequence to be updated.
+        table_name (str): the name of the table from which the maximum value is to be retrieved
+    """
+    with db.engine.connect() as connection:
+        connection.execute(text("""
+            SELECT setval('{seq_name}', max(id))
+              FROM {table_name}
+        """.format(seq_name=seq_name, table_name=table_name)))
+
+
+def update_sequences():
+    """ Update all sequences to the maximum value of id in the table.
+    """
+    # lowlevel_id_seq
+    update_sequence('lowlevel_id_seq', 'lowlevel')
+
+    # highlevel_model_id_seq
+    update_sequence('highlevel_model_id_seq', 'highlevel')
+
+    # version_id_seq
+    update_sequence('version_id_seq', 'version')
+
+    # model_id_seq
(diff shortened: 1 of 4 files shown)


db/dump.py
```diff
@@ -238,6 +241,56 @@ def generate_where(row_name, start_t=start_time, end_t=end_time):
...
+            SELECT setval('{seq_name}', max(id))
+              FROM {table_name}
+        """.format(seq_name=seq_name, table_name=table_name)))
+
+
+def update_sequences():
+    """ Update all sequences to the maximum value of id in the table.
+    """
+    # lowlevel_id_seq
+    logging.info('Updating lowlevel_id_seq...')
```
alastair: I've created https://tickets.metabrainz.org/projects/AB/issues/AB-364 to add some docs to remind us that we have to update this every time we make a new table.

db/dump.py
```diff
@@ -238,6 +241,56 @@ def generate_where(row_name, start_t=start_time, end_t=end_time):
...
+    logging.info('Updating user_id_seq...')
+    update_sequence('user_id_seq', '"user"')
+
+    # dataset_class_id_seq
+    logging.info('Updating dataset_class_id_seq...')
+    update_sequence('dataset_class_id_seq', 'dataset_class')
+
+    # dataset_eval_sets_id_seq
+    logging.info('Updating dataset_eval_sets_id_seq...')
+    update_sequence('dataset_eval_sets_id_seq', 'dataset_eval_sets')
```
alastair: Do we have a way of dumping production/private tables yet (e.g. "user"?). If we eventually add it then we'll need to add those tables somewhere too - here or a new method?

db/dump.py
```diff
@@ -20,6 +20,9 @@
 import os
 from sqlalchemy import text
 
+
+logging.basicConfig(level=logging.INFO)
```
alastair: can we use the flask logger here? (I think the management tasks will be running inside an app context, right?)

db/dump.py
paramsingh: We don't have a private dump yet, but I updated those sequences in this function too.

db/dump.py
```diff
@@ -273,6 +327,11 @@ def import_db_dump(archive_path):
         connection.close()
 
     pxz.stdout.close()
+    pxz.wait()
+
+    logging.info('Updating sequences...')
```
paramsingh: leaving this rn for consistency, I'll open a seperate PR for converting all these logging calls to use the flask logger.

paramsingh: [AB-366](https://tickets.metabrainz.org/browse/AB-366)

PR #303, opened 2018-10-12
Add working directory as volume to all docker containers

alastair:
This is so that changes in the code can be reflected immediately in the containers without having to rebuild them

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -65,7 +65,7 @@ RUN git clone https://github.com/MTG/essentia /tmp/essentia \
 # SVM models
 RUN mkdir /tmp/models \
     && cd /tmp/models \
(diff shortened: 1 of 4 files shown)


alastair:
In that case we should put the compiled tools and data files somewhere else, and keep /code for code. I'll look into it.

alastair:
Now we install the hl extractor to /usr/local/bin to not conflict with the volume mount. @paramsingh does this look good?

paramsingh:
@brainzbot retest this please

paramsingh:
seems like the model file on essentia.upf.edu is buggy?

```
☁  ~   curl --silent -o models.tar.gz http://essentia.upf.edu/documentation/svm_models/essentia-extractor-svm_models-v2.1_beta1.tar.gz
☁  ~  tar -xvzf models.tar.gz

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
☁  ~  
```

alastair:
gah, I made the change but didn't commit it: we recently added https to the essentia URL + a redirect so we need curl -L, fixing it now

PR #309, opened 2018-11-12
Fix uwsgi.service to start cron in the web container

paramsingh:
We explicitly stop cron in the Dockerfile so that it doesn't run in all the containers.

But I missed starting it up in the web container

### DIFF
--- a/docker/prod/uwsgi/uwsgi.service
+++ b/docker/prod/uwsgi/uwsgi.service
@@ -2,6 +2,9 @@
 
 if [ "${CONTAINER_NAME}" = "acousticbrainz-web-prod" ]
 then
+    rm -f /etc/service/cron/down
+    sv restart cron
+
     exec uwsgi /etc/uwsgi/uwsgi.ini
 fi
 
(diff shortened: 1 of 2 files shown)


paramsingh:
@brainzbot retest this please.

paramsingh:
@brainzbot retest this please

alastair:
Looking through the AB `Dockerfile.prod` file I have a few further followups:

* We install `cron` when installing the dependencies (line16), this probably isn't needed as it's already installed by the baseimage?
* Where is the `down` file for cron added? I've looked in `metabrainz/python:2.7` and it's not there, and I can't see anything in the Dockerfile that adds it, so...?
* Where is the cron configuration added? I don't see anything in the Dockerfile to add this

alastair:
@paramsingh with the changes to #319 we probably need to make a few changes here regarding the way that the Dockerfile is set up. What are your thoughts? Should we get a quick release out with these changes, or should we attack this issue with service names and configuration files? I'd rather that we fix this properly now, should we plan to fix this together some time this week?

alastair:
Actually, I'm confused by this now.. I think this PR has lost a few commit, I guess someone merged the updates to Dockerfile by mistake? I only see a small fix here now to run cron as a second container. Is cron already running in production in the web container?

paramsingh:
Closing in favor of #329

PR #311, opened 2018-12-20
AB-376: Add all stats rows in a single transaction

alastair:
# Issue
We had an issue where a compute_stats run must have skipped writing a single row, as a result, the stats page couldn't find the value for this date.

# Fixes 
Add all rows in a single transaction so that if there is any problem writing any one of them, we're not left in an inconsistent state

Perform an additional sanity check after computing stats to ensure that we're not left in this state any more

If a value is missing when rendering the stats page, ignore it instead of raising an error

Only return a stats value for each day to reduce the size of the returned js file 24x (5MB->200K), and add some indexes to make this process faster.

# deployment steps
This PR adds a new SQL update file which creates some indexes, `admin/updates/20190227-statistics-indexes.sql`, this should be run after deployment.

### DIFF
--- a/admin/sql/create_indexes.sql
+++ b/admin/sql/create_indexes.sql
@@ -23,4 +23,7 @@ CREATE INDEX highlevel_ndx_highlevel_model ON highlevel_model (highlevel);
 
 CREATE UNIQUE INDEX lower_musicbrainz_id_ndx_user ON "user" (lower(musicbrainz_id));
 
+CREATE INDEX collected_ndx_statistics ON statistics (collected);
+CREATE INDEX collected_hour_ndx_statistics ON statistics (date_part('hour'::text, timezone('UTC'::text, collected)));
+
 COMMIT;
--- a/admin/updates/20190227-statistics-indexes.sql
+++ b/admin/updates/20190227-statistics-indexes.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+CREATE INDEX collected_ndx_statistics ON statistics (collected);
+CREATE INDEX collected_hour_ndx_statistics ON statistics (date_part('hour'::text, timezone('UTC'::text, collected)));
+
+COMMIT;
--- a/db/stats.py
+++ b/db/stats.py
@@ -100,18 +100,18 @@ def compute_stats(to_date):
 
         while next_date < to_date:
             stats = _count_submissions_to_date(connection, next_date)
-            with connection.begin() as transaction:
-                _write_stats(connection, next_date, stats)
+            _write_stats(connection, next_date, stats)
             next_date = _get_next_hour(next_date)
 
 
 def _write_stats(connection, date, stats):
     """Records a value with a given name and current timestamp."""
-    for name, value in six.iteritems(stats):
-        q = text("""
-            INSERT INTO statistics (collected, name, value)
-                 VALUES (:collected, :name, :value)""")
-        connection.execute(q, {"collected": date, "name": name, "value": value})
+    with connection.begin():
+        for name, value in six.iteritems(stats):
+            q = text("""
+                INSERT INTO statistics (collected, name, value)
+                     VALUES (:collected, :name, :value)""")
+            connection.execute(q, {"collected": date, "name": name, "value": value})
 
 
 def add_stats_to_cache():
--- a/db/stats.py
+++ b/db/stats.py
@@ -196,12 +196,16 @@ def load_statistics_data(limit=None):
     # create an array of {"name": name, "value": value} objects and change
     # it in python
     args = {}
-    # TODO: use sqlalchemy select().limit()?
     qtext = """
             SELECT collected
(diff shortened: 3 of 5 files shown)


alastair:
yeah, I wondered about tests, I'm really not sure what the best way to do that would be. Your recommendation is probably the best option

alastair:
I'm in the middle of some more changes for this, stay tuned

alastair:
@paramsingh changes done, take a look and see if you have any comments

I'm still not happy with the stats output for highcharts - it takes quite a while to run on my development machine with the full stats database (about 600ms), however the version without indexes ran in 15s so I think it's significantly better. We should run some analyze queries in production to verify if this is better there.
We could also consider a proportional stats selection, where for older items we only select weekly/monthly, and for more recent ones we select daily. Perhaps there's a way that we can tell highcharts to select more detailed information if a user zooms in?
We're also spending a lot of space because the unix timestamps are in ms resolution. I'd like to experiment with dividing all values by 10000 and re-constituting them on the client to see if it's worth saving transfer space over. I'm not sure if this is a useful tradeoff to do to save just a few bytes.

alastair:
We should try these analyze queries before and after deploying:

Current stats query (run before):
```
explain analyze             SELECT collected
                 , json_agg(row_to_json(
                    (SELECT r FROM (SELECT name, value) r) )) AS stats
              FROM statistics
          GROUP BY collected
          ORDER BY collected DESC
```

Updated query (run after indexes are created):
```
            explain analyze SELECT collected
                 , json_agg(row_to_json(
                    (SELECT r FROM (SELECT name, value) r) )) AS stats
              FROM statistics
             WHERE date_part('hour', timezone('UTC'::text, collected)) = 0 
                OR collected = (SELECT collected
                                  FROM statistics
                              ORDER BY collected DESC
                                 LIMIT 1)
          GROUP BY collected
          ORDER BY collected DESC
```

At the very least, we're hoping that the second query is no slower than the first. If it's slower, we'll have to look into how to reduce the query time in a different way.

alastair:
Current stats query:
```
QUERY PLAN                                                           
-------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=23703.64..28592.95 rows=36898 width=40) (actual time=54.272..274.394 rows=36876 loops=1)
   Group Key: statistics.collected
   ->  Sort  (cost=23703.64..24257.15 rows=221404 width=31) (actual time=54.246..62.797 rows=221256 loops=1)
         Sort Key: statistics.collected DESC
         Sort Method: quicksort  Memory: 23430kB
         ->  Seq Scan on statistics  (cost=0.00..4047.04 rows=221404 width=31) (actual time=0.028..23.236 rows=221256 loops=1)
   SubPlan 1
     ->  Result  (cost=0.00..0.01 rows=1 width=32) (actual time=0.000..0.000 rows=1 loops=221256)
 Planning time: 0.591 ms
 Execution time: 276.249 ms
(10 rows)
```

alastair:
The updated query is fast even without indexes built:
```
                                                                    QUERY PLAN                                                                     
---------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=11471.46..11507.46 rows=1099 width=40) (actual time=74.251..84.712 rows=1538 loops=1)
   Group Key: statistics.collected
   InitPlan 2 (returns $2)
     ->  Limit  (cost=5154.06..5154.06 rows=1 width=8) (actual time=36.274..36.275 rows=1 loops=1)
           ->  Sort  (cost=5154.06..5707.57 rows=221404 width=8) (actual time=36.273..36.273 rows=1 loops=1)
                 Sort Key: statistics_1.collected DESC
                 Sort Method: top-N heapsort  Memory: 25kB
                 ->  Seq Scan on statistics statistics_1  (cost=0.00..4047.04 rows=221404 width=8) (actual time=0.002..19.292 rows=221256 loops=1)
   ->  Sort  (cost=6317.40..6320.18 rows=1113 width=31) (actual time=74.228..74.605 rows=9228 loops=1)
         Sort Key: statistics.collected DESC
         Sort Method: quicksort  Memory: 1105kB
         ->  Seq Scan on statistics  (cost=0.00..6261.08 rows=1113 width=31) (actual time=36.295..72.917 rows=9228 loops=1)
               Filter: ((date_part('hour'::text, timezone('UTC'::text, collected)) = '0'::double precision) OR (collected = $2))
               Rows Removed by Filter: 212028
   SubPlan 1
     ->  Result  (cost=0.00..0.01 rows=1 width=32) (actual time=0.000..0.000 rows=1 loops=9228)
 Planning time: 4.815 ms
 Execution time: 84.783 ms
(18 rows)
```
It looks like this is because the expensive part of the operation is generating the json. By only selecting 1 item in every 24 we are generating significantly less data. This means that creating the indexes should make the filter even faster, but we should continue to decide if there is a better way of storing the data (https://tickets.metabrainz.org/browse/AB-385)

PR #312, opened 2019-01-03
AB-318: Return HTTP401 on API endpoints if the user is not logged in

pulkit6559:
Added a new decorator @api_login_required in webserver/decorators.py for api endpoints which return JSON

### DIFF
--- a/webserver/decorators.py
+++ b/webserver/decorators.py
@@ -19,7 +19,7 @@ def api_login_required(f):
     @wraps(f)
     def decorated(*args, **kwargs):
         if current_user.is_authenticated:
-            return f(*args, **kwargs):
+            return f(*args, **kwargs)
         else: 
             return jsonify(message = "Request is not authenticated") , 401
     return decorated
--- a/webserver/decorators.py
+++ b/webserver/decorators.py
@@ -1,6 +1,6 @@
 from functools import update_wrapper, wraps
 from datetime import timedelta
-from flask import request, current_app, make_response
+from flask import request, current_app, make_response, jsonify
 from flask_login import current_user
 from werkzeug.exceptions import Unauthorized
 from six import string_types
--- a/webserver/decorators.py
+++ b/webserver/decorators.py
@@ -15,6 +15,14 @@ def auth_required(f):
             raise Unauthorized
     return decorated
 
+def api_login_required(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if current_user.is_authenticated:
+            return f(*args, **kwargs):
+        else: 
+            return jsonify(message = "Request is not authenticated") , 401
+    return decorated
 
 def crossdomain(origin='*', methods=None, headers=None,
                 max_age=21600, attach_to_all=True,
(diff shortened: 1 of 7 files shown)


paramsingh:
@brainzbot add to whitelist

pulkit6559:
Thanks a lot for the feedback @paramsingh :)
I'll make the required changes and make sure to run tests before submitting my next PR.

paramsingh:
No worries at all. :)

alastair:
@brainzbot retest this please

paramsingh:
@brainzbot retest this please

webserver/decorators.py
```diff
@@ -15,6 +15,14 @@ def decorated(*args, **kwargs):
             raise Unauthorized
     return decorated
 
+def api_login_required(f):
```
alastair: We actually already have an `auth_required` decorator above which is used for API endpoints. These two methods are exactly the same. We can reuse that one instead of making an new one

alastair: We actually already have an `auth_required` decorator above which is used for API endpoints. These two methods are exactly the same. We can reuse that one instead of making an new one

webserver/views/datasets.py
```diff
@@ -202,7 +203,7 @@ def view_json(id):
 
 
 @datasets_bp.route("/create", methods=("GET", "POST"))
-@login_required
+@api_login_required
 def create():
```
alastair: This view does two things - on GET it returns HTML. This means that we should do a redirect if the user isn't logged in, but for a POST (sent by the react app) we should return json in the message if the user isn't logged in. This is difficult to do in a single method, so we should split this method into two, one to handle the GET and one for the POST

webserver/views/datasets.py
```diff
@@ -230,7 +231,7 @@ def create():
 
 
 @datasets_bp.route("/import", methods=("GET", "POST"))
-@login_required
+@api_login_required
```
alastair: You need to check the functionality of these methods before making changes like this - this method only renders HTML pages, even during POST, therefore we should use `@login_required` here

webserver/views/datasets.py
```diff
@@ -304,7 +305,7 @@ def edit(dataset_id):
 
 
 @datasets_bp.route("/<uuid:dataset_id>/delete", methods=("GET", "POST"))
-@login_required
+@api_login_required
```
alastair: same as `import_csv` here, this page only returns HTML (even though there's a jsonify here it's a bug)

alastair:
@paramsingh can you take a look at this as well, I've tested the dataset editor with the changes that I made, but would appreciate a second set of 👀

paramsingh:
Should we make the [edit](https://github.com/metabrainz/acousticbrainz-server/pull/312/files#diff-b1772f3502d312f3d84deae191687459R275) page take get requests only too?

Other than that, looks good to me. :+1:

alastair:
Thanks @pulkit6559 for your contribution! 😺

PR #308, opened 2018-11-06
Update Flask to the latest version

TheFaR8:
Flask is keep making changes each version and in order not to break some things with outdated material, it is updated to its latest version

PR #313, opened 2019-01-06
Improve runserver command to work with reloader and debugger

alastair:
The runserver command was broken by a flask upgrade in #308, causing
it to lose the app context when the reloader restarted the app.
Copy the flask.cli.run_command command to our manage.py so that it acts
like `flask run`, including the debugger and reloader behaviour.
Ensure that we also reload on changes to config.RELOAD_ON_FILES

Fixes AB-327

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -6,9 +6,8 @@ import sys
 
 import click
 from brainzutils import cache
-from flask import current_app
-from werkzeug.serving import run_simple
-from flask.cli import FlaskGroup, shell_command
+import flask.cli
+from flask.cli import FlaskGroup
 from shutil import copyfile
 
 import db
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -18,27 +18,49 @@
...
+    if debugger is None:
+        debugger = debug
+
+    eager_loading = not reload
+
+    flask.cli.show_server_banner(flask.helpers.get_env(), debug, info.app_import_path, eager_loading)
+    app = flask.cli.DispatchingApp(info.load_app, use_eager_loading=eager_loading)
+    reload_on_files = info.load_app().config['RELOAD_ON_FILES']
+
+    from werkzeug.serving import run_simple
```
paramsingh: Why are we importing this here?

alastair: because I copied this directly from the flask cli: https://github.com/pallets/flask/blob/a74864ec229141784374f1998324d2cbac837295/flask/cli.py#L782
Perhaps I should add a comment to this method saying where it comes from?

alastair: Ah, I already do in the docstring:
> This is a copy of flask.cli.run_command

perhaps it can be more explicit with a github link

PR #316, opened 2019-01-17
Explicitly specify the name of the container running tests in Jenkins

paramsingh:


paramsingh:
@brainzbot retest this please

docker/jenkins-test.sh
```diff
@@ -46,14 +46,14 @@ function run_tests {
 
     # create database and user once
     echo "creating user and database"
-    docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME run --rm $TEST_CONTAINER_NAME \
+    docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME run --rm $TEST_CONTAINER_NAME --name $TEST_CONTAINER_REF\
```
alastair: I don't think we need `--name` here - we're already using `--rm`. This container only runs init_db and then quits and removes itself.

docker/jenkins-test.sh
```diff
@@ -46,14 +46,14 @@ function run_tests {
...
         dockerize \
         -wait tcp://db:5432 -timeout 60s bash -c \
         "cd /code && cp config.py.example config.py && ls -lR && python manage.py init_db"
 
     # run tests
     echo "running tests"
-    docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME run $TEST_CONTAINER_NAME \
+    docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME run $TEST_CONTAINER_NAME --name $TEST_CONTAINER_REF \
```
alastair: Should we add a comment saying why we explicitly set a name? (so that we have a consistent name so that we can use `docker cp` later in the script)

The only thing we have to be sure about here is that the containers gets removed properly after it runs, and that different runs of the tests don't conflict with each other. I think that we're OK here:
 * the project name contains the jenkins tag, which increases for each time that jenkins runs a job (and this is always how it's run in the past, without problems)
 * the cleanup function stops and removes all containers with the project name, so this should be removed as well with no problems

PR #318, opened 2019-01-29
Upgrade dependencies

alastair:
This PR upgrades some dependencies indicated by github as having security updates, and also does a general python version upgrade (from `pip list --outdated`). I left some outdated packages at their normal version because these are dependencies of brainzutils, and should be updated there.

Some js dependencies of gulp have issues that can only be fixed by upgrading to gulp4. Doing this as a second step as it will probably require changes to the gulpfile too.

upgrades to python packages tested manually: flask-admin (go to admin), click (run commands), ndg-httpsclient (perform oauth login), pytest (run tests)


alastair:
The new version of click has changed functions of the form `init_db` into `init-db`, causing tests to fail. Looking into it

alastair:
Thanks, merged!

PR #319, opened 2019-02-01
Use a single Dockerfile to build the development and production images

alastair:
I've encountered some times where I made a change to Dockerfile.dev and forgot to update Dockerfile.prod. @paramsingh says that he has made the same mistake in listenbrainz sometimes too.

This proposal is to use docker [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) to include both development and production setup steps in a single Dockerfile.
This means that common code (e.g. installation of external software) is shared. Given a base image of these shared steps, we can split into development and production images by using additional _targets_. In the case of the changes in this PR, all of the consul/service/cron changes will not appear in the development image. We could extend this example to include things like pytest and ipython only during development and testing.


docker/docker-compose.dev.yml
```diff
@@ -1,5 +1,5 @@
 # Docker Compose file for development
-version: "2"
+version: "3.4"
```
alastair: This is the lowest version of docker-compose file format that supports build targets. We could go higher if we only want to support recent versions of docker (this affects the version that contributors have installed [for .dev.] and whatever is installed on jenkins [for .test.]).

docker/push.sh
```diff
@@ -19,7 +19,7 @@ TAG=${2:-beta}
 
 echo "Building AcousticBrainz web image with env $ENV tag $TAG..."
 docker build -t metabrainz/acousticbrainz:$TAG \
-        -f Dockerfile.prod \
+        --target acousticbrainz-prod
```
alastair: We could use the `$ENV` here? but there's no `beta` target in the Dockerfile, so it would fail if we call this script with that as an argument

Dockerfile
```diff
@@ -91,6 +81,16 @@ RUN pip install --no-cache-dir -r requirements.txt
 COPY package.json /code
 RUN npm install
 
+
+FROM acousticbrainz-base AS acousticbrainz-dev
+
+COPY . /code
```
paramsingh: Do we even need this for dev. because the dir is a volume, and this command won't do anything afaik.

docker/push.sh
paramsingh: I'm fine with it as is, to be honest.

docker/docker-compose.dev.yml
paramsingh: We're running docker 18.06 in production.

```
param@frank:~$ docker --version
Docker version 18.06.0-ce, build 0ffa825
param@frank:~$ 
```

Jenkins has the following
```
Docker environment:
Docker version 1.12.3, build 6b644ec
docker-compose version 1.23.2, build 1110ad0
```

I'd be happy to go as up as this supports.

Dockerfile
alastair: You're right, code isn't needed in dev. I only added it because I didn't know if you could have a `FROM` line that has no other commands. In this case we'd only use it to name the target.
Eventually I'd like to add test-specific python installation in this section, but didn't do it now because it's just a demo.

docker/docker-compose.dev.yml
alastair: It looks like we'll have to upgrade docker on jenkins then (according to https://docs.docker.com/compose/compose-file/#compose-and-docker-compatibility-matrix docker 1.12 only supports compose-file 2.1)

alastair: Even though we're using docker 1.12 in jenkins, it's connecting to a 18.03 server, which is why the tests for this PR succeeded. However, we can upgrade jenkins anyway: https://github.com/metabrainz/jenkins/pull/1

It looks like we have no problem with updating this minimum version.

Dockerfile
alastair: Actually, I also use the acousticbrainz-dev target in jenkins, which doesn't have a volume. What do you think? Either we have a specific acousticbrainz-test target which COPYs the code, or we just leave this as-is.

I would prefer to use a -test or -dev target in jenkins so that we can install pytest _only_ during test, and not have it in the production image. What do you think?

alastair: I made the change to install python packages for testing only in the -dev image, and we continue to use this image with jenkins (some conflicts, but I'll fix these before merging).

Dockerfile
```diff
@@ -91,6 +81,19 @@ RUN pip install --no-cache-dir -r requirements.txt
 COPY package.json /code
 RUN npm install
 
+
+FROM acousticbrainz-base AS acousticbrainz-dev
+
+COPY requirements_development.txt /code/requirements_development.txt
```
paramsingh: Noice.

PR #321, opened 2019-02-12
AB-379: Reporting/re-running failed highlevel jobs

pulkit6559:
Added the following cli commands to manage.py:
check_highlevel : It checks for the number of submissions without highlevel meta data
clean_highlevel : It identifies the id's which do not have metadata and deletes respective rows from highlevel table


pulkit6559:
@alastair @paramsingh , do you have any feedback/suggestion for the changes made here?

alastair:
Please don't notify us before we've made a comment, especially not after only half a day - we'll get around to a review when we can, or let you know when we will.

pulkit6559:
I'll make sure not to do that next time

db/data.py
```diff
@@ -58,6 +58,44 @@ def _has_key(dictionary, key):
         dictionary = dictionary[part]
     return True
 
+def check_highlevel_meta():
```
alastair: This method name should more accurately describe what it does. Something like `get_highlevel_meta_count`. It should include a docstring that explains what it does and what the return value is.
Because this method explicitly returns a count, you should read this value from the query result and return a number, instead of having to do `h[0][0]` in the management command.
This method should also have a test to ensure that it operates as expected.

db/data.py
```diff
@@ -58,6 +58,44 @@ def _has_key(dictionary, key):
...
+    with db.engine.begin() as connection:
+        query = text("""
+                        SELECT COUNT(*)
+                        FROM highlevel_meta ;
```
alastair: there's a better way of performing this query - instead of doing one query for highlevel and one for highlevel_meta, we can do a join and just count the missing rows:
`select id from highlevel left join highlevel_meta using (id) where highlevel_meta.id is null`

Double-check the formatting of other queries in this file. We need to right-align keywords.

Queries run via sqlalchemy shouldn't have a ; at the end of them.

db/data.py
```diff
@@ -58,6 +58,44 @@ def _has_key(dictionary, key):
...
+                    """)
+        result = connection.execute(query).fetchall()
+    return result
+
+def clean_highlevel_data():
```
alastair: like the comment on `check_highlevel_meta`, think about the name of this method, a docstring, and tests.

db/data.py
```diff
@@ -58,6 +58,44 @@ def _has_key(dictionary, key):
...
+    connection = db.engine.connect()
+    trans = connection.begin()
+    try:
+        query = text("""
+                    DELETE
+                    FROM highlevel
+                    WHERE highlevel.id in
+                        (   SELECT highlevel.id
+                            FROM highlevel
+                            EXCEPT
```
alastair: I've never come across the select ... from ... except syntax before - that's not a problem, and it looks like this query works, but if we use left join in the count method, we can also use the same type of join here to do this query

manage.py
```diff
@@ -63,6 +63,28 @@ def runserver(info, host, port, debugger):
                extra_files=reload_on_files)
 
 
+@cli.command()
+def check_highlevel():
```
alastair: This is a good start, a few things here:
 * We explicitly add a `name` parameter to `cli.command` in this file. We should do the same here.
 * the name of this method should more accurately indicate what we are checking. It should include a docstring so that the automatically generated help for manage.py shows what it does.
 * I think it would be a good idea to put these commands behind a subcommand, like we do with `dump`, later in this file. This means that we can have two commands: `manage.py highlevel show_failed_rows` and `manage.py highlevel remove_failed_rows`

manage.py
```diff
@@ -63,6 +63,28 @@ def runserver(info, host, port, debugger):
...
+    try:
+       h = db.data.check_highlevel()
+       h_meta = db.data.check_highlevel_meta()
+       print("The no of submissions without highlevel metadata :",
```
alastair: I would like to see more information here. It's not very useful to only see the number of failed highlevel calculations, I would also like to know what MBIDs failed, and be able to download the lowlevel data file to experiment with it.
This means that this method should not only print the count, but the actual items themselves. This could include: 1) the highlevel.id, 2) the mbid (lowlevel.gid), 3) the offset (important if there have been more than 1 submission for an mbid so that we know which one do download)

It would be good to have this method able to show a count, and then the actual failed items when a flag is set, e.g.:

    $ manage.py highlevel failed_rows
    Number of highlevel rows that failed processing: 5

    $ manage.py highlevel failed_rows --detail
    Number of highlevel rows that failed processing: 5
    id, mbid, number
    3, 65e291fc-1bfa-4f06-a7e3-767b5951454b, 3
    ...

Note that the message describes the reason that we don't have a highlevel_meta row (failed processing), rather than just describing it. I think that it's a nice idea to make this more explicit.

manage.py
```diff
@@ -63,6 +63,28 @@ def runserver(info, host, port, debugger):
...
+             h[0][0]-h_meta[0][0])
+    except db.exceptions.DatabaseException as e:
+        click.echo("Error: %s" % e, err=True)
+        sys.exit(1)
+
+
+@cli.command()
+def clean_highlevel():
```
alastair: It would be nice to be able to optionally specify 1 or more row numbers here to remove, with an optional argument to be able to remove all at once. I think that more often we'll only want to retry one or two failed items instead of all of them.
Consider again the name of these methods. To me, "clean" sounds like we are removing all highlevel items, not just failed ones. Take care of the documentation as well to describe what this method does.

db/data.py
pulkit6559: i have removed check_highlevel and check_highlevel_meta functions and merged the functionality to one single function show_failed_rows which returns the list of failed rows and with their (id,mbid and offset)

manage.py
pulkit6559: For this, i removed the commands check_highlevel and clean_highlevel from manage.py and added them as subcommands in manage_highlevel.py file

pulkit6559:
@alastair are there any other changes to be made in this PR?

db/data.py
```diff
@@ -59,6 +59,53 @@ def _has_key(dictionary, key):
...
 
+def show_failed_rows():
+    """
+        checks for rows in highlevel table which havent been processed.
+        Returns:
+        (id, mbid, offset) of unprocessed rows sorted by submission date
+    """
+    with db.engine.begin() as connection:
+        query = text("""
+                        SELECT id,mbid, ROW_NUMBER ()
```
paramsingh: The SQL should be right aligned. :)

db/data.py
```diff
@@ -59,6 +59,53 @@ def _has_key(dictionary, key):
...
+    return result
+
+
+
+def remove_failed_rows():
+    """
+        Function for for removing rows in highlevel table which havent 
+        been processed.
+    """
+    connection = db.engine.connect()
```
paramsingh: A `with` block here would reduce a lot of logic.

```
    with connection.begin() as transaction:
        # do stuff here
```

db/manage_highlevel.py
```diff
@@ -0,0 +1,43 @@
...
+from db import dump
+import click
+import webserver
+
+cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app_flaskgroup)
+
+
+@cli.command(name='show_failed_rows')
+@click.option("--list", "-l", is_flag=True, help="Lists the details of failed rows.")
+def show_failed_rows(list):
```
paramsingh: the name `list` is used for the python type, maybe rename to something else?

db/data.py
```diff
@@ -59,6 +59,53 @@ def _has_key(dictionary, key):
...
+    trans = connection.begin()
+    try:
+        query = text("""
+                    DELETE
```
paramsingh: formatting again, sorry for being nitpicky. :(

db/manage_highlevel.py
```diff
@@ -0,0 +1,43 @@
...
+    """ Displays rows which do not contain highlevel metadata
+    This command displays the following information about each failed row:
+    1. Id (highlevel.id), 
+    2. Mbid (lowlevel.gid), 
+    3. Offset (important if there have been more than 1 submission for an mbid
+     so that we know which one do download)
```
paramsingh: `so that we know which one to delete`?

db/manage_highlevel.py
paramsingh: `--verbose`?

db/data.py
```diff
@@ -59,6 +59,53 @@ def _has_key(dictionary, key):
     return True
 
 
+def show_failed_rows():
+    """
```
paramsingh: A better description of what each of the values in the tuple returned are would be great.

For example: it isn't clear what offset means here.

pulkit6559: I have provided the detailed docstring about the values of the tuple in `show_failed_rows` command in `manage_highlevel.py` , should i provide it again here?

paramsingh: yes, because it is hard to keep track of where the function is used to get info about what it returns.

pulkit6559: Right, okay, i'll do that

pulkit6559:
@paramsingh i messed up while rebasing :3, is there a way to revert this?

db/data.py
```diff
@@ -58,6 +58,52 @@ def _has_key(dictionary, key):
...
+                    ROW_NUMBER ()
+                          OVER (PARTITION BY mbid) RowNum
+                          FROM highlevel
+                     LEFT JOIN highlevel_meta
+                         USING (id)
+                         WHERE highlevel_meta.id is null
+                      ORDER BY submitted
+                    """)
+        result = connection.execute(query).fetchall()
+    return result
```
paramsingh: we generally tend to return dicts instead of sqlalchemy resultproxy objects, I think we should do that here too.

alastair:
@paramsingh can you give a look at my changes here?

manage.py
```diff
@@ -240,9 +240,49 @@ def toggle_site_status():
         print('Done!')
 
 
-# Please keep additional sets of commands down there
+@cli.group()
+@click.pass_context
+def highlevel(ctx):
+    """Stuff"""
```
paramsingh: This docstring could be improved.

alastair: whoops. that was me testing the output of `--help`

PR #323, opened 2019-02-21
Only render the error message if it exists

alastair:
There was an existing error where we were not rendering the .error property, causing a react exception.

### DIFF
--- a/webserver/static/scripts/datasets/editor.js
+++ b/webserver/static/scripts/datasets/editor.js
@@ -276,13 +276,9 @@ var DatasetControlButtons = React.createClass({
             buttonText = "Update";
         }
         if (this.state.errorMsg) {
-            var errorMessage = "Unknown error";
-            if (typeof this.state.errorMsg.error !== 'undefined') {
-                errorMessage = this.state.errorMsg.error;
-            }
             var error = <p className='text-danger'>
                     <strong>An error occurred while submitting this dataset:</strong>
-                    &nbsp;{ errorMessage }
+                    &nbsp;{ this.state.errorMsg.error ||  "Unknown error" }
                 </p>
         } else {
             var error = ''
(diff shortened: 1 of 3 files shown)


webserver/static/scripts/datasets/editor.js
```diff
@@ -275,12 +275,21 @@ var DatasetControlButtons = React.createClass({
         if (this.props.mode == MODE_EDIT) {
             buttonText = "Update";
         }
+        if (this.state.errorMsg) {
+            var errorMessage = "Unknown error";
+            if (typeof this.state.errorMsg.error !== 'undefined') {
```
alastair: I'm not sure if this is the right way of seeing if an object has a propery/field

webserver/static/scripts/datasets/editor.js
```diff
@@ -275,12 +275,21 @@ var DatasetControlButtons = React.createClass({
         if (this.props.mode == MODE_EDIT) {
             buttonText = "Update";
         }
+        if (this.state.errorMsg) {
+            var errorMessage = "Unknown error";
```
alastair: please don't say anything about using `var`. this is old code

webserver/static/scripts/datasets/editor.js
MonkeyDo: I would do this:

```
var error = null;
if (this.state.errorMsg) {
  error = <p className='text-danger'>
                    <strong>An error occurred while submitting this dataset:</strong>
                    &nbsp;{ this.state.errorMsg.error ||  "Unknown error"}
                </p>
}
```

for succinctness, but the current PR will work just fine.
By having error as `null` by default, React won't render any node instead of a node with an empty string.

alastair: Thanks for the recommendation on making the code more readable. A review approval would be appreciated!

PR #324, opened 2019-02-24
Resolved TODO: get_next_pending_job now returns same data as get_job

pulkit6559:
**Code Improvement** 

Issue : The method get_pending_job in dataset_evel.py was running a query to return job id's which had status as Pending, which was then passed to get_job(job_id) to return information about these jobs

improvement : instead of running two separate queries for one task, the task is performed by a single query, and get_pending_job returns same data as the method get_job(job_id).

### DIFF
--- a/db/dataset_eval.py
+++ b/db/dataset_eval.py
@@ -13,6 +13,7 @@ STATUS_RUNNING = "running"
 STATUS_DONE = "done"
 STATUS_FAILED = "failed"
 
+# Columns to select when getting a job
 EVAL_COLUMNS = ["dataset_eval_jobs.id::text",
                 "dataset_snapshot.dataset_id::text",
                 "dataset_eval_jobs.snapshot_id::text",
@@ -25,7 +26,7 @@ EVAL_COLUMNS = ["dataset_eval_jobs.id::text",
                 "dataset_eval_jobs.created", 
                 "dataset_eval_jobs.updated", 
                 "dataset_eval_jobs.eval_location"]
-
+EVAL_COLUMNS_COMMA_SEPARATED = ", ".join(EVAL_COLUMNS)
 
 VALID_STATUSES = [STATUS_PENDING, STATUS_RUNNING, STATUS_DONE, STATUS_FAILED]
 
@@ -143,8 +144,12 @@ def validate_dataset_contents(dataset):
 
 
 def get_next_pending_job():
+    """
+    Get the earliest submitted job which is still in the pending state.
 
-    EVAL_COLUMNS_COMMA_SEPERATED = ", ".join([c for c in EVAL_COLUMNS])
+    Returns:
+         The next job to process
+    """
     with db.engine.connect() as connection:
         query = text(
             """SELECT %s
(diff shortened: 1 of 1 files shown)


pulkit6559:
@alastair  i've made the changes you requested

db/dataset_eval.py
```diff
@@ -13,6 +13,20 @@
...
+                "dataset_eval_jobs.status",
+                "dataset_eval_jobs.status_msg", 
+                "dataset_eval_jobs.result", 
+                "dataset_eval_jobs.options", 
+                "dataset_eval_jobs.training_snapshot", 
+                "dataset_eval_jobs.testing_snapshot", 
+                "dataset_eval_jobs.created", 
+                "dataset_eval_jobs.updated", 
+                "dataset_eval_jobs.eval_location"]
+ALL_EVAL_COLUMNS = ", ".join([c for c in EVAL_COLUMNS])
```
paramsingh: the name of this constant is a bit confusing. We should rename it to something that tells us that how it is different from `EVAL_COLUMNS`, maybe `eval_columns_comma_seperated`? 

Another thing that we could do is just keep the list and create the comma seperated string inside the query functions.

db/dataset_eval.py
```diff
@@ -129,40 +143,29 @@ def validate_dataset_contents(dataset):
...
     with db.engine.connect() as connection:
         query = text(
-            """SELECT id::text
+            """SELECT %s
                  FROM dataset_eval_jobs
-                WHERE status = :status
-                  AND eval_location = 'local'
-             ORDER BY created ASC
-                LIMIT 1""")
+                 JOIN dataset_snapshot ON dataset_snapshot.id = dataset_eval_jobs.snapshot_id
```
paramsingh: In general, we try to keep our SQL keywords right aligned.

```
  SELECT %s
    FROM dataset_eval_jobs
    JOIN dataset_snapshot
      ON dataset_snapshot.id = dataset_eval_jobs.snapshot_id
   WHERE status = :status
     AND eval_location = 'local'
ORDER BY created ASC
   LIMIT 1
```

paramsingh: https://github.com/metabrainz/guidelines/blob/master/Python.md#sql

pulkit6559: often while adding adding stuff to the query i loose track of the alignment :), i'll take care of it from now on

pulkit6559:
@paramsingh i've made the required changes

alastair:
Please make sure your commit messages are descriptive, and explain the changes that you made in the commit. Empty messages of `.` are not useful when looking at a list of commits. If you find that you forgot to make a change, you can consider using `git commit --amend` or `git rebase` to merge multiple commits into a single one.

pulkit6559:
@alastair  this was a life-changer :), i have merged them into a single commit

alastair:
I'm happy with this now, @paramsingh can you take a final look at it?

paramsingh:
Is the  test failure related? Otherwise, looks good to me.

pulkit6559:
@paramsingh i dont think so, i ran tests before submitting and didn't have this issue then...

PR #325, opened 2019-03-01
AB-386: Download dataset structure

pulkit6559:
https://tickets.metabrainz.org/browse/AB-386

**Feature Addition**
Added a download button to dataset view, which allows user to download contents of the dataset as a json file


paramsingh:
Is this a duplicate of #267?

pulkit6559:
@paramsingh yes i think they are the same, sorry i must have missed it, should i close this?

paramsingh:
I think so, yes. Next time, make sure that someone else isn't assigned to the ticket you're working on already, it helps us not duplicate work. :)

alastair:
It doesn't look like these two PRs are exactly the same... #267 deals with downloading the lowlevel files for a dataset, while this one just downloads the structure. I'm not sure if this is what @pulkit6559 intended to do, but I think that being able to download the dataset items is a good idea. 
Note that [we already have import functionality](https://github.com/metabrainz/acousticbrainz-server/blob/d7745afccf99aaf4014bea3014e2e648ace1bef9/webserver/views/datasets.py#L247-L264), and so this is a good complement. Perhaps we could have the same file structure so that you can take the output of download and use it to create a new dataset directly?

pulkit6559:
@alastair actually that's what i intended to do for which i used the existing get_json method, although i wasn't thinking of the same reason you mentioned :), i just thought it would be a cool new addition to the ui :)

pulkit6559:
How do i reopen this PR? or should i make a new one?

alastair:
I've reopened it. In the future it'd be better for a new feature like this for you to create a ticket in jira first. That would allow us to discuss functionality first before implementing it.

pulkit6559:
@alastair i've modified the file structure

webserver/templates/datasets/macros.html
```diff
@@ -3,28 +3,33 @@
...
+                href="{{ url_for('datasets.delete', dataset_id=dataset['id']) }}">
+                <span class="glyphicon glyphicon-trash" aria-hidden="true"></span>
+                Delete
+              </a>
+            {% endif %}
+
+        <a class="btn btn-info pull-right" title="Download this dataset" 
+          href="{{ url_for('datasets.download_csv', dataset_id=dataset['id']) }}">
+          <span class="glyphicon glyphicon-download-alt" aria-hidden="true"></span>
+          Download
```
alastair: Let's make this a bit more specific so that we don't confuse it with #267. Let's call it "Download schema"

webserver/views/datasets.py
```diff
@@ -92,6 +93,41 @@ def view(id):
     )
 
 
+def get_json(id):
```
alastair: you're not using this method twice any more, so I would move it back into the `view_json` function.

If we were going to keep this method I would give it a different name (`dataset_to_public_json`) and make it take the result of `get_dataset` as an argument instead of an id. It would also need documentation to say what it does.

webserver/views/datasets.py
```diff
@@ -92,6 +93,41 @@ def view(id):
...
+    for cls in dataset["classes"]:
+        dataset_clean["classes"].append({
+            "name": cls["name"],
+            "description": cls["description"],
+            "recordings": cls["recordings"],
+        })
+
+    return dataset_clean
+
+CSV_DIR = "./code/data/files/csv_files"
```
alastair: what's this? Don't think it's needed

webserver/views/datasets.py
```diff
@@ -92,6 +93,41 @@ def view(id):
...
+
+@datasets_bp.route("/<uuid:dataset_id>/download")
```
alastair: same with the text in the template, let's have a name here which indicates that we're just downloading the dataset schema

webserver/views/datasets.py
```diff
@@ -92,6 +93,41 @@ def view(id):
...
+def download_csv(dataset_id):
+    # return render_template("datasets/accuracy.html")
```
alastair: be careful when you're copying code from other parts of the project. Make sure you revise your code before committing it

webserver/views/datasets.py
```diff
@@ -92,6 +93,41 @@ def view(id):
...
+    ds = get_json(dataset_id)
+    ds_csv = ""
+    for ds_class in ds["classes"]:
+        class_name = ds_class["name"]
+        for rec in ds_class["recordings"]:
+            strList = [str(rec), str(class_name)]
+            ds_csv = ds_csv + ",".join(strList) + "\n"
```
alastair: Constructing CSV by hand is always a bad idea. Use the [csv.writer](https://docs.python.org/2/library/csv.html#csv.writer) module. You can write to a [stringio](https://docs.python.org/2/library/stringio.html) for a "file-like object":

```py
fp = StringIO.StringIO()
writer = csv.writer(fp)
for class in classes:
   for recording in recordings:
        writer.writerow([recording, class])

fp.getvalue() # the resulting csv file
```

webserver/views/datasets.py
```diff
@@ -92,6 +93,41 @@ def view(id):
...
+ 
+    return Response(ds_csv,
+                    mimetype="text/csv",
+                    headers={"Content-disposition":"attachment; filename=dataset.csv"})
```
alastair: You can use [flask.send_file](http://flask.pocoo.org/docs/1.0/api/#flask.send_file) to send a filepointer as an attachment (the stringio is a filepointer!)
Think about the filename that you send. It should be descriptive, and if I download many datasets I should know which one is which by looking at its filename.

alastair:
You don't have to write a message saying that you've made the changes - we receive notifications when you add new commits to a pull request, and will review them when possible.

alastair:
@pulkit6559 I've made some changes to this:
 * Move the csv generation to a separate function and add some tests
 * use `get_dataset` to get dataset for csv (to ensure that someone can't download a private dataset) and add tests
 * Rename the functionality from schema to annotations (I had a discussion with some coworkers and we decided that this was a clearer description)
 * fix the wrapping of the buttons with the extra item, without needing to use the `pull-right` class.

There were some bad bugs in the layout of this view when you deleted the `<div id="control-buttons">`, make sure that you carefully check the layout of the page when you make these kind of changes. If you're not sure how to make something work you can always say that you've made these changes because you're unable to keep it as it was, and ask for help and post a screenshot or something. Because of the changes that you made, the order of some buttons changed, and their alignment was incorrect. 

@paramsingh can you give this another quick review?

webserver/views/test/test_datasets.py
```diff
@@ -216,6 +201,36 @@ def test_delete(self):
...
+        self.assert404(resp)
+
+        # public dataset + not logged in, OK
+        dataset_id = dataset.create_from_dict(self.test_data, author_id=self.test_user_id)
+        resp = self.client.get(url_for(view_name, dataset_id=dataset_id))
+        self.assert200(resp)
+
+        self.temporary_login(self.test_user_id)
+
+        # private dataset + logged in, ok
```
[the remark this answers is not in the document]
alastair: this comment should read "public dataset + logged in, ok"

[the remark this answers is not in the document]
pulkit6559: I have corrected the comment

PR #326, opened 2019-03-03
AB-362: Cull/Filter clearly bad data

pulkit6559:
**Feature Addition**
## Problem
For a particular MbRecording Id acousticbrainz probably has multiple documents, but there may be some incorrectly matched mbids in that data, so for a handful of documents we can pick out rows which out lie the norms.

## Solution 
I added a new script filter_data.py which filters outlying data by calculating z-scores, it runs in a separate thread, i used a value of 1.65 for a 95% accuracy.
The PR is not yet complete but i wanted to see if i was going in the right direction.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -13,6 +13,7 @@ CREATE TABLE lowlevel_json (
   id          INTEGER, -- FK to lowlevel.id
   data        JSONB    NOT NULL,
   data_sha256 CHAR(64) NOT NULL,
+  label       TEXT, 
   version     INTEGER  NOT NULL-- FK to version.id
 );
 
--- a/admin/updates/20190330-add-mbid-labels.sql
+++ b/admin/updates/20190330-add-mbid-labels.sql
@@ -1,0 +1,6 @@
+
+BEGIN;
+
+ALTER TABLE lowlevel_json ADD COLUMN label TEXT;
+
+COMMIT;
--- a/db/data.py
+++ b/db/data.py
@@ -6,6 +6,7 @@ import json
 import os
 import db
 import db.exceptions
+import detect_outliers
 
 from sqlalchemy import text
 import sqlalchemy.exc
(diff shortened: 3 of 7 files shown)


alastair:
Thanks for making a start on this.
There are a few considerations that we have to make for this change. I'll list some ideas, and also add them to the ticket so that we can continue the discussion.

1. We should never delete data from the AcousticBrainz database. This will do things like remove data if it's referenced in other places (e.g. dataset_snapshot), and will make our incremental query parameters refer to different items (e.g. if we do ?n=3 on a low-level query).
2. Instead, I think it makes more sense to add a column or a table that marks the lowlevel row id which refers to the "most correct" submission that we think refers to this MBID. Once we've merged @rsh7's code, we can also include stuff like MBID redirects in this.
    2a: If we're absolutely sure that there are _bad_ submitted items, we can do the same with a 'bad' column, and use this column to skip rows in certain queries.
3. We can add an info endpoint (e.g. `/api/v1/<mbid>/info`) that reports what we believe is the mbid/offset item for this "best" mbid. We can also have a parameter of the low-level/high-level get methods that will directly give us this item when we query for it.
4. To make it easier to make changes to the tool which selects the best submission, it should be part of a separate file, instead of integrated as part of the filter script.
5. as always, we should have tests that ensure that what we implemented works as expected.

If you want to continue working on this functionality, I recommend that we perform the following steps:
* Add the ability to mark a submission as the "correct" one for an MBID
* Update our db query methods and API endpoints to select this submission
* Independently work on the selection/comparison method (this could continue to happen in acousticbrainz-labs)

Let me know if you have any questions/feedback to this plan

pulkit6559:
Those are some good ideas, i'll move the selection method to a different script to work on it independently, also, instead of a good/bad column, we could add a confidence column instead?.
And to select the mbid which we best believe is the one, we can pick the one with highest confidence, which could be easily done in sql

alastair:
I'm going to close this because I think we need to work more on the techniques for calculating bad items instead of working on the interface for it first. We can address these changes later.

PR #329, opened 2019-03-07
Proposal: Init refactor

alastair:
This PR replaces #309, and addresses my comments in that PR.

Instead of running a command based on `CONTAINER_NAME` in `/etc/service/uwsgi/run`, I create a separate runit service for each container that we want to run. All of these services have a `down` file to prevent the service from running by default, created in the Dockerfile. A new `rc.local` file does the same as the previous if statements in `uwsgi/run`, removing the relevant `down` file depending on what container is being run. This file is run by the [phusion baseimage init script before runit is started.](https://github.com/phusion/baseimage-docker#running_startup_scripts)
The `run` and `rc.local` files are already chmod +x in this repository, and so we don't need change their permissions in the Dockerfile.

Additionally make some changes to cron, fixing the installation of the file in `/etc/cron.d` (items in this file require a column for username, which was missing). I'm now running these cron items as root, and have removed the acousticbrainz user as it seems to only have been used for this.

## Potential release changes
I don't know if there is any monitoring on the MeB infrastructure that expects the services in all of these containers to be called `uwsgi`. This should be checked before deploying.

Cron jobs are being run as root now. If they do things with the cache (which they do) then there's a possibility that it'll write to the cache_namespaces folder as root, causing problems when the www-data user wants to write something. Is this volume shared between containers? Should we run cron jobs as www-data?

From what I understand, there is no configuration to start up cron in the deploy scripts. This should be added (configure a new container, and correctly set CONTAINER_NAME)


mayhem:
> I don't know if there is any monitoring on the MeB infrastructure that expects the services in all of these containers to be called uwsgi. This should be checked before deploying.

Scripts that do no expose network ports are not really being monitored -- we rely on runit to restart services if they crash. I'm still not certain that is enough...

mayhem:
> From what I understand, there is no configuration to start up cron in the deploy scripts. This should be added (configure a new container, and correctly set CONTAINER_NAME)

I'm not 100% certain what you mean by this. cron is started automatically by the base image and nothing special needs to be done. Only if you do NOT want cron running in all the instances based on one image do you need to take action.

alastair:
I mean the production deployment configuration that starts a new container to run cron in, and sets CONTAINER_NAME

mayhem:
Ok, I think I understand. @paramsingh should be able to handle the remaining bits without trouble. I like it!

alastair:
failing test is fixed in https://github.com/metabrainz/acousticbrainz-server/pull/328

docker/beta/rc.local
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# AcousticBrainz rc.local
+# Enables a specific runit service depending on what the
+# CONTAINER_NAME variable is set to
+
+
+if [ "${CONTAINER_NAME}" = "acousticbrainz-web-beta" ]
```
zas: Imho this env variable should be named `CONTAINER_ROLE`, container names are set externally, and can change, this is why we register "services" in prod.
But, since those roles are exclusive, the correct approach would be to have separate containers (and images).

docker/crontab
```diff
@@ -1,8 +1,5 @@
-PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
-PYTHONPATH=/usr/local/lib/python27.zip:/usr/local/lib/python2.7:/usr/local/lib/python2.7/lib-dynload:/usr/local/lib/python2.7/site-packages
-
 # Compute stats
-5 * * * * /usr/local/bin/python /code/manage.py compute_stats >> /var/log/stats_calc.log 2>&1
+5 * * * * root /usr/local/bin/python /code/manage.py compute_stats >> /var/log/stats_calc.log 2>&1
```
zas: Is there a reason to run this as root ?
Also what's this log file ? How is it rotated / size-controlled ?

Dockerfile
```diff
@@ -96,30 +95,38 @@ FROM acousticbrainz-base AS acousticbrainz-prod
 
 ARG deploy_env
 
-RUN pip install uWSGI==2.0.17.1
+RUN pip install --no-cache-dir uWSGI==2.0.17.1
+
+RUN mkdir /cache_namespaces
+RUN chown -R www-data:www-data /cache_namespaces
```
zas: Where are www-data user/group created ?

Dockerfile
```diff
@@ -96,30 +95,38 @@ FROM acousticbrainz-base AS acousticbrainz-prod
...
-RUN chmod 0644 /etc/cron.d/ab-crontab && crontab -u acousticbrainz /etc/cron.d/ab-crontab
-RUN touch /var/log/stats_calc.log /var/log/stats_cache.log && chown acousticbrainz:acousticbrainz /var/log/stats_calc.log /var/log/stats_cache.log
-
-# Make sure that the cron service doesn't start automagically
-# http://smarden.org/runit/runsv.8.html
+ADD docker/crontab /etc/cron.d/acousticbrainz
+RUN chmod 0644 /etc/cron.d/acousticbrainz
 RUN touch /etc/service/cron/down
 
+COPY ./docker/$deploy_env/rc.local /etc/rc.local
```
zas: Instead of duplicating scripts, what about using a common one and env var passed to the container ?
https://github.com/phusion/baseimage-docker#environment-variables

docker/dataset_eval/dataset_eval.service
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cd /code
```
zas: Is this strictly needed ?
I mean:

```bash
#!/bin/bash
exec python /code/worker_manage.py hl_extractor
```

Or change to `/code` is required ?
In this case, better `cd /code && exec ....`

alastair: again, this has been like this since we released. I'm happy to change to either of the two options, as long as it still works.

This will be running as root too, if you want us to run as unprivileged user we'll have to use a uid-dropping script. Does phusion provide one?

Dockerfile
alastair: I don't know. They've always been here (we've been running uwsgi as www-data since we deployed it)

docker/crontab
alastair: No strong reason to be root other than "it's easy". I'm happy to change this to any other user. The easiest would be www-data, like the webserver, as this allows us to share permissions. The next easiest would be to have a single acousticbrainz user in this image which is shared over all containers. Because this will potentially write to volumes mounted from a host, we should consider having a metabrainz standard of UIDs for each service that we can make consistent on all hosts.

This logfile will contain the output of the script that's run in the cronjob. Is there any logrotate happening in other MeB images that I can copy? The alternative is to redirect to the container's stdout so that we let docker deal with the logs

docker/beta/rc.local
alastair: I'm happy to call this CONTAINER_ROLE, but CONTAINER_NAME is what's currently used in AB and other python-brainzes (e.g. https://github.com/metabrainz/listenbrainz-server/blob/6cd87707144cac9bafa2529f9c93e3c3f48ff5cc/docker/prod/uwsgi/uwsgi.service). What other changes are needed (in deploy scripts?) to ensure that a new environment variable gets set?

What do you mean about separating containers and images? We already run each of these services as a separate container (web, hl extractor, dataset evaluator, cron)

Dockerfile
alastair: Currently (in listenbrainz), we have separate container names for production and beta:
https://github.com/metabrainz/listenbrainz-server/blob/6cd87707144cac9bafa2529f9c93e3c3f48ff5cc/docker/prod/uwsgi/uwsgi.service
https://github.com/metabrainz/listenbrainz-server/blob/6cd87707144cac9bafa2529f9c93e3c3f48ff5cc/docker/beta/uwsgi/uwsgi.service

We could definitely have a single rc.local file that shares the same container names in beta and prod, I'm not sure what this implies in terms of startup scripts/consul, and I don't know if it has the potential to cause confusion when we're trying to debug a problem under pressure where beta/prod are getting mixed up for some reason...

alastair:
Thanks for the feedback. In response to your general remarks:
 * yes, we need to fix logging/rotation for cron somehow. I'm in favour of sending it to the container's stdout
 * running as root: let's come up with a metabrainz-wide way of generating a per-project user that shares a UID over the whole infrastructure and run all services as that
 * too many processes in one container: We only run consul + main service in each container. this is no different from how AB currently works and is how all other MeB projects work
 * one image + many roles: This is how we currently do all python-brainzes (LB, CB, AB ) and it works  well for us, we don't want to change this system

docker/dataset_eval/dataset_eval.service
alastair: It looks like the runit way of doing this is by using `exec chpst -ufoo:group python ....`

http://smarden.org/runit/faq.html#userservices

paramsingh:
>yes, we need to fix logging/rotation for cron somehow. I'm in favour of sending it to the container's stdout

Sounds good to me.

paramsingh:
>f they do things with the cache (which they do) then there's a possibility that it'll write to the cache_namespaces folder as root, causing problems when the www-data user wants to write something. Is this volume shared between containers? Should we run cron jobs as www-data?

Right now, it isn't even a volume. We should fix it. All the acousticbrainz containers run on boingo right now, but I was planning on running the cron container on frank in anticipation of data dumps, which will need storage space (not enough on boingo)

docker/beta/rc.local
paramsingh: I've changed this to CONTAINER_ROLE and opened a sister PR in docker-server-configs (https://github.com/metabrainz/docker-server-configs/pull/78)

docker/crontab
paramsingh: Logging to container stdout/stderr now.

Dockerfile
paramsingh: I think keeping it like this for now is the better idea because this PR has been lingering for a long while.

[AB-408](https://tickets.metabrainz.org/browse/AB-408)

paramsingh:
>let's come up with a metabrainz-wide way of generating a per-project user that shares a UID over the whole infrastructure and run all services as that user

We could create a user (acousticbrainz for AB, listenbrainz for LB) and so on. Do we specify the UID in the Dockerfile?

paramsingh:
@brainzbot retest this please

alastair:
@paramsingh we still need to make a decision about how to deal with user ids here - use www-data, or create our own user and specify an id. Cronjobs and webserver need to be configured to run with this user.

paramsingh:
We created a separate user for the dumps job in LB. https://github.com/metabrainz/listenbrainz-server/blob/52705348/Dockerfile#L67

I think creating a user here and specifying it's ID sounds reasonable

docker/crontab
```diff
@@ -1,5 +1,5 @@
 # Compute stats
-5 * * * * root /usr/local/bin/python /code/manage.py compute_stats >> /var/log/stats_calc.log 2>&1
+5 * * * * root /usr/local/bin/python /code/manage.py compute_stats
```
alastair: do we know if this is going to do what we want? If cron is running as a job in runit, the output of these cron jobs isn't going to go to the container's stdout. I don't know how to make this do what we want it to do in this case. Thoughts?

PR #330, opened 2019-03-07
AB-91: Map model classes to more meaningful labels

pulkit6559:


pulkit6559:
should i add an flag/variable in this, to make these mappings optional?

pulkit6559:
@alastair the reason why I created a JSON file to do this was because I noticed not all models required a mapping, ie, I found 4 models out of 18 to have shortened keys,
Also, to enter this data into the model table would have required me to build a similar dict. I thought it would be better to to it this way, but if it is not complying with the AB coding standards, I'll revise the implementation

pulkit6559:
And I had that query because I couldn't find a usage where we didn't need the class mappings, so I thought if it would be okay to replace them altogether

webserver/views/api/v1/core.py
```diff
@@ -60,17 +60,21 @@ def get_high_level(mbid):
...
     ``n``. Documents are sorted by the submission time of their associated
     low-level documents.
+    You can get the full key values for model classes by specifying a
+    parameter ``map=True``
 
     You can get the total number of low-level submissions using ``/<mbid>/count``
     endpoint.
 
     :query n: *Optional.* Integer specifying an offset for a document.
+         map: *Optional.* Boolean flag to get mapped keys
```
alastair: check the syntax of this query for generating docs. I think that it should be `:query map:` to correctly build the docs. you should build them and verify the output.

webserver/views/api/v1/core.py
```diff
@@ -182,10 +186,15 @@ def get_data_for_multiple_recordings(collect_data):
     recordings = check_bad_request_for_multiple_recordings()
 
     recording_details = {}
-
+    map_keys = request.args.get("map")
     for recording_id, offset in recordings:
         try:
-            recording_details.setdefault(recording_id, {})[offset] = collect_data(recording_id, offset)
+            if map_keys:
+                recording_details.setdefault(recording_id, {})[offset] = collect_data(
```
alastair: there's a bug here if someone calls the lowlevel get multiple function and provides a `map` query parameter. It will try and call `load_low_level` with a `map_keys` parameter, which will fail

webserver/views/api/v1/core.py
```diff
@@ -182,10 +186,15 @@ def get_data_for_multiple_recordings(collect_data):
     recordings = check_bad_request_for_multiple_recordings()
 
     recording_details = {}
-
+    map_keys = request.args.get("map")
```
alastair: One thing that we try and do in our code is separate input/out for the user with the calculation or retrieval of data.
This makes it easier to test. This means that you should get the value of the `map` parameter in 
 the `get_many_higlevel` view, and pass it to this method as an argument. This will also fix the bug of passing the parameter to the get_many_lowlevel view

db/data.py
```diff
@@ -409,11 +409,46 @@ def load_high_level(mbid, offset=0):
             data = row[1]
             version = row[2]
             data["version"] = version
+            if map_keys:
+                data = map_model_classes(model,data)
```
alastair: I think that this would be better if we separated the logic for getting the highlevel data with the logic for mapping the class names. This means I think we should leave this method as it used to be, and have another method to load the mapping from the database. This could return: `{'model_name': {the mapping}, 'another model': {mapping}`, etc.
Then, in each highlevel API endpoint, if the parameter is set, perform a processing step that converts these values. This will make it easier to test that we can perform the mapping.

webserver/views/api/legacy.py
```diff
@@ -42,10 +42,13 @@ def get_high_level(mbid):
     If there is more than one document with the same mbid, you can specify
     an offset as a query parameter in the form of ?n=x, where x is an integer
     starting from 0
+    If you need the full key values instead of the shortened keys, you can specify
+    a parameter 'map=True' where map is a boolean value
     """
     mbid, offset = _validate_data_arguments(mbid, request.args.get("n"))
+    map_keys = request.args.get("map")
```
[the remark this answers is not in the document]
alastair: We should make the parameter name more explicit. How about `map_model_classe_names`?
The documentation could be updated to be a bit clearer:
> Some models use short codes as class names. To get more descriptive class names for these models, set a query parameter `map_model_classe_names=True`.

alastair:
One additional place that we can put the mappings, and the most important from the original bug report - we want to show the mapped versions on the recording page: http://acousticbrainz.org/c92dfc48-3428-4da3-b40d-eaec5fdbdbda
I think it would be nice to show the translated values on the page, but show the original values if you hover the mouse over them.

pulkit6559:
Thanks for the review :), i can see now why these changes were important, ill make sure to resolve the requested changes at the earliest

webserver/views/api/v1/core.py
pulkit6559: Now we don't have to perform this check, as `load_high_level` performs how it used to, now mapping will be done via a pre-processing step as you suggested

webserver/views/api/legacy.py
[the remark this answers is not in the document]
pulkit6559: i wanted to keep the parameter name concise as it has to be passed with the api url manually, dont you think?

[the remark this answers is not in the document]
pulkit6559: i thought it would be better to keep the parameter name short as it has to be passed with the api url, should i just change the documentation and keep the name as is?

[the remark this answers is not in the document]
alastair: I prefer to have the parameter easy to understand, rather than short. Can we name it `map_model_class_names`?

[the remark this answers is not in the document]
pulkit6559: sure, i'll rename it :)

db/data.py
```diff
@@ -414,6 +426,55 @@ def load_high_level(mbid, offset=0):
...
+    
+    Returns:
+        dict with mapped model class keys
+
+    """
+    mappings = db.data.model_class_mappings()
+    high_level = high_level_dict["highlevel"]
+
+    for model, map in mappings.iteritems():
+        for key, val in map.iteritems():
```
alastair: This is a small thing, but I would do split this into 2 methods and do these loops the other way around. Instead of looping through the mapping and changing the highlevel data, loop through the highlevel data and look at the mapping data.
Also, it may result in a bit more duplicate code, but if you pass the mapping into the function as well then you can perform tests without having to access the database:

```py

def map_highlevel_class_labels(highlevel, mappings):
   for model, data in highlevel.items():
        if model in mappings:
          map_single_highlevel_class_labels(data, mappings[model])
         #construct return value

def map_single_highelvel_class_labels(data, mapping):
    value = data["value"]
    data["value"] = mapping.get(value, value)
    all = {}
    for k, v in data["all"]:
        all[mapping.get(k, k)] = v
    data["all"] = all
    return data
```
(not tested)
Make sure the methods also have correct docstrings.
It's a bit longer, but I think it allows us to consider the workflow in terms of processing the data, rather than processing the mapping. Now you can also test them without having to add the data to the database first, instead you can create a generic mapping directory and test that the methods work.

webserver/views/api/legacy.py
```diff
@@ -66,6 +72,22 @@ def submit_low_level(mbid):
     return jsonify({"message": "ok"})
 
 
+def map_class_labels(high_level_json):
```
alastair: why do you have this method twice? It doesn't look like you call this one at all

webserver/views/api/v1/core.py
```diff
@@ -248,14 +256,25 @@ def get_many_highlevel():
...
     """
+
     recording_details = get_data_for_multiple_recordings(db.data.load_high_level)
-    return recording_details
+    map_keys = True if request.args.get(
+        "map_model_class_names") == 'True' else False
+    if map_keys:
+        for recording_id, offset in recording_details.iteritems():
+            for key, data in offset.iteritems():
+                recording_details.setdefault(recording_id, {})[key] = db.data.map_class_labels(
```
alastair: Is there a reason why you use `setdefault` here? can you not just use `recording_details[recording_id][key]`?

webserver/views/api/v1/test/test_core.py
```diff
@@ -118,7 +118,7 @@ def test_ll_no_item(self, ll):
     @mock.patch("db.data.load_high_level")
     def test_hl_numerical_offset(self, hl):
         hl.return_value = {}
-        resp = self.client.get("/api/v1/%s/high-level?n=3" % self.uuid)
+        resp = self.client.get("/api/v1/%s/high-level?n=3&map_model_class_names=False" % self.uuid)
```
alastair: You shouldn't need to modify these tests, since the default is to treat the parameter as False.

Instead, we should have a separate test which checks that the right thing is one when this parameter is set to true.

db/test/test_data.py
```diff
@@ -320,6 +320,51 @@ def test_load_high_level_missing_offset(self):
...
+            "hiphop":0,
+            "jazz":0,
+            "metal":0,
+            "pop":0,
+            "reggae":0,
+            "rock":0 },
+            "value":" "}}}
+
+        db.data.add_model('genre_tzanetakis', 'v1', 'show')
+        db.data.set_model_mappings('genre_tzanetakis', 'v1', """{
```
alastair: You shouldn't need this, because tests that inherit from `DatabaseTestCase` will load the database schema and run the `create_models.sql` script. Did you find that this model didn't exist in the `models` table?

pulkit6559: I found that the mappings didn't exist when ran the test script, but now that I recall, I must have written this test before i moved the mappings to `create_models.sql`. And yes that add_model shouldn't be there, ill remove it

db/data.py
```diff
@@ -239,6 +239,18 @@ def set_model_status(model_name, model_version, model_status):
              "model_version": model_version,
              "model_status": model_status})
 
+
+def set_model_mappings(model_name, model_version, mappings):
```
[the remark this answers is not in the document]
pulkit6559: I wrote it explicitly for the test but we dont need this after the test is modified, so ill remove it

alastair:
With some changes that we made this patch now has many difficult conflicts to resolve. I copied the main parts of these changes and made some more additions to https://github.com/metabrainz/acousticbrainz-server/pull/348, so I'm closing this

PR #335, opened 2019-04-11
AB-148, now dataset evaluation returns job_id and set of ignored MBIDs

lehinevych:
Modified the dataset validator to not raise an exception during validation, now the set of ignored recordings are returned. 
When dataset is submitted for evaluation the message with ignored recordings appear if there are any.

image

### DIFF
--- a/db/dataset_eval.py
+++ b/db/dataset_eval.py
@@ -18,13 +18,13 @@ EVAL_COLUMNS = ["dataset_eval_jobs.id::text",
                 "dataset_snapshot.dataset_id::text",
                 "dataset_eval_jobs.snapshot_id::text",
                 "dataset_eval_jobs.status",
-                "dataset_eval_jobs.status_msg", 
-                "dataset_eval_jobs.result", 
-                "dataset_eval_jobs.options", 
-                "dataset_eval_jobs.training_snapshot", 
-                "dataset_eval_jobs.testing_snapshot", 
-                "dataset_eval_jobs.created", 
-                "dataset_eval_jobs.updated", 
+                "dataset_eval_jobs.status_msg",
+                "dataset_eval_jobs.result",
+                "dataset_eval_jobs.options",
+                "dataset_eval_jobs.training_snapshot",
+                "dataset_eval_jobs.testing_snapshot",
+                "dataset_eval_jobs.created",
+                "dataset_eval_jobs.updated",
                 "dataset_eval_jobs.eval_location"]
 EVAL_COLUMNS_COMMA_SEPARATED = ", ".join(EVAL_COLUMNS)
 
@@ -62,14 +62,17 @@ def evaluate_dataset(dataset_id, normalize, eval_location, filter_type=None):
 
     Returns:
         ID of the newly created evaluation job.
+        Dataset
     """
     with db.engine.begin() as connection:
         if _job_exists(connection, dataset_id):
             raise JobExistsException
 
+        dataset = db.dataset.get(dataset_id)
         # Validate dataset contents
-        validate_dataset_contents(db.dataset.get(dataset_id))
-        return _create_job(connection, dataset_id, normalize, eval_location, filter_type)
+        validate_dataset_structure(dataset)
+        job_id = _create_job(connection, dataset_id, normalize, eval_location, filter_type)
+        return job_id, dataset
 
 
 def job_exists(dataset_id):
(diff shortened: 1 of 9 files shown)


alastair:
@brainzbot add to whitelist

db/dataset_eval.py
```diff
@@ -130,6 +131,7 @@ def validate_dataset_contents(dataset):
     found.
```
alastair: This documentation is out of date now, and should be updated

db/dataset_eval.py
```diff
@@ -68,8 +68,9 @@ def evaluate_dataset(dataset_id, normalize, eval_location, filter_type=None):
             raise JobExistsException
 
         # Validate dataset contents
-        validate_dataset_contents(db.dataset.get(dataset_id))
-        return _create_job(connection, dataset_id, normalize, eval_location, filter_type)
+        ignore_rec = validate_dataset_contents(db.dataset.get(dataset_id))
```
alastair: We have two types of dataset validation. One to check that each class has enough recordings (`validate_dataset_structure`), and another to check if the recordings are in AB. If some recordings are ignored, then the number if items in the class may have fallen below our threshold. We should check again after removing these recordings

webserver/views/datasets.py
```diff
@@ -227,6 +227,8 @@ def evaluate(dataset_id):
             flash.error("Cannot add this dataset because of a validation error: %s" % e)
         except db.dataset_eval.JobExistsException:
             flash.warn("An evaluation job for this dataset has been already created.")
+        if len(ignored_rec) > 0:
+            flash.warn("The following recording are ignored: %s" % ', '.join(ignored_rec))
```
alastair: I think it'd be a good idea to include more information in this message. At the very least I'd like to see the classes for each recording.
A great additional feature would be to use our metadata endpoint to also get the recording title and artist: 
https://github.com/metabrainz/acousticbrainz-server/blob/607167ba1078df7853dc7b498fa5c706b4cb3c72/webserver/views/datasets.py#L393

We have some react code for the dataset editor to do this, I'm not sure if we can reuse it on this page or if it's better to duplicate the code:
https://github.com/metabrainz/acousticbrainz-server/blob/f9afa4017c49b7d412b6557b68744d8fbbb2da07/webserver/static/scripts/datasets/editor.js#L598

db/dataset_eval.py
```diff
@@ -68,8 +68,9 @@ def evaluate_dataset(dataset_id, normalize, eval_location, filter_type=None):
...
+        job_id = _create_job(connection, dataset_id, normalize, eval_location, filter_type)
```
alastair: This is going to create an evaluation job containing all of the MBIDs in the dataset, including the ones that we have ignored. We should make sure that when we create the dataset snapshot that we don't include these recording ids, otherwise the dataset evaluation will fail:
https://github.com/metabrainz/acousticbrainz-server/blob/aeeec4d02b689b843efb65d4f4c58dd40ec20972/db/dataset.py#L280

db/dataset_eval.py
```diff
@@ -68,8 +68,9 @@ def evaluate_dataset(dataset_id, normalize, eval_location, filter_type=None):
...
+        return job_id, ignore_rec
```
alastair: With the changed return signature, the docstring should be changed too

db/dataset_eval.py
lehinevych: I add additional check to `evaluate_dataset` function

db/dataset_eval.py
lehinevych: I used the same `filter_dataset_contents` before creating a snap

webserver/views/datasets.py
lehinevych: I added the classes information to the output.  Regarding the metadata endpoint, I'm planning to work on that in separate PR.

db/dataset_eval.py
```diff
@@ -62,14 +63,20 @@ def evaluate_dataset(dataset_id, normalize, eval_location, filter_type=None):
...
         if _job_exists(connection, dataset_id):
             raise JobExistsException
 
+        dataset = db.dataset.get(dataset_id)
+        # Filter dataset
+        dataset, ignore_recordings = filter_dataset_contents(dataset)
         # Validate dataset contents
-        validate_dataset_contents(db.dataset.get(dataset_id))
-        return _create_job(connection, dataset_id, normalize, eval_location, filter_type)
+        validate_dataset_structure(dataset)
```
alastair: (we talked about this on irc)
this will raise an exception if the dataset structure isn't complete - after doing a filter, some class may have less than 2 recordings. This means that you don't get to send the flash message about recordings being removed, and the user only sees a message about the dataset being invalid, which is a bit confusing.

db/dataset_eval.py
```diff
@@ -143,6 +148,33 @@ def validate_dataset_contents(dataset):
...
+        Dictionary of class name and corresponding list of ignored recordings
+    """
+    rec_memo = {}
+    ignored_recordings = defaultdict(list)
+
+    for cls in dataset["classes"]:
+        for i, recording_mbid in enumerate(cls["recordings"]):
+            if recording_mbid in rec_memo and rec_memo[recording_mbid]:
+                pass
+            if db.data.count_lowlevel(recording_mbid) > 0:
```
alastair: (we talked about this on irc)
For large datasets (e.g. hundreds or thousands of recordings) a check like this will take a long time, especially if the dataset is valid (and it has to check _every_ recording). I'd prefer to try something like a join of the dataset table against the lowlevel table on the mbid field to get all missing recordings at once

PR #336, opened 2019-04-14
AB-385: Improve storage and retrieval of statistics

pulkit6559:
This PR addresses the issue of storage and retrieval of statistics. The changes mage in the pr are as follows:
- created a table `submission_stats` to store stats as single json entities for a given collected timestamp
- moved rows from the table `statistics` to the table `submission_stats` and added manage.py command to perform cleanup to store only daily data
- stored hourly data as cache 

## Deploy steps:
Run sql migration script `admin/updates/20190611-statistics.sql` before starting server again

### DIFF
--- a/admin/sql/create_primary_keys.sql
+++ b/admin/sql/create_primary_keys.sql
@@ -7,7 +7,7 @@ ALTER TABLE highlevel_meta ADD CONSTRAINT highlevel_meta_pkey PRIMARY KEY (id);
 ALTER TABLE highlevel_model ADD CONSTRAINT highlevel_model_pkey PRIMARY KEY (id);
 ALTER TABLE model ADD CONSTRAINT model_pkey PRIMARY KEY (id);
 ALTER TABLE version ADD CONSTRAINT version_pkey PRIMARY KEY (id);
-ALTER TABLE statistics ADD CONSTRAINT statistics_pkey PRIMARY KEY (name, collected);
+ALTER TABLE statistics ADD CONSTRAINT statistics_pkey PRIMARY KEY (collected);
 ALTER TABLE incremental_dumps ADD CONSTRAINT incremental_dumps_pkey PRIMARY KEY (id);
 ALTER TABLE "user" ADD CONSTRAINT user_pkey PRIMARY KEY (id);
 ALTER TABLE dataset ADD CONSTRAINT dataset_pkey PRIMARY KEY (id);
--- a/db/stats.py
+++ b/db/stats.py
@@ -14,7 +14,7 @@ import json
 
 from sqlalchemy import text
 
-STATS_CACHE_TIMEOUT = 60 * 10  # 10 minutes
+STATS_CACHE_TIMEOUT = 60 * 60  # 1 hour
 LAST_MBIDS_CACHE_TIMEOUT = 60  # 1 minute (this query is cheap)
 
 STATS_CACHE_KEY = "recent-stats"
@@ -75,7 +75,7 @@ def get_last_submitted_recordings():
 
 
 def compute_stats(to_date):
-    """Compute hourly stats to a given date and write them to
+    """Compute daily stats to a given date and write them to
     the database.
 
     Take the date of most recent statistics, or if no statistics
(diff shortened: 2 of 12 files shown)


db/stats.py
```diff
@@ -218,8 +247,8 @@ def load_statistics_data(limit=None):
         ret = []
         for line in stats_result:
```
alastair: I think that it would be a good idea to also select the statistics cached in redis for the most recent hour and add it to the list of statistics shown on the graph. This way the graph will update hourly throughout the day with the most recent data.
We should do something like the following:
* Get data from redis
* If it doesn't exist, skip it
* If it does exist, and is later than the most recent value in the database, append it

alastair:
@pulkit6559 I had a feeling that you had asked me some questions about this. Did you do it here or in IRC? Can you repost the comments here?

pulkit6559:
@alastair I did ask you some questions regarding this pr on IRC, mainly clarifying about what changes should i make in this, ie, as you said in your first comment, should I figure out a way to do this in the same table, or should make the changes you suggested in the current way and make two separate pr's, one for migration and the other for deploying the changes. 
I wasn't much active on the IRC the past week as I'm busy with my end-term examinations

db/test/test_stats.py
```diff
@@ -206,23 +202,6 @@ def test_write_and_get_statistics_data(self):
         ]
         self.assertEqual(list(expected_data), list(data))
 
-    def test_write_stats_error(self):
-        stats1 = {"lowlevel-lossy": 10, "lowlevel-lossy-unique": 6,
-                  "lowlevel-lossless": 15,
-                  "lowlevel-lossless-unique": 10,
-                  "lowlevel-total": 25, "lowlevel-total-unique": "not-a-number"}
```
alastair: todo: restore a test that ensures that we can render a row with a missing key or non-numeric value

db/stats.py
```diff
@@ -191,40 +177,20 @@ def format_statistics_for_highcharts(data):
...
-             WHERE date_part('hour', timezone('UTC'::text, collected)) = 0 
-                OR collected = (SELECT collected
-                                  FROM statistics
-                              ORDER BY collected DESC
-                                 LIMIT 1)
-          GROUP BY collected
-          ORDER BY collected DESC
+    qtext = """SELECT collected, stats
+                 FROM statistics
+             ORDER BY collected DESC
```
alastair: todo: restore adding the most up-to-date cached value to the end of the stats here when drawing the graph

db/test/test_stats.py
```diff
@@ -360,23 +338,22 @@ def test_get_statistics_history(self):
         self.assertEqual(history[0], {"data": [[1452384000000, 15], [1452470400000, 20]], "name": "Lossless (all)"})
 
     def test_stats_daily(self):
```
alastair: todo: this test only made sense when we got only daily data even when the database contained hourly information (it checked the one last item at the end). Rewrite to check the cache to get the last item

alastair:
@paramsingh can you have a look at this?

db/test/test_stats.py
```diff
@@ -16,16 +14,16 @@
...
+        date1 = datetime.datetime(2016, 1, 7, 10, 20, 39, tzinfo=pytz.utc)
+        next_day = db.stats._get_next_day(date1)
+        expected = datetime.datetime(2016, 1, 8, 0, 0, 0, tzinfo=pytz.utc)
+        self.assertEqual(next_day, expected)
 
-        date2 = datetime.datetime(2016, 01, 07, 13, 0, 0, tzinfo=pytz.utc)
-        next_hour = db.stats._get_next_hour(date2)
-        expected = datetime.datetime(2016, 01, 07, 14, 0, 0, tzinfo=pytz.utc)
-        self.assertEqual(next_hour, expected)
+        date2 = datetime.datetime(2016, 1, 7, 13, 0, 0, tzinfo=pytz.utc)
```
alastair: todo: this test was specifically checking that h:00:00 would result in h+1 for the hour. It doesn't make much sense to have these two tests when skipping days. Perhaps a check at midnight for this test

admin/sql/create_primary_keys.sql
```diff
@@ -7,7 +7,7 @@ ALTER TABLE highlevel_meta ADD CONSTRAINT highlevel_meta_pkey PRIMARY KEY (id);
 ALTER TABLE highlevel_model ADD CONSTRAINT highlevel_model_pkey PRIMARY KEY (id);
 ALTER TABLE model ADD CONSTRAINT model_pkey PRIMARY KEY (id);
 ALTER TABLE version ADD CONSTRAINT version_pkey PRIMARY KEY (id);
-ALTER TABLE statistics ADD CONSTRAINT statistics_pkey PRIMARY KEY (name, collected);
+ALTER TABLE statistics ADD CONSTRAINT statistics_pkey PRIMARY KEY (collected);
```
alastair: todo: this primary key needs to be added in the migration

PR #337, opened 2019-04-24
AB-394 - Bulk get items with a single database query

aidanlw17:
#### Check for database query method in endpoint query string 
API endpoint get_many_lowlevel takes optional parameter to specify
the use of this method: `query_method` if the value of
query_method is `single`, the new single query method will be used.
Else, it will continue using multi query method. This is meant to
simplify the testing process.

#### Single query implemented for low-level data
Single query uses subquery to partition recordings and attach
offset values, outer query is used to filter the results.

#### To Do:
- Test single query against previous method on larger datasets
- Consider changes that may allow the query to be better optimized
- Implement similar method for high level data

### DIFF
--- a/db/data.py
+++ b/db/data.py
@@ -6,6 +6,7 @@ import json
 import os
 import db
 import db.exceptions
+from collections import defaultdict
 
 from sqlalchemy import text
 import sqlalchemy.exc
@@ -385,7 +386,7 @@ def load_many_low_level(recordings):
             SELECT id, gid,
         ROW_NUMBER () 
               OVER (PARTITION BY gid 
-          ORDER BY submitted DESC) - 1 submission_offset
+          ORDER BY submitted) - 1 submission_offset
               FROM lowlevel
             )   AS partitions
          LEFT JOIN lowlevel_json 
@@ -395,9 +396,9 @@ def load_many_low_level(recordings):
         
         result = connection.execute(query, { 'recordings': tuple(recordings) })
 
-        recordings_info = {}
+        recordings_info = defaultdict(dict)
         for row in result.fetchall():
-            recordings_info.setdefault(str(row['gid']), {})[str(row['submission_offset'])] = row['data']
+            recordings_info[str(row['gid'])][str(row['submission_offset'])] = row['data']
 
         return recordings_info
 
--- a/db/data.py
+++ b/db/data.py
@@ -334,6 +334,7 @@ def write_high_level(mbid, ll_id, data, build_sha1):
             for model_name, data in json_high.items():
                 write_high_level_item(connection, model_name, model_version, ll_id, version_id, data)
 
+
 def load_low_level(mbid, offset=0):
     """Load lowlevel data with the given mbid as a dictionary.
     If no offset is given, return the first. If an offset is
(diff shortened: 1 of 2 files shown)


aidanlw17:
@alastair This pull request contains the single query method for AB-394. Still a work in progress but I think it would be beneficial to discuss the current state of the query, and how I can approach testing it on a larger scale. Thus far, I've tested the query with 40-50 recordings and seen ~11% improvement in the speed at this small scale. I still need to do unit testing for this work as well (as soon as I get some breathing room from my exams this week). 

One other thing I noticed: when adding offsets to a query string, if you specify an offset that does not exist (out of range for recordings in the db), the api just returns data for all other recordings but makes no note that one is missing. I feel as though the functionality of this endpoint might be improved with an error message for an offset out of range, or something to that effect?

db/data.py
```diff
@@ -359,6 +360,58 @@ def load_low_level(mbid, offset=0):
         return row[0]
 
 
+def load_many_low_level(recordings):
+    """Recordings is a list of tuples (mbid, offset).
```
alastair: We have some guidelines here regarding how to write docstrings: https://github.com/metabrainz/guidelines/blob/master/Python.md#docstrings

Specifically, I wouldn't have the "Steps" section below, or the text starting "Sample query response..."

db/data.py
```diff
@@ -359,6 +360,58 @@ def load_low_level(mbid, offset=0):
...
+    """
+    recordings_info = {}
+    for mbid, offset in recordings:
+        recordings_info.setdefault(mbid, {})[offset] = {}
+
+    recordings_info_json = json.dumps(recordings_info, separators=(',', ':'))
+    
+    with db.engine.connect() as connection:
+        query = text(
+            """ SELECT gid, RowNum, data
```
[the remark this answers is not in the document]
aidanlw17: Could you elaborate further on how this would look? I was under the impression that select will be executed after the where clause, and thus an alias from select cannot be used in the where clause - the full expression must be used instead. When trying to use the alias, I see an error in which the column name is not found. The postgres docs support this: https://www.postgresql.org/docs/current/sql-select.html#SQL-SELECT-LIST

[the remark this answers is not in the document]
aidanlw17: That being said, I see an alternative to both. Below, in the inner query, I can select `ROW_NUMBER () OVER (PARTITION BY gid) -1 RowNum`, then the column RowNum (soon to be submission_offset) will start from 0 in the first place.

db/data.py
aidanlw17: Thanks for the link! I've read and bookmarked those guidelines, and have altered the docstring accordingly.

db/data.py
```diff
@@ -361,53 +361,43 @@ def load_low_level(mbid, offset=0):
...
-                WHERE (:recordings_info)::jsonb ? gid::text 
-                AND (:recordings_info)::jsonb->gid::text ? (RowNum-1)::text
-            """)
+        query = text("""
+            SELECT gid, submission_offset, data
+              FROM (
+            SELECT id, gid,
+        ROW_NUMBER () 
+              OVER (PARTITION BY gid 
+          ORDER BY submitted DESC) - 1 submission_offset
```
aidanlw17: @alastair I've reformatted the query to match the styling that you suggested. I've also decremented the row number within the sub query, so we can use submission_offset in both the outer `SELECT` and `WHERE` clauses. I've also ordered the recordings by submission date such that the most recent submission is always offset 0.

db/data.py
```diff
@@ -361,53 +361,43 @@ def load_low_level(mbid, offset=0):
...
+              FROM lowlevel
+            )   AS partitions
+         LEFT JOIN lowlevel_json 
+             USING (id)
+             WHERE (gid, submission_offset) IN :recordings
```
aidanlw17: @alastair here, I've simplified the query using tuples rather than jsonb. I agree that the previous complexity was unnecessary!

webserver/views/api/v1/core.py
```diff
@@ -227,19 +222,20 @@ def get_many_lowlevel():
...
 
     :resheader Content-Type: *application/json*
     """
     # Query option allowed for testing purposes:
     if request.args.get("query_method") == "single":
         # Use single-query method
-        recording_details = get_data_for_multiple_recordings(db.data.load_many_low_level)
+        recordings = check_bad_request_for_multiple_recordings()
+        recording_details = db.data.load_many_low_level(recordings)
+        return jsonify(recording_details)
```
aidanlw17: I've moved the calls to check_bad_request_for_multiple_recordings() and load_many_low_level(recordings) to make this process easier to follow. However, I've assumed that I should leave the pre-existing flow for the multiple query method as is for now?

db/data.py
alastair: this query is much better! Just a small thing, we want to order by ASC - offset 0 is the first submission for a given gid.

db/data.py
```diff
@@ -359,6 +360,48 @@ def load_low_level(mbid, offset=0):
...
+        """)
+        
+        result = connection.execute(query, { 'recordings': tuple(recordings) })
+
+        recordings_info = {}
```
alastair: you can use `defaultdict(dict)` here, then you dont need to use setdefault below

alastair:
Thanks! let's take this out of draft mode.
Don't worry about the failing tests - we'll fix them in #338

aidanlw17:
@alastairp great thanks, I’ll make the change to ascending on the query and then take it out of draft!

alastair:
We're going to close this because the changes that we're making in AB-406 and #339/#340 will fix this in a better way

PR #338, opened 2019-04-25
Upgrade dependencies

paramsingh:
Test PR.

### DIFF
--- a/webserver/views/api/v1/test/test_core.py
+++ b/webserver/views/api/v1/test/test_core.py
@@ -68,7 +68,7 @@ class CoreViewsTestCase(ServerTestCase):
         self.assertEqual(404, resp.status_code)
         load_low_level.assert_not_called()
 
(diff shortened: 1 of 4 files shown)


alastair:
great, these failing tests are exactly what I expected. Let's fix them in this PR, then when BU is released, we can rebase to update the requirements line, and merge

webserver/views/api/v1/test/test_core.py
```diff
@@ -67,8 +67,8 @@ def test_ll_bad_uuid_404(self, load_low_level):
         resp = self.client.get("/api/v1/nothing/low-level")
         self.assertEqual(404, resp.status_code)
         load_low_level.assert_not_called()
-        
-        expected_result = {"message": "The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again."}
+
+        expected_result = {u"message": u"The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again."}
```
alastair: don't need the `u` here, it should work as long as we remove the extra space

PR #339, opened 2019-05-08
AB-406 Step 1: Offset migration for existing submissions

aidanlw17:
#### Migration to add nullable submission_offset column
This will allow for us to populate the submission offset but the field 
will be null for recordings that are submitted during the process, 
allowing us to catch and update them in the next step. Note that the
offset column will only exist in the lowlevel table.

#### Migration to add offset for existing submissions
Uses a subquery with a window function to calculate the offsets for all
existing entries, then update the submission_offset column in lowlevel 
with the results.

### DIFF
--- a/add_submission_offsets.py
+++ b/add_submission_offsets.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 from flask.cli import FlaskGroup
 import click
 from collections import defaultdict
@@ -9,19 +11,22 @@ from sqlalchemy import text
 
 cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app_flaskgroup)
 
+
 @cli.command(name='add-offsets')
 @click.option("--limit", "-l", default=10000)
 def add_offsets(limit):
     """Update lowlevel submission offsets with a specified limit."""
     incremental_add_offset(limit)
 
+
 def incremental_add_offset(limit):
     with db.engine.connect() as connection:
 
         # Find number of items in table
         size_query = text("""
-            SELECT MAX(id) AS size
+            SELECT count(*) AS size
               FROM lowlevel
+             WHERE submission_offset IS NULL
         """)
         size_result = connection.execute(size_query)
         table_size = size_result.fetchone()["size"]
@@ -53,13 +58,13 @@ def incremental_add_offset(limit):
         print("Starting batch insertions...")
         print("============================")
         while True:
-            batch_result = connection.execute(batch_query, { "limit": limit })
+            batch_result = connection.execute(batch_query, {"limit": limit})
             if not batch_result.rowcount:
-                print("\nSubmission offset exists for all items. Exiting...")
+                print("Submission offset exists for all items. Exiting...")
                 break
 
             batch_count += 1
-            print("\nUpdating batch {}:".format(batch_count))
+            print("Updating batch {}:".format(batch_count))
             with connection.begin() as transaction:
                 for id, gid in batch_result.fetchall():
                     if gid in max_offsets:
@@ -75,14 +80,10 @@ def incremental_add_offset(limit):
                            SET submission_offset = :offset
                          WHERE id = :id
                     """)
-                    connection.execute(query, { "id": id, "offset": offset })
+                    connection.execute(query, {"id": id, "offset": offset})
                     item_count += 1
-                    print("\r   Inserted {}/{} items...".format(item_count, table_size)),
+            print(" Batch done, inserted {}/{} items...".format(item_count, table_size)),
         print("")
-
 
         print("============================")
         print("Batch insertions finished.")
-
-
-
--- a/admin/updates/20190508-submission-offsets-1.sql
+++ b/admin/updates/20190508-submission-offsets-1.sql
@@ -1,0 +1,15 @@
+BEGIN;
+
+ALTER TABLE lowlevel ADD COLUMN submission_offset INTEGER;
+
+-- Updating the submission offset column based on determination of offset
+UPDATE lowlevel 
+   SET submission_offset = offset_table.submission_offset
+  FROM (
+SELECT id, ROW_NUMBER () 
+	   OVER (PARTITION BY gid ORDER BY submitted) - 1 submission_offset 
+  FROM lowlevel)
+    AS offset_table
+ WHERE lowlevel.id = offset_table.id;
+
+COMMIT;
--- a/admin/updates/20190508-submission-offsets-1.sql
+++ b/admin/updates/20190508-submission-offsets-1.sql
@@ -2,14 +2,4 @@ BEGIN;
 
 ALTER TABLE lowlevel ADD COLUMN submission_offset INTEGER;
 
--- Updating the submission offset column based on determination of offset
-UPDATE lowlevel 
-   SET submission_offset = offset_table.submission_offset
-  FROM (
-SELECT id, ROW_NUMBER () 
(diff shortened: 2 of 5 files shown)


admin/updates/20190508-submission-offsets-1.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN;
+
+ALTER TABLE lowlevel ADD COLUMN submission_offset INTEGER;
+
+-- Updating the submission offset column based on determination of offset
```
alastair: Doing this in a single query is a good idea, but this will block the entire table and stop other insert actions from taking place. Unfortunately we're going to have to do it in a script which is a bit more complex and will take longer to run, but will leave the site up while it runs.

Take a look at making a script like manage.py that uses a click `cli` in order to start up the application and connect to the database:
https://github.com/metabrainz/acousticbrainz-server/blob/f9afa4017c49b7d412b6557b68744d8fbbb2da07/manage.py#L24-L29

Do a query like:
`select id, gid from lowlevel where submission_offset is null limit 10000`

the limit can be a parameter in the script that we can change

Use something like our offset counting code in the dumps to find the current max offset for each mbid:
https://github.com/metabrainz/acousticbrainz-server/blob/e8fceb76b434bb97aabf1b006e087097b2f66089/db/dump.py#L546-L555
this query might be good:
`select gid, max(submission_offset) from lowlevel where submission_offset is not null group by gid`

then you can loop through the set of 10000 items, and insert the offset counts. Do this loop in a transaction.

You don't need to write the sql in the db module, just write it inline in this script.

Make sure that you print some debugging information - each time it loops around, each time it starts and finishes doing the inserts, how many items have been done and how many remain.

submission_offsets.py
```diff
@@ -0,0 +1,68 @@
...
+		offset_result = connection.execute(offset_query)
+
+		max_offsets = defaultdict(int)
+		for gid, max_offset in offset_result.fetchall():
+			max_offsets[gid] = max_offset
+
+		print("Starting batch insertions...")
+		print("============================")
+		count = 0
+		for id, gid in batch_result.fetchall():
```
aidanlw17: @alastair When you said that the loop should be within a transaction, is it sufficient for the entire operation to be within the same transaction (starting with `with db.engine.connect() as connection:` on line 19)? Or should I embed another transaction here that is just for the loop?

alastair: this loop should be in an additional transaction - you can do this with `with connection.begin() as transaction:`

submission_offsets.py
```diff
@@ -0,0 +1,68 @@
...
+			print("Finished {}/{} items".format(count, limit))
```
alastair: will these print messages occur for every item in the 10000 batch?

submission_offsets.py
```diff
@@ -0,0 +1,68 @@
...
+@cli.command(name='add-offsets')
+@click.option("--limit", "-l", default=10000)
+def add_offsets(limit):
+	"""Update lowlevel submission offsets with a specified limit."""
+	incremental_add_offset(limit)
+
+def incremental_add_offset(limit):
+	with db.engine.connect() as connection:
+		# Find the next batch of items to update
+		batch_query = text("""
```
alastair: It looks to me like this script will only do 10000 items and then quit, is that right?
Instead, it should repeatedly do 10000 items until there are no more remaining. In a loop, get x items, then for each of them count the submission offsets and add them in a transaction.
Then, loop around again and get another x items. You can check the number of rows returned by the batch_result query, and when it's 0, quit.

submission_offsets.py
```diff
@@ -0,0 +1,68 @@
...
+			SELECT id, gid
+			  FROM lowlevel
+			 WHERE submission_offset IS NULL
```
alastair: this query will need an order by id, to ensure that we're getting items in submission order

submission_offsets.py
```diff
@@ -0,0 +1,68 @@
...
+			 LIMIT :limit
+		""")
+		batch_result = connection.execute(batch_query, { "limit": limit })
+
+		# Find max existing offsets
+		offset_query = text("""
```
alastair: if we keep the max_offsets dictionary, we only need to do this query once when the script starts up, instead of for each 10,000 items

submission_offsets.py
aidanlw17: Yes, the printing occurs for each item in the 10000 batch as the offset is being found and inserted. If we are running this continuously with the added loop that you mentioned, would it be best for me to remove some of the printing? I.e. print for each batch starting and ending rather than on each item in a batch?

alastair: exactly, there should be one print statement for each 10000 items

submission_offsets.py
aidanlw17: Does this query order by id when no other order is specified? I was trying this query using an order by id, and trying it without and it was returning the same order (by id) in both cases.

submission_offsets.py
aidanlw17: That’s correct, currently it only does 10000 and then quits. I understand what you’re saying. I’ll add the loop so that it completes all the batches one after another.

aidanlw17:
@alastair The latest commit here should now allow the offsets for the whole table to be updated in one execution.

submission_offsets.py
alastair: the sql standard doesn't specify a behaviour if you don't specify `order by`. perhaps by a programming decision or luck postgres appears to return items in id order, but this is not guaranteed. We should always be safe and explicitly set it.

aidanlw17: Okay thanks, I've specified order by id in the latest commit.

add_submission_offsets.py
```diff
@@ -0,0 +1,89 @@
+from __future__ import print_function
```
alastair: I renamed this file to make its name make a bit more sense, and reformatted a few things. Take a look at the whitespace changes for a better idea of the formatting that we use. an ide or other tool can help you do this too

add_submission_offsets.py
```diff
@@ -0,0 +1,89 @@
...
+    """Update lowlevel submission offsets with a specified limit."""
+    incremental_add_offset(limit)
+
+
+def incremental_add_offset(limit):
+    with db.engine.connect() as connection:
+
+        # Find number of items in table
+        size_query = text("""
+            SELECT count(*) AS size
```
alastair: we're missing some rows in this table, so max(id) is not the same as the number of rows, I changed it

add_submission_offsets.py
```diff
@@ -0,0 +1,89 @@
...
+                    offset = max_offsets[gid]
+
+                    query = text("""
+                        UPDATE lowlevel
+                           SET submission_offset = :offset
+                         WHERE id = :id
+                    """)
+                    connection.execute(query, {"id": id, "offset": offset})
+                    item_count += 1
+            print(" Batch done, inserted {}/{} items...".format(item_count, table_size)),
```
alastair: This is good, but I moved the print message to the end of each batch. It turns out that printing stuff to the screen can take a lot of time, and it's a thing that can actually slow down processes like this that run really quickly. We're not missing out on any information by not updating every item

add_submission_offsets.py
aidanlw17: Thanks for the changes. For whitespace, I see the change to two lines between functions, and also the two lines before the first function definition after the cli setup. Are there any other key whitespace things that I should be aware of?

PR #340, opened 2019-05-10
AB-406 Part 2: Offset for missing entries, offset computed on submission

aidanlw17:
#### Add offset for missing items
When submissions are stopped, this migration adds the submission 
offset to all rows that have submission offset null. This ensures that
there are no rows missing an offset.

#### Compute offset on submission
Computes submission when write_low_level is called, and offset is
inserted as part of the low level data when it is written. This step
ensures that all submissions in the future are tracked with an offset
as soon as they enter the database.


db/data.py
```diff
@@ -187,6 +189,22 @@ def _insert_lowlevel_json(connection, ll_id, data_json, data_sha256, version_id)
...
+        """ Get highest existing submission offset for mbid, then increment """
+        query = text("""
+            SELECT MAX(submission_offset)
+              FROM lowlevel
+             WHERE gid::text = :mbid
+        """)
+        result = connection.execute(query, {"mbid": mbid})
+
+        row = result.fetchone()
+        if row[0] is not None:
```
alastair: if you give a column alias in the sql statement then you can use `row['max_offset']` here to make it a bit more explicit what we're reading

db/test/test_stats.py
```diff
@@ -46,7 +46,7 @@ def test_get_last_submitted_recordings(self, dbset, dbget):
                     "version": {"essentia_build_sha": "sha"},
                 },
             }
-            db.data.write_low_level(rand_mbid, data, gid_types.GID_TYPE_MBID)
+            db.data.write_low_level(str(rand_mbid), data, gid_types.GID_TYPE_MBID)
```
alastair: did this stop working without the call to `str()`? I'm interested to know if we should make sure that other places that call this still work - do all places that call this method in the main code pass a string, or a uuid object?

db/data.py
```diff
@@ -200,7 +218,8 @@ def _insert_lowlevel_json(connection, ll_id, data_json, data_sha256, version_id)
             return
 
         try:
-            ll_id = _insert_lowlevel(connection, mbid, build_sha1, is_lossless_submit, is_mbid)
+            submission_offset = _get_submission_offset(connection, mbid)
+            ll_id = _insert_lowlevel(connection, mbid, build_sha1, is_lossless_submit, is_mbid, submission_offset)
```
alastair: We should have a test here to ensure that the offset is the expected value (0 if we add an mbid once, 1 if we add it twice, and another mbid will be 0 again)

db/test/test_stats.py
aidanlw17: This did stop working without the call to str(), and I believe that it was because I casted the gid to text in the query for submission offset. This was the error: 
Screen Shot 2019-05-18 at 12 00 20 PM

I removed the cast to text in the query and removing the call to `str()` allowed the test to pass. The api endpoint for `submit_low_level` makes a call to `str()` on the mbid, but in `db/data.py` the function `submit_low_level_data` also casts the mbid as a string before passing it to `write_low_level`. 

So typically, the main code passes a string rather than a uuid. But it appears to me that the comparison between gid in the lowlevel table and a string mbid works without using `gid::text`. So by removing gid::text, we can then use write_low_level with both strings and uuids?

db/data.py
```diff
@@ -187,6 +189,22 @@ def _insert_lowlevel_json(connection, ll_id, data_json, data_sha256, version_id)
                                    "data_sha256": data_sha256,
                                    "version": version_id})
 
+    def _get_submission_offset(connection, mbid):
+        """ Get highest existing submission offset for mbid, then increment """
+        query = text("""
+            SELECT MAX(submission_offset) as max_offset
+              FROM lowlevel
+             WHERE gid = :mbid
```
aidanlw17: @alastair I've removed the cast `gid::text` here, which allowed the query to work without a call to `str()` in `test_stats` file.

db/test/test_data.py
```diff
@@ -72,6 +77,34 @@ def test_submit_low_level_data_missing_keys(self, clean, write, sanity):
             db.data.submit_low_level_data(self.test_mbid, self.test_lowlevel_data, gid_types.GID_TYPE_MBID)
 
 
+    def test_write_low_level_submission_offset(self):
```
aidanlw17: @alastair This is a test to ensure the offset value is expected. Given that `_get_submission_offset` is a nested function, I was not sure about the best way to use unittestt. Does querying for the offset after writing data each time here seem like an appropriate method?

alastair: this test is fine, for now let's leave the duplicated code. If we need the same functionality somewhere else we should copy it out in the data module

alastair:
Can you rebase this against master so that it's ready to merge when we deploy part 1?

PR #342, opened 2019-05-15
Add ratelimit header injector and ratelimit decorators

mayhem:
As it says on the tin -- nothing super special.

### DIFF
--- a/webserver/views/api/v1/dataset_eval.py
+++ b/webserver/views/api/v1/dataset_eval.py
@@ -55,7 +55,7 @@ def get_jobs():
 
 @bp_dataset_eval.route("/jobs/<uuid:job_id>", methods=["GET"])
 @auth_required
-@ratelimit
+@ratelimit()
 def job_details(job_id):
     """Returns the details of a particular job.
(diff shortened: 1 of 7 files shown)


webserver/views/api/v1/dataset_eval.py
```diff
@@ -11,6 +12,7 @@
 
 @bp_dataset_eval.route("/jobs", methods=["GET"])
 @auth_required
+@ratelimit()
 def get_jobs():
```
alastair: this module is used by js

webserver/views/api/v1/datasets.py
```diff
@@ -21,6 +22,7 @@ def get_dataset(dataset_id):
 
 @bp_datasets.route("/", methods=["POST"])
 @auth_required
+@ratelimit()
```
alastair: this module is used by js

alastair:
can we have a management command to set cache values?

alastair:
The failing tests related to the error messages were fixed in #338, so a rebase against master should fix those

PR #343, opened 2019-05-15
Store the maximum row ID done in hl_calc and get rows after that row ID

paramsingh:
As discussed on IRC.

The query plan (from my local environment):

```
                                                  QUERY PLAN                                            
       
--------------------------------------------------------------------------------------------------------
-------
 Limit  (cost=45.95..45.96 rows=1 width=68)
   ->  Sort  (cost=45.95..45.96 rows=1 width=68)
         Sort Key: ll.id
         ->  Nested Loop  (cost=23.79..45.94 rows=1 width=68)
               ->  Hash Right Join  (cost=23.65..45.27 rows=2 width=20)
                     Hash Cond: (hl.id = ll.id)
                     Filter: (hl.mbid IS NULL)
                     ->  Seq Scan on highlevel hl  (cost=0.00..19.20 rows=920 width=20)
                     ->  Hash  (cost=20.02..20.02 rows=290 width=20)
                           ->  Bitmap Heap Scan on lowlevel ll  (cost=6.40..20.02 rows=290 width=20)
                                 Recheck Cond: (id > 0)
                                 ->  Bitmap Index Scan on lowlevel_pkey  (cost=0.00..6.33 rows=290 width
=0)
                                       Index Cond: (id > 0)
               ->  Index Scan using lowlevel_json_pkey on lowlevel_json llj  (cost=0.14..0.33 rows=1 wid
th=36)
                     Index Cond: (id = ll.id)

```

### DIFF
--- a/db/data.py
+++ b/db/data.py
@@ -470,8 +470,14 @@ def get_unprocessed_highlevel_documents_for_model(highlevel_model, within=None):
         docs = result.fetchall()
         return docs
 
-def get_unprocessed_highlevel_documents():
-    """Fetch up to 100 low-level documents which have no associated high level data."""
+def get_unprocessed_highlevel_documents(min_row_id=0):
+    """Fetch up to 100 low-level documents which have no associated high level data.
+
+    Args:
+        min_row_id (int): the minimum lowlevel row ID the function should return. This
+                        is used to restrict the query to only get docs submitted
+                        after a particular time.
+    """
     with db.engine.connect() as connection:
         query = text(
                 """SELECT ll.gid::text
@@ -483,8 +489,10 @@ def get_unprocessed_highlevel_documents():
             LEFT JOIN highlevel AS hl
                    ON ll.id = hl.id
                 WHERE hl.mbid IS NULL
+                  AND ll.id >= :min_row_id
+             ORDER BY ll.id
                 LIMIT 100""")
-        result = connection.execute(query)
+        result = connection.execute(query, {"min_row_id": min_row_id})
         docs = result.fetchall()
         return docs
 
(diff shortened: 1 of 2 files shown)


db/data.py
```diff
@@ -483,8 +489,10 @@ def get_unprocessed_highlevel_documents():
             LEFT JOIN highlevel AS hl
                    ON ll.id = hl.id
                 WHERE hl.mbid IS NULL
+                  AND ll.id >= :min_row_id
+             ORDER BY ll.id
```
alastair: does this give more or less the same results as without the `ORDER BY`? part of what we're trying to do here is reduce the amount of work that the query has to do. Instinctively I think it doesn't matter _what_ ids we get as long as they're above min_row_id.

@mayhem had a thought that performing operations such as join and sort without the lowlevel_json table would be a bit faster. We saw that with queries that return a lot of data (such as this one, joining against lowlevel_json) often have to use the disk to perform a sort/order because there's not enough free memory. I worry that adding a sort but keeping lowlevel_json will cause this to happen again.

Can we perform some additional query plans (using explain analyze to get real results, instead of just expected plans),

* select only lowlevel and have an order by
* select lowlevel + lowlevel_json with order by
* the two above without order by

If the query plan without lowlevel_json is much better, we'll also have to check that selecting just 200-ish items from lowlevel_json in a second query will also not take much time. Again, I believe that this should be fine with our index - we'll be doing no complex sorts or filtering.

The query plan looks ok - I see that the WHERE on ll.id happens as the inner-most operation, so I think that this means that it will restrict the number of rows that we have to deal with when we do the left join and where hl.mbid is null. This is good. To confirm, can we do an `explain analyze` with and without the ll.id limit to verify that the execution speed changes?

paramsingh:
https://gist.github.com/paramsingh/741a06789221f128a93a3c9c2543a6d5

alastair:
With my changes to the hl extractor, I re-implemented this in 9f33d69401aa0d17bf7a0df717680376baa87382, closing.

PR #345, opened 2019-06-03
Add ratelimiting

paramsingh:
# Deployment steps

* Add `ratelimit_per_ip` (100) and `ratelimit_window` (10) key to `config.prod.json` in `docker-server-configs/consul/AB`
* commit and push to https://github.com/metabrainz/docker-server-configs
* restart containers.


manage.py
```diff
@@ -241,6 +241,14 @@ def toggle_site_status():
         print('Done!')
 
 
+@cli.command(name='set_rate_limits')
+@click.argument('per_ip')
+@click.argument('per_token')
+@click.argument('window_size')
+def set_rate_limits(per_ip, per_token, window_size):
```
alastair: we don't have token auth in AB, so how about we remove this parameter (and optionally set it to the same as per_ip?)
Please add documentation and types to the arguments.
It'd make sense to also print what the parameters are set to when you run this function.
Do we want any sanity checking to make sure that these settings aren't set to anything really strict? e.g. a warning if the effective limit goes lower than 1 per second

docs/api.rst
```diff
@@ -23,3 +23,31 @@ Datasets
    :blueprints: api_v1_datasets
    :include-empty-docstring:
    :undoc-static:
+
+Rate limiting
+^^^^^^^^^^^^^
+
+The AcousticBrainz API is rate limited via the use of rate limiting headers that
+are sent as part of the HTTP response headers. Each call will include the
+following headers:
```
alastair: What's the default rate limit if we don't make any changes? It'd be nice to say this here. "We typically set the limit to 10 queries every 10 seconds, but these values may change. Make sure you check the response headers if you want to know the specific values"

alastair: looking at the BU code it seems like the default is 30 queries in 10 seconds. Is this actually useful for us - should we check the mean number of queries per IP address for AB to ensure that this actually results in a reduction of queries, or see if we need to reduce the default.
Our custom defaults should be in the config file instead of hard-coded.

webserver/testing.py
```diff
@@ -1,8 +1,14 @@
 from db.testing import DatabaseTestCase
 
+from brainzutils.ratelimit import set_rate_limits
 
 class ServerTestCase(DatabaseTestCase):
 
+
+    def setUp(self):
+        super(ServerTestCase, self).setUp()
+        set_rate_limits(1000, 1000, 10000)
```
alastair: instead of selecting large magic numbers here, it'd be nice to be able to actually disable the rate-limiting during tests. Can we add a ticket to BU to add this?

webserver/views/api/v1/core.py
```diff
@@ -10,12 +10,14 @@
...
 from db.exceptions import NoDataFoundException, BadDataException
 from webserver.decorators import crossdomain
+from brainzutils.ratelimit import ratelimit
 
 bp_core = Blueprint('api_v1_core', __name__)
 
 
 @bp_core.route("/<uuid:mbid>/count", methods=["GET"])
 @crossdomain()
+@ratelimit()
```
alastair: I'd like to see at least one test to see that we're returning ratelimiting headers on one of these methods

manage.py
```diff
@@ -241,6 +241,21 @@ def toggle_site_status():
         print('Done!')
 
 
+@cli.command(name='set_rate_limits')
+@click.argument('per_ip')
+@click.argument('window_size')
```
alastair: still missing the type checks in these arguments

manage.py
```diff
@@ -241,6 +241,21 @@ def toggle_site_status():
...
+def set_rate_limits(per_ip, window_size):
+    """ Set rate limit parameters for the AcousticBrainz webserver
+
+    Args:
+        per_ip (int): the number of requests allowed per IP address
+        window_size (int): the window in number of seconds for how long the limit is applied
+    """
+    if per_ip / float(window_size) <= 1:
```
alastair: this doesn't have the full functionality of the original script - it doesn't print current limits if no arguments are given

webserver/views/api/legacy.py
```diff
@@ -51,6 +55,7 @@ def get_high_level(mbid):
 
 
 @api_legacy_bp.route("/<uuid:mbid>/low-level", methods=["POST"])
+@ratelimit()
```
alastair: I wonder if we should skip this one for now - this will come from our existing submission apps, which don't have any support for ratelimiting.
New versions of these apps should use the /api/v1 endpoint and support ratelimit responses.
If we decide to leave the limit in place we should be OK, since we wait until we calculate features before submitting it, and computation can take 20-30 seconds. Thoughts?

webserver/views/api/v1/test/test_core.py
```diff
@@ -35,6 +35,10 @@ def test_get_low_level(self):
 
         resp = self.client.get("/api/v1/%s/low-level" % mbid)
         self.assertEqual(resp.status_code, 200)
+        self.assertIn('X-RateLimit-Limit', resp.headers)
```
alastair: are we able to test specific values? or is that functionality already tested in BU?

webserver/views/api/legacy.py
paramsingh: I think skipping for now should be ok, if we still have problems after deploying this and there's a lot of requests to this, then we'll look into ratelimiting this as well.

webserver/views/api/v1/test/test_core.py
paramsingh: Tested here: https://github.com/metabrainz/brainzutils-python/commit/3738c96d669e678e2ab5b05faa7b28c9526f8d11#diff-3f0b606a8b8d994781dff384af815f83R14

PR #346, opened 2019-06-06
Set the maximum bulk query number to 25 and add documentation

alastair:
This number tries to seek a balance between a good number of items and the time taken to respond to the query. Tests show that 1 item takes about .3s while 25 take .8s and 50 takes over 1.5s

Also document this number in the API docs.

### DIFF
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -14,6 +14,16 @@ from webserver.decorators import crossdomain
 bp_core = Blueprint('api_v1_core', __name__)
 
 
+# If this value is increased, ensure that it still fits within the uwsgi header size:
+# https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html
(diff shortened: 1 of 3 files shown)


webserver/views/api/v1/core.py
```diff
@@ -203,6 +220,8 @@ def get_many_lowlevel():
       Takes the form `mbid[:offset];mbid[:offset]`. Offsets are optional, and should
       be >= 0
 
+      You can specify up to :py:const:`~webserver.views.api.v1.core.MAX_ITEMS_PER_BULK_REQUEST` MBIDs in a request.
```
paramsingh: noice!

alastair: this is kind of OK. It links to the definition elsewhere in the docs, but I would have preferred to make it put the actual number in the docstring.

alastair:
https://tickets.metabrainz.org/browse/AB-416

PR #347, opened 2019-06-07
AB-404: Endpoint for users to select specific lowlevel features

aidanlw17:
# AB-404: Endpoint for users to select specific lowlevel features

This feature allows users to select only a subset of the lowlevel data available when they do not require the whole file, and should help to reduce loads as a result. It makes this possible with the following changes:

- Create an endpoint, get_many_select_features which takes the required features as a parameter.
- Parse the features to get a string of paths that access different features of lowlevel.data.
- Query for these features in a similar fashion to other bulk get methods.

### DIFF
--- a/db/data.py
+++ b/db/data.py
@@ -598,15 +598,17 @@ def load_many_select_features(recordings, features, aliases):
     
     Args:
         recordings: A list of tuples (mbid, offset).
-        features: A string of paths to features 
-            "llj.data->'feature1' AS feature1, ..., llj.data->'featureN' AS featureN"
+        features: A list of string of paths to features. 
+            ["llj.data->'feature1', ..., llj.data->'featureN']
+        aliases: A list of aliases for querying the features.
+            ["lowlevel.feature1", ..., "metadata.featureN"]
 
     Returns:
-        {"mbid-1": {"offset-1": {"feature1": feature1, ..., "featureN": featureN},
+        {"mbid-1": {"offset-1": {lowlevel document},
                     ...
-                    "offset-n": {"feature1": feature1, ..., "featureN": featureN}},
+                    "offset-n": {lowlevel document},
          ...
-         "mbid-n": {"offset-1": {"feature1": feature1, ..., "featureN": featureN}}
+         "mbid-n": {"offset-1": {lowlevel document}}
         } 
     """
     # Build string of select features with aliases
@@ -634,7 +636,6 @@ def load_many_select_features(recordings, features, aliases):
             data = defaultdict(dict)
             for alias in aliases:
                 current = temp = {}
-                count = 0
                 alias_keys = alias.split('.')
                 for key in alias_keys[1:-1]:
                     temp[key] = {}
--- a/db/data.py
+++ b/db/data.py
@@ -593,7 +593,7 @@ def load_many_high_level(recordings, map_classes=False):
         return dict(recordings_info)
 
 
-def load_many_select_features(recordings, features):
+def load_many_select_features(recordings, features, aliases):
     """Load select features for multiple recordings.
     
     Args:
@@ -609,24 +609,39 @@ def load_many_select_features(recordings, features):
          "mbid-n": {"offset-1": {"feature1": feature1, ..., "featureN": featureN}}
         } 
     """
-    with db.engine.connect() as connection:
-        result = connection.execute(
-            "SELECT ll.gid::text, ll.submission_offset::text, %(features)s " 
-              "FROM lowlevel ll "
-              "JOIN lowlevel_json llj "
-                "ON ll.id = llj.id "
-             "WHERE (ll.gid, ll.submission_offset) "
-                "IN %(recordings)s" % {'recordings': tuple(recordings), 
-                                        'features': features})
+    # Build string of select features with aliases
+    feature_string = features[0] + ' AS "' + aliases[0] + '", '
+    for feature, alias in zip(features[1:len(features)-1], aliases[1:len(aliases)-1]):
+        feature_string += feature + ' AS "' +  alias + '", '
+    feature_string += features[len(features)-1] + ' AS "' + aliases[len(aliases)-1] + '"'
 
-        feature_names = result.keys()[2:]
-        features_info = {}
+    with db.engine.connect() as connection:
+        query = text("""
+            SELECT ll.gid::text
+                 , ll.submission_offset::text
+                 , %(features)s
+              FROM lowlevel ll
+              JOIN lowlevel_json llj
+                ON ll.id = llj.id
+             WHERE (ll.gid, ll.submission_offset)
+                IN :recordings
+        """ % {'features': feature_string})
+        result = connection.execute(query, {'recordings': tuple(recordings)})
+
         recordings_info = defaultdict(dict)
         for row in result.fetchall():
             # Build dictionary of feature columns
-            for name in feature_names:
-                features_info[name] = row[name]
-            recordings_info[row['gid']][row['submission_offset']] = features_info
+            data = defaultdict(dict)
+            for alias in aliases:
+                current = temp = {}
+                count = 0
+                alias_keys = alias.split('.')
+                for key in alias_keys[1:-1]:
+                    temp[key] = {}
+                    temp = temp[key]
+                temp[alias_keys[-1]] = row[alias]
+                data[alias_keys[0]].update(current)
+            recordings_info[row['gid']][row['submission_offset']] = data
 
         return dict(recordings_info)
 
--- a/db/data.py
+++ b/db/data.py
@@ -595,11 +595,11 @@ def load_many_high_level(recordings, map_classes=False):
 
 def load_many_individual_features(recordings, features):
     """Load individual lowlevel features for multiple recordings.
-    
+
     Args:
         recordings: A list of tuples (mbid, offset).
         features: Contains information about the
-        individual features, a list of tuples of the form: 
+        individual features, a list of tuples of the form:
             [(<feature_path>, <alias>, <default_type>), ...]
 
             <feature_path> is a string holding the path to a feature:
@@ -607,7 +607,7 @@ def load_many_individual_features(recordings, features):
 
             <alias> is a string alias for a feature:
             "lowlevel.feature_name"
-            
+
             <default_type> is the type to which a feature value will default
             if it is non-existent: None, {}, or [].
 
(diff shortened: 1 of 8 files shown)


webserver/views/api/v1/core.py
```diff
@@ -267,6 +267,93 @@ def get_many_highlevel():
...
+    raised stating the missing features message.
+    
+    Returns a string of feature paths with alias:
+        "llj.data->'feature1' AS feature1, ..., llj.data->'featureN' AS featureN"
+    """
+    features_param = request.args.get("features")
+    if not features_param:
+        raise webserver.views.api.exceptions.APIBadRequest("Missing `features` parameter")
+
+    selectable_features = ['lowlevel.average_loudness',
```
aidanlw17: @alastair I'm unsure of the best place to locate this hardcoded array of features. Maybe it should belong here, or perhaps it should be moved to the bottom of the file at the top level?

webserver/views/api/v1/core.py
```diff
@@ -267,6 +267,93 @@ def get_many_highlevel():
...
+            raw_paths.append(feature)
+            # Build feature path
+            feature_path = 'llj.data'
+            for element in feature.split('.'):
+                feature_path += '->\'' + element + '\''
+            parsed_features.append(feature_path)
+
+    # Remove duplicates, preserving order
+    seen = set()
+    parsed_features = [x for x in parsed_features if not (x in seen or seen.add(x))]
```
aidanlw17: @alastair I iterate over the features here twice, once to add them to this array to remove duplicates and then again to form the string. It would be better to only iterate once but I wasn't sure of a better approach with the need to remove duplicates?

webserver/views/api/v1/core.py
```diff
@@ -267,6 +267,93 @@ def get_many_highlevel():
...
+
+    features_string = parsed_features[0] + ' AS "' + raw_paths[0] + '", '
+    for feature, alias in zip(parsed_features[1:len(parsed_features)-1], raw_paths[1:len(raw_paths)-1]):
+        features_string += feature + ' AS "' +  alias + '", '
+    features_string += parsed_features[len(parsed_features)-1] + ' AS "' + raw_paths[len(raw_paths)-1] + '"'
```
aidanlw17: @alastair I had trouble passing the list of features into the sql query without concatenating them all into a string like this, and obviously the above is hard to follow. Is it preferable to use str.join(), or a different method to concatenate?

db/data.py
```diff
@@ -520,6 +520,44 @@ def load_many_high_level(recordings):
...
+                                        'features': features})
+
+        feature_names = result.keys()[2:]
+        features_info = {}
+        recordings_info = defaultdict(dict)
+        for row in result.fetchall():
+            # Build dictionary of feature columns
+            for name in feature_names:
+                features_info[name] = row[name]
+            recordings_info[row['gid']][row['submission_offset']] = features_info
```
alastair: Let's replicate the structure of the lowlevel document here instead of just adding it to a dictionary. This gives us the advantage of allowing clients who already access this data to update their query url and have the selection logic still work

webserver/views/api/v1/core.py
```diff
@@ -267,6 +267,93 @@ def get_many_highlevel():
...
+                            'rhythm.bpm_histogram_first_peak_bpm.mean', 
+                            'rhythm.bpm_histogram_second_peak_bpm.mean', 
+                            'rhythm.danceability', 
+                            'rhythm.onset_rate', 
+                            'tonal.chords_key', 
+                            'tonal.chords_scale', 
+                            'tonal.key_key', 
+                            'tonal.key_scale', 
+                            'tonal.tuning_frequency', 
+                            'tonal.tuning_equal_tempered_deviation']
```
alastair: let's also add metadata.tags as an option here, and also *always* include metadata.version and metadata.audio_properties in all responses.

webserver/views/api/v1/core.py
```diff
@@ -267,6 +267,93 @@ def get_many_highlevel():
...
+       }
+    
+    :query recording_ids: *Required.* A list of recording MBIDs to retrieve.
+        
+        Takes the form `mbid[:offset];mbid[:offset]`. Offsets are optional, and should
+        be >= 0
+    
+    :query features: *Required.* A list of features to be returned for each mbid.
+        
+        Takes the form `feature1;feature2`.
```
alastair: we should list the features here. One way you could do this is like I did for the max number of bulk queries: 
https://github.com/metabrainz/acousticbrainz-server/blob/cc359f8ea5c580fbdb0e0a1c0bb71004ca4f380c/webserver/views/api/v1/core.py#L260

webserver/views/api/v1/core.py
```diff
@@ -267,6 +267,93 @@ def get_many_highlevel():
...
+        features_string += feature + ' AS "' +  alias + '", '
+    features_string += parsed_features[len(parsed_features)-1] + ' AS "' + raw_paths[len(raw_paths)-1] + '"'
+
+    return features_string
+
+
+@bp_core.route("/low-level/select", methods=["GET"])
+@crossdomain()
+def get_many_select_features():
+    """Get a specified subset of low-level data for many recordings at once.
```
alastair: make sure we also say here what happens if no data exists for a mbid/offset (we skip it and it's not returned in the data). Do we have that for the other bulk queries? If not we should add it there too.

db/data.py
```diff
@@ -520,6 +520,44 @@ def load_many_high_level(recordings):
...
+        {"mbid-1": {"offset-1": {"feature1": feature1, ..., "featureN": featureN},
+                    ...
+                    "offset-n": {"feature1": feature1, ..., "featureN": featureN}},
+         ...
+         "mbid-n": {"offset-1": {"feature1": feature1, ..., "featureN": featureN}}
+        } 
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(
+            "SELECT ll.gid::text, ll.submission_offset::text, %(features)s "
```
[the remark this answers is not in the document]
alastair: and use """ strings so that we only need to quote at the beginning and end

webserver/views/api/v1/core.py
```diff
@@ -267,6 +289,97 @@ def get_many_highlevel():
...
+    metadata_version_alias = "metadata.version"
+    metadata_audio_properties = "llj.data->'metadata'->'audio_properties'"
+    metadata_audio_properties_alias = "metadata.audio_properties"
+    
+    parsed_features.append((metadata_version, metadata_version_alias, {}))
+    parsed_features.append((metadata_audio_properties, metadata_audio_properties_alias, {}))
+
+    # Remove duplicates, preserving order
+    ret = []
+    return [x for x in parsed_features if not (x in ret or ret.append(x))]
```
aidanlw17: @alastair although we added the remove duplicates utility, that method did not work once I added support for default types - some of the default types were {}, and they are unhashable so it didn't work with the usage of set(). Is this a suitable method instead?

webserver/views/api/v1/test/test_core.py
```diff
@@ -274,14 +274,132 @@ def test_get_bulk_hl_absent_mbid(self, load_many_high_level):
...
+        self.assertEqual(resp.status_code, 200)
+
+        # We expect that we get returned whatever we set the mock to, but the argument
+        # to load_many_high_level is always lower-case regardless of what we pass in
+        self.assertEqual(resp.json, expected_result)
+        recordings = [("c5f4909e-1d7b-4f15-a6f6-1af376bc01c9", 0)]
+        features = [("llj.data->'lowlevel'->'average_loudness'", "lowlevel.average_loudness", None),
+                    ("llj.data->'metadata'->'version'", "metadata.version", {}),
+                    ("llj.data->'metadata'->'audio_properties'", "metadata.audio_properties", {})]
+        load_many_select_features.assert_called_with(recordings, features)
```
aidanlw17: @alastair this assertion will fail until PR #349 adds the conversion to lowercase in the _parse_bulk_params method

webserver/views/api/v1/core.py
alastair: OK, if you need to use the same functionality but in a different form then that's fine to duplicate it

webserver/views/api/v1/test/test_core.py
alastair: Is this because you proactively added a test for this specific case, even though the previous PR hasn't been merged yet? This is interesting. Normally I would say that this specific PR should be standalone, and that you shouldn't write a test that intentionally fails, but you're right that it's a good idea to have this test eventually, and we need to keep track somehow that it's necessary.
One thing we could do here is mark this test function with `@unittest.skip` so that it's not run, and make a note that it has to be un-skipped when the other is merged.

aidanlw17: Yes I added it proactively because once we merge #349 we will need it anyways, so my thought was that this would be easier than adding a new PR just for the test once this one and #349 are already merged? I will use `@unittest.skip` and add a comment so we can keep track of it for now.

webserver/views/api/v1/test/test_core.py
```diff
@@ -274,14 +274,134 @@ def test_get_bulk_hl_absent_mbid(self, load_many_high_level):
...
+        # Only features parameter missing
+        recordings = self.test_recording1_mbid + ';' + self.test_recording2_mbid
+        resp = self.client.get('/api/v1/low-level/select?recording_ids=%s' % recordings)
+        self.assertEqual(400, resp.status_code)
+
+        expected_result = {"message": "Missing `features` parameter"}
+        self.assertEqual(expected_result, resp.json)
+
+    # Once PR #349 adds conversion to lowercase in _parse_bulk_params, this skip should be removed
+    @unittest.skip('Conversion to lowercase in _parse_bulk_params not yet added via PR #349.')
```
aidanlw17: @alastair this now uses unitttest.skip until #349 is merged - sorry to forget about adding this until now.

db/data.py
```diff
@@ -520,6 +520,71 @@ def load_many_high_level(recordings):
         return dict(recordings_info)
 
 
+def load_many_select_features(recordings, features):
```
alastair: load_many_individual_features

db/data.py
```diff
@@ -520,6 +520,71 @@ def load_many_high_level(recordings):
...
+                ON ll.id = llj.id
+             WHERE (ll.gid, ll.submission_offset)
+                IN :recordings
+        """ % {'features': feature_string})
+        result = connection.execute(query, {'recordings': tuple(recordings)})
+
+        recordings_info = defaultdict(dict)
+        for row in result.fetchall():
+            # Build dictionary of feature columns
+            data = defaultdict(dict)
```
alastair: I would break this indented section into a separate function (input row, output data), and then have some tests for it

webserver/views/api/v1/core.py
```diff
@@ -267,6 +289,97 @@ def get_many_highlevel():
     return jsonify(recording_details)
 
 
+def parse_select_features():
```
alastair: again, _select_

utils/remove_duplicates.py
```diff
@@ -0,0 +1,3 @@
+def remove_duplicates(arr):
```
alastair: I'd name this file something else - when I see remove_duplicates I think of duplicate submissions. How about `utils/container_utils.py`? (since it's currently got utils for sets, perhaps we might add other things for other container types too)

alastair:
@aidanlw17 is this finished now with the spelling changes you made? Can you rebase to fix the merge conflicts?

aidanlw17:
@alastairp not quite finished, I had another set of commits with a refactor that I didn’t push yet. Sorry! I will get to it either today or tomorrow and then can rebase.

aidanlw17:
@alastair This is now good to go! I refactored parts of db/data/load_many_individual_features into smaller functions and added unit tests for them.

alastair:
I just rebased on master to take into account some improvements in master. Working on final tests on this now

alastair:
oh cool! I didn't realise that pep8speaks would talk on this too! Let's skip it for now - I still need to configure some things like line length, then I was planning on doing a single commit to fix all code layout in the whole project so we can start fresh

alastair:
After some discussion in IRC, we decided to merge this functionality into the bulk lowlevel api, that is, `/api/v1/low-level`. I'll make this change and update the tests and documentation.

PR #348, opened 2019-06-10
AB-91: Map model classes to more meaningful labels

alastair:
A continuation of #330 with a few additional changes to improve the code flow and number of queries that we make.
Also add some more mappings and improve the way that we display this data on the summary page.

### DIFF
--- a/db/data.py
+++ b/db/data.py
@@ -312,17 +312,15 @@ def set_model_status(model_name, model_version, model_status):
                             "model_status": model_status})
 
 
-def get_model(model_name, model_version):
+def get_active_models():
     with db.engine.begin() as connection:
         query = text(
             """SELECT *
                  FROM model
-                WHERE model = :model_name
-                  AND model_version = :model_version""")
+                WHERE status = :model_status""")
         result = connection.execute(query,
-                                    {"model_name": model_name,
-                                     "model_version": model_version})
-        return result.fetchone()
+                                    {"model_status": STATUS_SHOW})
+        return [dict(row) for row in result.fetchall()]
 
 
 def _get_model_id(model_name, version):
@@ -471,12 +469,34 @@ def load_many_low_level(recordings):
         return dict(recordings_info)
 
 
-def load_high_level(mbid, offset=0):
+def map_highlevel_class_names(highlevel, mapping):
+    """Convert class names from the classifier output to human readable names.
+
+    Arguments:
+        highlevel (dict): highlevel data dict containing shortened keys
+        mapping (dict): a mapping from class names -> human readable names
+
+    Returns:
+        the highlevel input with the keys of the `all` item, and the `value` item
+        changed to the values from the provided mapping
+    """
+
+    new_all = {}
+    for cl, val in highlevel["all"].items():
+        new_all[mapping[cl]] = val
+    highlevel["all"] = new_all
+    highlevel["value"] = mapping[highlevel["value"]]
+
+    return highlevel
+
+
+def load_high_level(mbid, offset=0, map_classes=False):
     """Load high-level data for a given MBID.
 
     Arguments:
         mbid (str): MBID to load
         offset (int): submission offset for this MBID, starting from 0
+        map_classes (bool): if True, map class names to human readable values in the returned data
 
     Raises:
         NoDataFoundException: if this mbid doesn't exist or the offset is too high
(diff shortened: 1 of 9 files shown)


admin/sql/create_models.sql
```diff
@@ -19,4 +19,96 @@ INSERT INTO model (model, model_version, date, status) VALUES ('timbre', 'v2.1_b
...
 INSERT INTO model (model, model_version, date, status) VALUES ('voice_instrumental', 'v2.1_beta1', now(), 'show');
 
+UPDATE model SET class_mapping = '{
+        "danceable": "Danceable",
+        "not_danceable": "Not danceable"
+    }'::jsonb WHERE model.model = 'danceability';
+
+UPDATE model SET class_mapping = '{
+        "acoustic": "Acoustic",
+        "not_acoustic": "Not acoustic"
```
alastair: I also updated these items so that we don't have to automatically remove _ from them.

admin/sql/create_models.sql
```diff
@@ -19,4 +19,96 @@ INSERT INTO model (model, model_version, date, status) VALUES ('timbre', 'v2.1_b
...
+        "not_relaxed": "Not relaxed"
+    }'::jsonb WHERE model.model = 'mood_relaxed';
+
+UPDATE model SET class_mapping = '{
+        "sad": "Sad",
+        "not_sad": "Not sad"
+    }'::jsonb WHERE model.model = 'mood_sad';
+
+UPDATE model SET class_mapping = '{
+        "Cluster1": "passionate, rousing, confident, boisterous, rowdy",
```
alastair: These cluster descriptions are listed here: https://acousticbrainz.org/datasets/accuracy#moods_mirex

db/data.py
```diff
@@ -265,17 +265,15 @@ def set_model_status(model_name, model_version, model_status):
                             "model_status": model_status})
 
 
-def get_model(model_name, model_version):
```
alastair: this method wasn't used anywhere, so I hijacked the implementation to make a method that we needed

db/data.py
```diff
@@ -424,31 +422,54 @@ def load_many_low_level(recordings):
...
+        mapping (dict): a mapping from class names -> human readable names
+
+    Returns:
+        the highlevel input with the keys of the `all` item, and the `value` item
+        changed to the values from the provided mapping
+    """
+
+    new_all = {}
+    for cl, val in highlevel["all"].items():
+        new_all[mapping[cl]] = val
```
alastair: this assumes that the mapping has all possible values in it. Since we control the both the models and the mapping definition I didn't bother checking it here

webserver/views/data.py
```diff
@@ -195,72 +195,85 @@ def get_tag(name):
         return {}
 
 
-def _interpret_high_level(hl):
+def _interpret_high_level(hl, models):
 
-    def interpret(text, data, threshold=.6):
-        if data['probability'] >= threshold:
```
alastair: I removed this threshold because it didn't make much sense with models that had many classes - e.g. with 6 classes, 40% is much higher than the random baseline (16%)

alastair:
@paramsingh an eye over this would be appreciated, thanks

admin/sql/create_models.sql
```diff
@@ -19,4 +19,96 @@ INSERT INTO model (model, model_version, date, status) VALUES ('timbre', 'v2.1_b
 INSERT INTO model (model, model_version, date, status) VALUES ('tonal_atonal', 'v2.1_beta1', now(), 'show');
 INSERT INTO model (model, model_version, date, status) VALUES ('voice_instrumental', 'v2.1_beta1', now(), 'show');
 
+UPDATE model SET class_mapping = '{
```
paramsingh: Is there a reason why we're not inserting mappings for `gender`, `voice_instrumental`, `timbre` and `ismir04_rhythm`? Would that json be trivial? I think it might be a good idea to make the `class_mapping` column NOT NULL and insert mappings regardless?

paramsingh: Then we could also just put this inside the insert statement which might look a bit cleaner?

db/data.py
```diff
@@ -424,31 +422,54 @@ def load_many_low_level(recordings):
         return dict(recordings_info)
 
 
-def load_high_level(mbid, offset=0):
+def map_highlevel_class_names(highlevel, mapping):
+    """Convert class names from the classifier output to human readable names.
+
+    Arguments:
+        highlevel (dict): highlevel data dict containing shortend keys
```
paramsingh: typo - shortened

db/data.py
```diff
@@ -265,17 +265,15 @@ def set_model_status(model_name, model_version, model_status):
                             "model_status": model_status})
 
 
-def get_model(model_name, model_version):
+def get_active_models():
```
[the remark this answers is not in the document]
alastair: I originally had this method for "future use" and we never used it. I'd prefer to implement stuff as we need it, and if we need a variation in the future that we make the change then

admin/sql/create_models.sql
alastair: The original idea was to turn class codes -> real text. In my view, the ones that you said as examples already have names that are nicely readable.
In the future, when we release models built from the dataset editor, I expect that we'll name the classes as nicely readable words (there's no reason that we have to use specific short codes). This means that I expect that we won't need a mapping for any models going forward from here, it's just a quick measure for the existing models that we have.

db/data.py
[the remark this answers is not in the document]
paramsingh: ok, sounds good.

admin/sql/create_models.sql
paramsingh: Alright, makes sense.

PR #349, opened 2019-06-13
AB-411: Fix exceptions when uuids have upper-case characters

alastair:
https://tickets.metabrainz.org/browse/AB-411

With the changes in AB-406 we introduced a bug where GET for a single item failed because postgresql returned mbids in lower-case and we tried to use the upper/mixed case value that was passed as a parameter.

Additionally, the Flask-uuid extension only accepts lower-case uuids by default, meaning that some of our API endpoints returned HTTP404 when presented with upper-case MBIDs, even if they existed in the database.

Convert all uuids to lower-case when returning them, and document this. Accept API queries using upper-case uuids.

I made no changes to the internal datasets api because we always use uuids returned from the api/postgres here, which are always lower-case.

### DIFF
--- a/db/test/test_data.py
+++ b/db/test/test_data.py
@@ -127,6 +127,21 @@ class DataDBTestCase(DatabaseTestCase):
         self.assertEqual(one, db.data.load_low_level(self.test_mbid, 0))
         self.assertEqual(two, db.data.load_low_level(self.test_mbid, 1))
 
+    def test_load_low_level_uuid_case(self):
+        """A query with an upper-case uuid will return the correct data"""
+        one = {"data": "one",
+               "metadata": {"audio_properties": {"lossless": True}, "version": {"essentia_build_sha": "x"}}}
+        db.data.write_low_level(self.test_mbid, one, gid_types.GID_TYPE_MBID)
+
+        self.assertEqual(one, db.data.load_low_level(self.test_mbid.upper()))
+
+        many_ll_expected = {
+            self.test_mbid: {'0': one}
+        }
+        recordings = [(self.test_mbid.upper(), 0)]
+
+        self.assertEqual(many_ll_expected, db.data.load_many_low_level(list(recordings)))
+
     def test_load_low_level_none(self):
         """If no lowlevel data is loaded, or offset is too high, an exception is raised"""
         with self.assertRaises(db.exceptions.NoDataFoundException):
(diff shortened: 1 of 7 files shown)


webserver/__init__.py
```diff
@@ -81,10 +81,6 @@ def create_app(debug=None):
     else:
         raise Exception('One or more redis cache configuration options are missing from config.py')
 
-    # Extensions
-    from flask_uuid import FlaskUUID
```
alastair: this wasn't needed because it was already done in the `use_flask_uuid` argument to `CustomFlask`

db/test/test_data.py
```diff
@@ -127,6 +127,21 @@ def test_load_low_level_offset(self):
...
+        db.data.write_low_level(self.test_mbid, one, gid_types.GID_TYPE_MBID)
+
+        self.assertEqual(one, db.data.load_low_level(self.test_mbid.upper()))
+
+        many_ll_expected = {
+            self.test_mbid: {'0': one}
+        }
+        recordings = [(self.test_mbid.upper(), 0)]
+
+        self.assertEqual(many_ll_expected, db.data.load_many_low_level(list(recordings)))
```
aidanlw17: We recognize here https://github.com/metabrainz/acousticbrainz-server/pull/349/files#diff-9486bd3dcb5080bd1fabd45af163d4b1R194 that only lowercase uuids are ever passed as an argument to load_many_low_level and load_many_high_level. So why do we test them here by passing in an MBID converted to uppercase? If an uppercase MBID is passed directly into one of the load_many methods it is never converted to lowercase, so shouldn't this result in an error? 

We call on the load_many methods from the load_low_level and load_high_level as well, so if the uppercase mbid doesn't result in an error at this assertion, why do we bother to convert to lowercase below?
https://github.com/metabrainz/acousticbrainz-server/pull/349/files#diff-973d6aba9f5f7b7acda5e7780e0c45a7R380

alastair: It's more about being double-sure that no matter what you pass each method, the right thing will happen. We could either document the methods and say "ensure that you pass a lower-case item" to this, or we could just lower-case it every time we need to

aidanlw17: Okay I see. How come we don't test the load_many_low_level and load_many_high_level with uppercase MBIDs?

alastair: What do you mean? This test method is testing both `db.data.load_low_level` and `db.data.load_many_low_level`

aidanlw17: That was not phrased correctly, sorry. Will there be an error if we pass an uppercase MBID into `db.data.load_many_low_level`?

alastair:
What do you mean mixed case? (as opposed to purely upper-case? I don't think there's a problem having only one, besides, there are _some_ mixed-case ones here: https://github.com/metabrainz/acousticbrainz-server/pull/349/files#diff-9486bd3dcb5080bd1fabd45af163d4b1R280)

PR #353, opened 2019-06-19
AB-412: Compute recording similarity metrics on database

aidanlw17:
# AB-412: Compute recording similarity metrics on database

The first step in adding similarity to AB is to compute metrics for measuring similarity between recordings, since these measurements can later be indexed. This PR will add classes for the metrics and scripts to compute them for the entire database of recordings.

##### 'similarity' Table
- Holds computed vectors for each metric, for each recording in the database.
- These vectors will be added to the indices for querying.

##### 'similarity_metrics' Table
- Holds metadata about each metric that has been computed for the database

##### 'similarity_stats' Table
- Hold statistical information, mean and standard deviation of each computed metric.
- These statistics may be useful for representations at a later stage in the project.

### Metrics Classes
- Adds classes with information about each metric and methods for it to be created, deleted, and 
  computed for a recording.
- Includes formation of hybrid metrics, useful at a later stage in the project.

### Incremental Computation of Metrics
- After the similarity table has all recordings added to it using the cli command 
  `./develop.sh run --rm webserver python2 manage.py similarity init`, then a metric
   can be computed on the entire database incrementally using the following command:
  `./develop.sh run --rm webserver python2 manage.py similarity add-metric <metric_name>`
- For now, the metric must exist within the list of base metrics - later, hybrid metrics will allow for this 
  functionality to change.
- Also includes cli commands for adding and deleting hybrid metrics, useful at a later stage in the 
  project.

Much of this work is modified from the following thesis authored by Philip Tovstogan:
https://doi.org/10.5281/zenodo.1479769


similarity/manage.py
```diff
@@ -0,0 +1,147 @@
...
+
+from sqlalchemy import text
+
+PROCESS_BATCH_SIZE = 10000
+
+cli = FlaskGroup(add_default_commands=False, create_app=webserver.create_app_flaskgroup)
+
+
+@cli.command()
+def init_similarity():
```
alastair: can you add some documentation to all commands for the output of `—help`? If this is a subcommand, we perhaps don’t need similarity in the name - e.g. `python manage.py similarity init` would be fine as a command

similarity/manage.py
```diff
@@ -0,0 +1,147 @@
...
+    click.echo('Copying data')
+    with db.engine.begin() as connection:
+        query = text("""
```
alastair: we shouldn’t have sql in a management command. Can you move it (and other stuff in this file) into `db/similarity.py`?

admin/updates/20190619-similarity.sql
```diff
@@ -0,0 +1,40 @@
...
+
+ALTER TABLE similarity ADD CONSTRAINT similarity_pkey PRIMARY KEY (id);
+
+ALTER TABLE similarity
+  ADD CONSTRAINT similarity_fk_lowlevel
+  FOREIGN KEY (id)
+  REFERENCES lowlevel (id);
+
+CREATE TABLE similarity_metrics (
+  metric TEXT, -- PK
```
alastair: I’d like to know the size of these tables (because this is going to have an impact on the full database size). Can you run some disk usage statistics on your subset for the lowlevel table and these tables when they’re full? That way we should be able to say something like “the similarity tables will be x% of the lowlevel tables”

similarity/manage.py
```diff
@@ -0,0 +1,147 @@
...
+    query = text("""
+        UPDATE similarity
+           SET %(metric)s = %(value)s
+         WHERE id = %(id)s
+    """ % {'metric': name, 'value': 'ARRAY' + value, 'id': row_id})
+    connection.execute(query)
+
+
+@cli.command(name="add-metric")
+@click.argument("name")
```
alastair: is there a way of adding all metrics at once without running this command many times?

similarity/operations.py
```diff
@@ -0,0 +1,102 @@
...
+                 VALUES (:metric, :hybrid, :description, :category, TRUE)
+            ON CONFLICT (metric)
+          DO UPDATE SET visible=TRUE
+        """)
+        self.connection.execute(metrics_query, {'metric': self.name,
+                                                'hybrid': hybrid,
+                                                'description': self.description,
+                                                'category': self.category})
+
+        # This can be removed if not using postgres similarity solution
```
alastair: we don’t need to keep commented code - we have a git history available which we can restore from if needed. you can delete this

similarity/manage.py
```diff
@@ -0,0 +1,147 @@
...
+        connection.execute(query)
+
+
+def _get_recordings_without_similarity(connection, name, batch_size):
+    query = text("""
+        SELECT id
+          FROM similarity
+         WHERE %(metric)s IS NULL
+         LIMIT %(limit)s
+    """ % {'metric': name, 'limit': batch_size})
```
alastair: let’s fix all of these queries to use sqlalchemy query parameters instead of string interpolation

alastair:
One more thing - I’d like to start being a bit more careful about the documentation that we include with AB. Let’s start adding some documentation about how similarity works. It could be a README.md file in the `similarity/` folder that gives an outline of what it is, how to start running it, and how things are stored. I’m not too worried if this is added in this PR or in #354 (since we do more stuff there and so we can add more detail)

similarity/manage.py
```diff
@@ -0,0 +1,147 @@
...
+    result = connection.execute(query)
+    rows = result.fetchall()
+    if not rows:
+        return []
+    ids = zip(*rows)[0]
```
alastair: this should be `[r[‘id’] for r in rows]` for readability

alastair: and it also means we don’t need the `if not rows` check

aidanlw17:
@alastair For some reason, I seem to have missed your comment about the README.md file before. I think it would be a great idea to add something like that. I think your idea to include it in #354 might be best - once annoy is being added as well I think the big picture will be more clear, and I can also be more specific once we have reviewed that work.

similarity/manage.py
aidanlw17: I wrote this query in a different way, so I was able to remove this function entirely. I mainly altered it so we could add row by row, with all the metrics at the same time. Doing it that way, it didn't make sense to query for recordings like this anymore (plus, we set the metrics columns as not nullable).

admin/updates/20190619-similarity.sql
aidanlw17: This is a pastebin with all of the general table size information, should I extract anything more specific to the similarity and lowlevel tables, or is this what you had in mind?
https://pastebin.com/Wxg7fBKV

aidanlw17: I think this https://pastebin.com/KxLFKZMS also gives a more simple and clear depiction.

aidanlw17: With that in mind, the similarity table is about 9 times as large as the lowlevel_json table.

alastair: The json data in the lowlevel_json table is actually stored in `pg_toast.pg_toast_196767`. This is a side effect of the way that postgres stores the data (the ‘total’ column of the first paste). So we’re actually looking at about 30x smaller for the similarity table. This means that this table will be able 15gb in size with 450gb of items.

aidanlw17: Aha... I had a feel that didn’t really make much sense😅. So does this impact our full database size in a way that requires us to make any changes?

db/stats.py
```diff
@@ -73,6 +75,51 @@ def get_last_submitted_recordings():
...
+        """)
+        result = connection.execute(query, {"metric": metric_name})
+        return result.fetchone()
+
+
+def insert_similarity_stats(metric_name, means, stddevs):
+    with db.engine.connect() as connection:
+        query = text("""
+            INSERT INTO similarity_stats (metric, means, stddevs)
+                 VALUES (%(metric)s, %(means)s, %(stddevs)s)
```
alastair: when inserting values you can use query parameters as well, instead of string interpolation - there are a few places where this is still used

db/similarity.py
```diff
@@ -0,0 +1,239 @@
...
+    with db.engine.connect() as connection:
+        query = text("""
+            SELECT COUNT(*)
+              FROM similarity
+        """)
+        result = connection.execute(query)
+        return result.fetchone()[0]
+
+
+def submit_similarity_by_id(id):
```
aidanlw17: @alastair This function and the one below are meant to submit only a single row, this will be necessary when we need to update the table with new submissions.

db/stats.py
aidanlw17: I've been investigating these things for the last week and for some of them I haven't been able to successfully complete the query with string interpolation, e.g. here https://pastebin.com/D0aQBj9y

I will comment some other errors that I'm getting with them as well.

db/similarity.py
```diff
@@ -0,0 +1,239 @@
...
+def submit_similarity_by_mbid(mbid, offset):
+    """Computes similarity metrics for a single recording specified
+    by (mbid, offset) combination, then inserts the metrics as a new
+    row in the similarity table."""
+    id = db.data.get_lowlevel_id(mbid, offset)
+    submit_similarity_by_id(id)
+
+
+# TODO: Write tests for these and for stats, also add docstrings.
+def insert_similarity_meta(metric, hybrid, description, category):
```
aidanlw17: @alastair for methods such as this and the four below it, which use sql queries for inserting, updating, and altering, how can I verify their success in a unit test?

alastair:
Just as a reminder for us when we deploy: when I was looking at this I found commit 2aab1252d4d5723ed71f0e81b3bc82a4b1c0a29e which added a 50mb sql dump (which was then deleted a few commits later). We don't want this file in the history, so we should rewrite the history to remove it before we merge. I'm happy to do this myself.

db/similarity.py
```diff
@@ -0,0 +1,193 @@
...
+    with db.engine.connect() as connection:
+        query = text("""
+            SELECT COUNT(*)
+              FROM similarity
+        """)
+        result = connection.execute(query)
+        return result.fetchone()[0]
+
+
+def submit_similarity_by_id(id, metrics=None):
```
aidanlw17: @alastair I've refactored and simplified this method, and written some tests for it using mocked classes - I felt like this was a good approach to the tests, but can you confirm? Additionally, this function no longer queries the database itself. Should I move it somewhere like similarity/utils?

db/similarity.py
```diff
@@ -0,0 +1,133 @@
...
+    # Check that lowlevel submission exists for given id
+    if not db.data.check_for_submission(id):
+        raise NoDataFoundException('No submission for parameter `id`.')
+
+    if not metrics:
+        metrics = similarity.utils.init_metrics()
+
+    vectors_info = []
+    for metric in metrics:
+        data = metric.get_data(id)
```
alastair: I'm sorry that I didn't catch this part earlier - I only just saw it now that I'm running through the whole insertion process.
This step is really inefficient because we're calling get_data once for each metric, but using exactly the same id. This means that for each id in the database we're running:
* 12 queries to get data for each metric
* 1 query to validate the id
* 1 query to do the actual insertion

multiplied by 10 million rows, this is going to take a _really_ long time.
We should change this so that we get the lowlevel data in a single sql query, and then each metric acts on the returned json data and selects the field that we need. This means we can only do one query per id.
We could leave the get_data functionality in case we need it in the future. We can add another method to each metric.

A further improvement would be to select all lowlevel items for a batch in a single query (e.g. 1 query instead of 1000), but that could be done as a later improvement

db/similarity.py
```diff
@@ -0,0 +1,144 @@
...
+    lowlevel_count = count_all_lowlevel()
+
+    with db.engine.connect() as connection:
+        metrics = similarity.utils.init_metrics()
+        sim_count = count_similarity()
+        print("Processed {} / {} ({:.3f}%)".format(sim_count,
+                                                   lowlevel_count,
+                                                   float(sim_count) / lowlevel_count * 100))
+        
+        batch_query = text("""
```
aidanlw17: @alastair we need to get both the low and high level data. I altered this query to collect it all in bulk. I think it has made our querying much faster - it takes ~28.5 seconds to do a 10 000 recording batch, so near 10h for the whole 12m recording database. I think this is an alternative approach to the query as well:

```
            SELECT id, llj.data AS ll_data, jsonb_object_agg(hl.model, hl.data) AS hl_data
              FROM lowlevel
         LEFT JOIN lowlevel_json AS llj
             USING (id)
         LEFT JOIN highlevel_model AS hl
             USING (id)
         LEFT JOIN similarity
             USING (id)
             WHERE similarity.id IS NULL GROUP BY (id, llj.data)
             LIMIT :batch_size
```
I used explain and tested both, the one that I implemented seems slightly less complex and took about 2 seconds less per batch.

A further improvement could be using some code from #347 to select only some features rather than the whole lowlevel document, which we don't need.

db/similarity.py
```diff
@@ -0,0 +1,144 @@
...
+             LIMIT :batch_size
+        """)
+
+        while True:
+            with connection.begin():
+                result = connection.execute(batch_query, {"batch_size": batch_size})
+                if not result.rowcount:
+                    break
+
+                row = result.fetchone()
```
aidanlw17: I had an issue here that was unexpected, using result.fetchall() gave odd behaviour where it took very long to fetch, then eventually timed out (or something like that), and the only way to escape it seemed to be to close the terminal session and restart docker. I thought this might be because of the size of the rows? Fetching one row at a time works without the issue.

db/similarity.py
```diff
@@ -0,0 +1,144 @@
...
+    similarity table."""
+    try:
+        id = int(id)
+    except ValueError:
+        raise BadDataException('Parameter `id` must be an integer.')
+
+    if not metrics:
+        metrics = similarity.utils.init_metrics()
+
+    if not data:
```
aidanlw17: I added this so we can use the same function when we compute metrics for a new recording submitted to the db.

db/test/test_similarity.py
```diff
@@ -0,0 +1,120 @@
...
+
+import mock
+
+import db
+import db.similarity
+import db.exceptions
+from db.testing import DatabaseTestCase, TEST_DATA_PATH, gid_types
+
+
+class SimilarityDBTestCase(DatabaseTestCase):
```
aidanlw17: I'll fix the tests in here once we get some of the other things in this review solidified!

db/submission_stats.py
```diff
@@ -39,6 +39,8 @@
     LOWLEVEL_TOTAL_UNIQUE: "Total (unique)",
 }
 
+NORMALIZATION_SAMPLE_SIZE = 10000
```
alastair: I think this should be in the other file?

aidanlw17: Good catch! Thank you.

aidanlw17: Turns out it just needs to be removed from this one

similarity/metrics.py
```diff
@@ -0,0 +1,290 @@
...
+from operations import BaseMetric
+
+NORMALIZATION_SAMPLE_SIZE = 10000
+
+
+class LowLevelMetric(BaseMetric):
+    path = ''
+    indices = None
+
+    def get_data(self, id):
```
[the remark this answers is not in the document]
alastair: sorry that I didn't reply to this - one easy way would be to just mock the call to db.data, as long as that method also has a test, then this would be enough for the test

similarity/metrics.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def length(self):
+        return len(self.indices) if self.indices else 1
+
+
+class NormalizedLowLevelMetric(LowLevelMetric):
+    means = None
+    stddevs = None
+
+    def calculate_stats(self):
```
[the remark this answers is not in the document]
alastair: I think it's good that we moved the sql into db.similarity, thanks

similarity/metrics.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def length(self):
+        return len(self.models)
+
+
+class MoodsMetric(BinaryCollectiveMetric):
+    name = 'moods'
+    description = 'Moods'
+    models = [
+        11,  # mood_happy
```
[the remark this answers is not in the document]
alastair: My idea was always to use the name of the model to identify it. is there any reason that you can see why we can't select from the model table by the name instead of the id?

similarity/utils.py
```diff
@@ -0,0 +1,16 @@
...
+def init_metrics():
+    """Initializes and creates all base metrics, returning
+    their instances as a list."""
+    metrics = []
+    for name in similarity.metrics.BASE_METRICS:
+        metric_cls = similarity.metrics.BASE_METRICS[name]
+        metric = metric_cls()
+        metrics.append(metric)
+        try:
+            metric.calculate_stats()
```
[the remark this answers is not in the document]
alastair: we talked about moving this statistics calculation into an init method

db/data.py
```diff
@@ -593,6 +593,107 @@ def load_many_high_level(recordings, map_classes=False):
         return dict(recordings_info)
 
 
+def get_lowlevel_metric_feature(id, path):
```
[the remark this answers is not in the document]
aidanlw17: It's possible to remove this function, the associated tests, and the get_data function in the LowLevelMetric class. However, we discussed leaving this functionality incase we use it in the future. We have the git history, and I think it can be removed without too much concern.

db/data.py
```diff
@@ -593,6 +593,107 @@ def load_many_high_level(recordings, map_classes=False):
...
+        """)
+        result = connection.execute(query, {"id": id})
+        if not result.rowcount:
+            return None
+        return result.fetchone()["data"]
+
+
+def get_highlevel_models(id):
+    # Get highlevel model data for a specified id.
+    try:
```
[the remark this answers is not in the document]
aidanlw17: Added the docstring.

db/similarity.py
```diff
@@ -0,0 +1,173 @@
...
+
+    with db.engine.connect() as connection:
+        metrics = similarity.utils.init_metrics()
+
+        # Collect and assign stats to metrics that require normalization
+        for metric in metrics:
+            db.similarity_stats.assign_stats(metric)
+
+        sim_count = count_similarity()
+        current_app.logger.info("Processed {} / {} ({:.3f}%)".format(sim_count,
```
aidanlw17: @alastair I looked for a solution to echo this information in the manage.py command rather than the db module but wasn't very successful. Since it all happens in one loop I'm not sure how we could send info back intermittently? Replaced with logger for now.

alastair:
The `eval` branch already contains all of this work, so I'm going to close this PR and continue any necessary work in #364 

```
alastair@apmbp:~/code/acousticbrainz-server$ git checkout eval
Already on 'eval'
Your branch is up to date with 'aidanlw17/eval'.
alastair@apmbp:~/code/acousticbrainz-server$ git merge aidanlw17/compute-metrics
Already up to date.
```

PR #355, opened 2019-06-25
AB-415: Create similarity API endpoints

aidanlw17:
# AB-415: Create similarity API endpoints

After building indices, we need to create endpoints that will allow users to interact with the similarity data that we have computed. In order to do this, we should start by creating the following endpoints:

### Get similar recordings

/api/v1/<uuid:mbid>/similarity

This endpoint should take required parameters that are the metric to be used, and the MBID in question, with an optional parameter for the offset. For now, it will also accept index parameters which are the number of trees in the index, and the distance type. These parameters will allow us to select specific index builds while we are evaluating the effectiveness of parameter combinations.

### Get bulk similar recordings

/api/v1/similarity

This endpoint takes the same index parameters metric, distance type, and number of trees as the single get similar recordings endpoint, however instead of taking mbid and offset parameters, it should accept a list of recordings similarly to other bulk endpoints.

### Get similarity between recordings

/api/v1/<uuid(strict=False):mbid>/<uuid(strict=False):mbid>/similarity
This will return the distance (vector) for similarity between two specified recordings.


aidanlw17:
Closed as the content of this PR is covered in #364.

PR #356, opened 2019-07-01
AB-413: Redirect to HTTPS when accessing the login page in production

paramsingh:
https://tickets.metabrainz.org/browse/AB-413

### DIFF
--- a/webserver/__init__.py
+++ b/webserver/__init__.py
@@ -11,7 +11,9 @@ import urlparse
 
 API_PREFIX = '/api/'
 PRODUCTION_BASE_URL = 'acousticbrainz.org'
-PRODUCTION_BASE_URL_HTTPS = 'https://acousticbrainz.org'
+BETA_BASE_URL = 'beta.acousticbrainz.org'
+PRODUCTION_URL_HTTPS = 'https://acousticbrainz.org'
+BETA_URL_HTTPS = 'https://beta.acousticbrainz.org'
 
 
 # Check to see if we're running under a docker deployment. If so, don't second guess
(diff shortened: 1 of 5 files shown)


webserver/__init__.py
```diff
@@ -131,6 +134,17 @@ def before_request_gdpr_check():
         elif current_user.is_authenticated and current_user.gdpr_agreed is None:
             return redirect(url_for('index.gdpr_notice', next=request.full_path))
 
+    @app.before_request
+    def prod_https_login_redirect():
+        """ Redirect to HTTPS on the production login page
+        """
+        split_url = urlparse.urlsplit(request.url)
```
alastair: just checking - what will this be when going through the MeB load balancers? I’ve not done it with flask before, but for example when I use loadbalancer https -> webserver http -> uwsgi -> django, the scheme is always http. @zas can you comment on what you think this might be for AB? Perhaps we can deploy a test app to check it.

How about doing this only when config.DEBUG is false?

webserver/__init__.py
```diff
@@ -131,6 +134,17 @@ def before_request_gdpr_check():
...
+        if split_url.scheme == 'http' \
+            and split_url.netloc == PRODUCTION_BASE_URL \
+            and split_url.path == url_for('login.index'):
+            redirect_url = urlparse.urljoin(PRODUCTION_BASE_URL_HTTPS, url_for('login.index', next=request.args.get('next')))
```
alastair: It seems a bit funny to do this before _every_ request, but only check if the url path is the login page. Why not do only when rendering the login button?

I was going to suggest that we could do this redirect on every url if it’s https, but that would cause problems with non-GET requests, so perhaps we should skip that.

webserver/__init__.py
zas: Internal network queries (between gateways & backend servers) are always unencrypted. Gateways are there to handle the SSL mess.
Else we need to manage certs to access to backends...

webserver/__init__.py
```diff
@@ -131,6 +136,20 @@ def before_request_gdpr_check():
         elif current_user.is_authenticated and current_user.gdpr_agreed is None:
             return redirect(url_for('index.gdpr_notice', next=request.full_path))
 
+    @app.before_request
+    def prod_https_login_redirect():
+        """ Redirect to HTTPS in production except for the API endpoints
+        """
+        split_url = urlparse.urlsplit(request.url)
+        if request.headers.get('X-Forwarded-Proto') == 'http' and not request.path.startswith(API_PREFIX):
```
alastair: I'd like to do this only in production, why can't we only do it if DEBUG is false?
This is going to cause problems if we do it on POST operations, this should only redirect on GET.
The important API endpoint that we don't want to redirect on is the lowlevel submission endpoint: https://github.com/metabrainz/acousticbrainz-server/blob/46f670d62546ab44aa9086c2724aba57d41e5387/webserver/views/api/legacy.py#L54 which is not under the API_PREFIX. There are other GET methods here that we shouldn't redirect on either.
If we only redirect when DEBUG is false, why can't you get the url directly from the request url instead of special-casing the main and beta domains?

paramsingh:
Ok, i have no idea what is happening here. I'll look at it tomorrow.

PR #359, opened 2019-07-09
AB-418: Format track lengths greater than an hour

alastair:
https://tickets.metabrainz.org/projects/AB/issues/AB-418

We were only formatting minutes and seconds of a length, not hours. Add them in

### DIFF
--- a/webserver/views/data.py
+++ b/webserver/views/data.py
@@ -142,6 +142,21 @@ def _get_youtube_query(metadata):
         )
 
 
+def _format_length(length_ms):
+    try:
+        length = float(length_ms) / 1000
+    except ValueError:
+        return "?:??"
+    mins, secs = divmod(length, 60)
+    hours, mins = divmod(mins, 60)
+    if hours >= 1:
+        return "%d:%02d:%02d" % (hours, mins, secs)
+    elif mins >= 1:
+        return "%d:%02d" % (mins, secs)
+    else:
+        return "00:%02d" % secs
+
+
 def _get_recording_info(mbid, metadata):
     info = {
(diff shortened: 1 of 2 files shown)


webserver/views/data.py
```diff
@@ -142,6 +142,21 @@ def _get_youtube_query(metadata):
...
+    except ValueError:
+        return "?:??"
+    mins, secs = divmod(length, 60)
+    hours, mins = divmod(mins, 60)
+    if hours >= 1:
+        return "%d:%02d:%02d" % (hours, mins, secs)
+    elif mins >= 1:
+        return "%d:%02d" % (mins, secs)
+    else:
+        return "00:%02d" % secs
```
alastair: I’m not sure if something less than a minute should be `00:13` or `0:13`. Musicbrainz seems to prefer the second, but I’ve gone for the first:
https://beta.musicbrainz.org/recording/489ef3ab-2fd1-4435-9e41-80553ff07ed1

alastair: @paramsingh feedback on this?

Freso: I feel for "0" it doesn’t matter too much if it’s "00" or "0", but it would feel weird to see e.g., "01:XX" or "02:XX". And at that point it may seem inconsistent to go "00" → "1" … "9" → "10". Maybe if there are hours as well, I might want "1:01:42". Maybe it’s that it’s weird to see zero‐padding on the highest value part? (Like, you wouldn’t normally write "00,05 g".)

alastair: This question is only for the case of less than a minute. For 1-9:59 we show just the 1 digit

PR #360, opened 2019-07-09
AB-419: Encode unicode text when exporting dataset csv

alastair:
https://tickets.metabrainz.org/projects/AB/issues/AB-419

The csv module in python 2 requires that all values to it are ascii.
Explicitly convert dataset description, class names, class descriptions before adding them to the csv.
Don't convert uuids because they're valid ascii

I tested csv import with unicode text in the csv file and it worked fine

### DIFF
--- a/webserver/views/datasets.py
+++ b/webserver/views/datasets.py
@@ -119,23 +119,28 @@ def _convert_dataset_to_csv_stringio(dataset):
 
     Returns:
         A rewound StringIO containing a CSV representation of the dataset"""
+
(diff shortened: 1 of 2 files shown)


alastair:
update: it seems that in some cases importing a csv causes errors: https://tickets.metabrainz.org/projects/AB/issues/AB-420

alastair:
@paramsingh should we merge this as-is and do another PR for import, or should we do it in this pr too?

PR #361, opened 2019-07-09
AB-417: Correctly delete datasets

alastair:
https://tickets.metabrainz.org/projects/AB/issues/AB-417

We had no tests for dataset delete, and so didn't know that the delete endpoint used the incorrect method to get the dataset before deleting it.

### DIFF
--- a/webserver/views/api/v1/datasets.py
+++ b/webserver/views/api/v1/datasets.py
@@ -87,9 +87,7 @@ def create_dataset():
 @auth_required
 def delete_dataset(dataset_id):
     """Delete a dataset."""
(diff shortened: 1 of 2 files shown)


webserver/views/api/v1/datasets.py
```diff
@@ -87,9 +87,7 @@ def create_dataset():
 @auth_required
 def delete_dataset(dataset_id):
     """Delete a dataset."""
-    ds = get_dataset(dataset_id)
-    if ds["author"] != current_user.id:
```
alastair: this check not needed because get_check_dataset does it

PR #362, opened 2019-07-11
Add Similarity Documentation

aidanlw17:
## Add Similarity Documentation

This PR is meant to introduce documentation about similarity in AB, specifically using Annoy. It outlines the use of metrics and indices for both developers and other users.


alastair:
I merged this into #364, we can continue any further changes there

PR #364, opened 2019-07-31
Baseline Similarity Evaluation with Annoy

aidanlw17:
# Baseline Similarity Evaluation with Annoy

This PR contains the full body of similarity work on AB using Annoy.
It adds an evaluation, /<uuid:mbid>/similar/<string:metric>, allowing
for users to find similar recordings to an (MBID, offset) pair in terms
of a specific metric. Users can provide ratings and suggestions about 
the recommendations.

The use of this evaluation, in combination with a variety of parameters
for building the Annoy index, will help us to improve and tune the
entire similarity engine.


aidanlw17:
@alastair This is no longer a WIP, it is ready for review. And it has been rebased with the work done on #353 as well. :)

similarity/script.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import print_function
```
alastair: to be removed when no longer needed

webserver/static/scripts/similarity/eval-recording.js
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+
+function EvalRecording(props) {
```
alastair: see if you can write all your new react components as classes

webserver/__init__.py
```diff
@@ -155,21 +155,25 @@ def register_ui(app):
...
         from webserver.views.user import user_bp
         from webserver.views.datasets import datasets_bp
+        from webserver.views.similarity import similarity_bp
         app.register_blueprint(index_bp)
         app.register_blueprint(data_bp)
         app.register_blueprint(stats_bp)
         app.register_blueprint(login_bp, url_prefix='/login')
         app.register_blueprint(user_bp)
         app.register_blueprint(datasets_bp, url_prefix='/datasets')
+        app.register_blueprint(similarity_bp, url_prefix='/similarity')
```
alastair: we should be able to disable user-facing parts of this project (similarity pages, eval) even if it's in master until we're ready for other people to use it. Use a config item like `FEATURE_EVAL_LOCATION`

webserver/views/data.py
```diff
@@ -1,10 +1,17 @@
...
 from webserver.external import musicbrainz
 from werkzeug.exceptions import NotFound, BadRequest
 from six.moves.urllib.parse import quote_plus
+
 import db.data
 import db.exceptions
+import db.similarity
+import similarity.exceptions
+from similarity.index_model import AnnoyModel
+from webserver import forms
```
alastair: lots of imports with no other changes in the file... unused imports?

webserver/static/scripts/similarity/eval-recording.js
aidanlw17: I'll leave this one for now as other people might be looking over this evaluation frontend. But now, with the use of react hooks, the functional components are more in-style than ever!

similarity/index_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    file_path = os.path.join(os.getcwd(), 'annoy_indices')
+    name = '_'.join([metric, distance_type, str(n_trees)]) + '.ann'
+    full_path = os.path.join(file_path, name)
+    if os.path.exists(full_path):
+        os.remove(full_path)
+
+
+def get_similar_recordings(metric, mbid, offset, distance_type="angular", n_trees=10, n_neighbours=200):
+    # Initializes index with given params, and calls for similar recordings.
+    index = AnnoyModel(metric, n_trees=n_trees, distance_type=distance_type, load_existing=True)
```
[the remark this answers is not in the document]
aidanlw17: I'm not entirely sure what you mean by this usage. Do you mean something like:
```
model = new AnnoyModel(...)
model.load_model(metric, n_trees, distance_type)
```
Or do you mean we should add a static method that will load it and return an instance of AnnoyModel?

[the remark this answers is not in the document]
alastair: In this case I was meaning a static method

PR #366, opened 2019-09-26
Similarity Path Hack

aidanlw17:
# Similarity Path Hack
 
The goal of this PR is to create a path of similar recordings from one recording to another. First we should select two recordings, then query Annoy for similar recordings to one of them. We should select the similar recording that gives the smallest distance between itself and the recoding that was not queried for similarity. We should repeat this iteratively until we have defined a path of recordings.

We should also expose this utility with an API endpoint.


alastair:
2 missing things here - similarity module has no `__init__.py`, and numpy isn't in requirements

alastair:
@aidanlw17 I added a basic webservice endpoint that we can use to call the path generation from

alastair:
This was an experiment to use similarity information to generate recommended recordings to listen to. It has been mostly replaced by troi.

PR #369, opened 2020-04-21
AB-391: Use webpack to compile js and css assets

alastair:
JS and less->css compilation is working, including the generation of a manifest file

todo:
 - [ ] ensure that production build correctly strips unneeded code and minifies
 - [x] Add watch mode for dev mode
 - [x] See if babel caching is needed
 - [x] Ensure that production mode adds hashes to filenames
 - [x] In local dev mode, don't use use the manifest file (this will mean that we no longer need our custom run command to watch the manifest file)
- [ ] revise [webpack production guidelines](https://webpack.js.org/guides/production/), including [mini-css-extract-plugin minification](https://webpack.js.org/plugins/mini-css-extract-plugin/#minimizing-for-production)


webpack.config.js
```diff
@@ -0,0 +1,60 @@
...
+                        // Fixes issue where we @import url(https://googlefont)
+                        options: {
+                            paths: [path.resolve(__dirname, 'node_modules')],
+                        }
+                    }
+                ]
+            }
+        ],
+    },
+    plugins: [new ManifestPlugin(), new MiniCssExtractPlugin()]
```
alastair, line 57: It would be great to run ManifestPlugin only in production. What's the recommended way to do this? Set an option? (I can't see [any specific option that might help us](https://github.com/danethurber/webpack-manifest-plugin#api) here), or is it better to just not load the plugin if we're in development mode?

webpack.config.js
```diff
@@ -0,0 +1,60 @@
+const path = require('path');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const ManifestPlugin = require('webpack-manifest-plugin');
+
+const production = process.env.NODE_ENV === 'production';
```
alastair, line 5: this differs from listenbrainz - https://github.com/metabrainz/listenbrainz-server/blob/3850306efc3674ba1601429862280049102bedae/webpack.config.js#L6
which takes `env` as an argument to module.exports. How do these methods differ?

package.json
```diff
@@ -6,31 +6,39 @@
...
+    "build:prod": "NODE_ENV=production webpack -p"
+  },
   "dependencies": {
+    "@babel/preset-react": "^7.0.0",
     "babel-core": "6.26.0",
+    "babel-loader": "^7.1.5",
     "babel-plugin-transform-react-constant-elements": "6.23.0",
     "babel-plugin-transform-react-inline-elements": "6.22.0",
     "babel-plugin-transform-react-jsx": "6.24.1",
     "babel-preset-es2015": "6.24.1",
```
MonkeyDo, line 20: This will be obsolete with Babel 7, replaced by @babel/preset-env

webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+        stats: ['./scripts/stats.js'],
+        main: ['./styles/main.less']
+    },
+    output: {
+        chunkFilename: production ? '[name].[chunkhash].js' : '[name].js',
+        filename: production ? '[name].[chunkhash].js' : '[name].js',
+        path: path.resolve(__dirname, 'webserver', 'static', 'build'),
+        publicPath: '/static/build/'
+    },
+    mode: production ? 'production' : 'development',
```
MonkeyDo, line 25: I *think* this is redundant, since calling webpack with the `-p` flag sets mode to production automatically, and the default is development.

webpack.config.js
MonkeyDo: As we discussed on IRC, this could be solved by exporting a function rather than an object, and using env arguments passed to the cli command: https://webpack.js.org/guides/environment-variables/#root

package.json
```diff
@@ -6,31 +6,39 @@
...
+    "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "6.24.1",
```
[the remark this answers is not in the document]
alastair: I've upgraded to babel7, pending some additional testing before I push it.

webpack.config.js
alastair: I'll update it to reflect the LB setup for some consistency, and this will make it easier to do this, thanks

PR #370, opened 2020-04-22
AB-392: Upgrade react

alastair:
To be merged after #369

This change upgrades from react 0.14 -> 15.0 -> 15.5 -> 16

Two major changes that were made are switching from React.createClass to ES2015 classes, and swapping React.PropTypes to the PropTypes package

TODO:
- [x] Change use of `createReactClass` in dataset/editor.js `RecordingAddForm` to use a class. It seems like the automatic upgrader skipped this one because of a specific structure of the class
- [x] New warning that appears when editing a dataset and going to a class that doesn't have a description: ```Warning: `value` prop on `textarea` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components.```. The cause is that the dataset json from the API has `null` for this value when it's not set. It should either be `''` from the API, or the frontend should change `null`->`''`


alastair:
Warning is covered by https://tickets.metabrainz.org/browse/AB-436. It doesn't cause any major issues using the editor, so will skip it for now.

README.md
```diff
@@ -88,7 +88,7 @@ To build stylesheets and javascript bundles, run webpack:
 You will need to rebuild static files after you modify JavaScript or CSS. If you want to rebuild
 these source files as you change them then you can run webpack in watch mode:
 
-    ./develop.sh npm run build:dev -- --watch
+    ./develop.sh npm run build:watch
```
paramsingh, line 91: We have automatic watching and building of javascript assets in LB btw, if you want to go that way, it's definitely very convenient.

webserver/static/scripts/datasets/class-viewer.js
```diff
@@ -1,3 +1,4 @@
+var PropTypes = require('prop-types');
```
paramsingh, line 1: we've started using typescript in LB to solve the typing problem, fyi. Maybe there's an incentive to stay consistent here.

paramsingh: Not right now, obviously, but maybe as a end goal. Maybe new react components could be in typescript.

alastair: yeah, some more cleanup of these scripts and potentially moving to ts is an option here. For now I'm just getting it working with new react.

package.json
```diff
@@ -8,36 +8,36 @@
...
-    "babel-core": "^6.26.3",
-    "babel-loader": "^7.1.5",
-    "babel-plugin-transform-react-constant-elements": "6.23.0",
-    "babel-plugin-transform-react-inline-elements": "6.22.0",
-    "babel-plugin-transform-react-jsx": "6.24.1",
-    "babel-preset-es2015": "6.24.1",
-    "babel-preset-react": "^6.24.1",
-    "babel-preset-stage-2": "6.24.1",
     "bootstrap": "3.4.1",
+    "create-react-class": "^15.6.3",
```
alastair, line 16: this is duplicated in devDependencies, and is now no longer needed anyway because we removed all instances of createClass

README.md
alastair: we talked about this in IRC. In the future if we need constant asset compilation we'll set up a container to build assets like in LB

PR #372, opened 2020-04-27
AB-437: Make files in /code owned by the acousticbrainz user

alastair:
This was a little bit more involved than I had hoped, because npm installs things in the working directory... Because of this I had to make the `acousticbrainz` user earlier, and then `COPY --chown` everywhere, and use `USER` to switch between users. Now, everything in /code is owned by acousticbrainz.

Some upsides: now, everywhere that you run this, it's running as a non-privileged user!
Some downsides
 * now you can't run `pip install` in a shell if you temporarily want to install something while developing. Should we revert back to running as root for this?
 * Running in dev on a linux host, we've still got the issue where commands are being run as uid 901, and therefore files which are created by it can't be read/deleted by the user on their host without `sudo chown` magic

@paramsingh any feedback on what uid we should be running as in dev?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM metabrainz/python:2.7 AS acousticbrainz-base
 
-ARG deploy_env
-
 # Dockerize
 ENV DOCKERIZE_VERSION v0.6.1
 RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
@@ -41,7 +39,6 @@ RUN wget -q -O - https://deb.nodesource.com/setup_12.x | bash - && apt-get updat
     && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /code
-RUN mkdir /code/hl_extractor
 RUN mkdir /data
 WORKDIR /code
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -75,36 +72,39 @@ RUN mkdir /tmp/models \
...
-COPY requirements.txt /code/requirements.txt
+RUN mkdir /code/docs/ && chown acousticbrainz:acousticbrainz /code/docs/
+COPY --chown=acousticbrainz:acousticbrainz docs/requirements.txt /code/docs/requirements.txt
+COPY --chown=acousticbrainz:acousticbrainz requirements.txt /code/requirements.txt
 RUN pip install --no-cache-dir -r requirements.txt
 
-COPY package.json /code
+COPY --chown=acousticbrainz:acousticbrainz package.json /code
+
+USER acousticbrainz
```
alastair, line 88: Thinking out loud about this - if we don't want to create the user in the -base target, we could skip the installation of npm packages here. Instead, only install in -prod. During dev we get users to run npm install again anyway to create node_modules in their acousticbrainz-server which is bind-mounted into the container.
This allows us to keep running as root in development (and fixes the concern in the PR body).

Still a pending question about when we should chown /code (and requirements.txt) in the -prod build

docker/push.sh
```diff
@@ -19,9 +19,7 @@ TAG=${2:-beta}
 
 echo "Building AcousticBrainz web image with env $ENV tag $TAG..."
 docker build -t metabrainz/acousticbrainz:$TAG \
-        --target acousticbrainz-prod \
-        --build-arg GIT_COMMIT_SHA=$(git rev-parse HEAD) \
-        --build-arg deploy_env=$ENV .
```
alastair, line 24: now that we have no env-specific startup scripts, we don't need this. Not sure if we still need the git data - it would be useful in sentry, but we don't install it

alastair:
This change causes jenkins to fail too because it tries to write results to `/`. this should be changed

Dockerfile
paramsingh: I would prefer not to add an extra step to the build instructions. In ListenBrainz, we use a different container to build npm assets automatically into the correct place, which is ideally what I'd want for AB too.

docker/push.sh
paramsingh: We should definitely try to install it!

alastair:
we also decided that we should add the test script from jenkins into the repo, so that jenkins isn't the only location that it's stored.

docker/push.sh
alastair: https://tickets.metabrainz.org/projects/BU/issues/BU-31

alastair:
Tests fixed. In the end I decided to leave dev/test running as root. The jenkins script was already in the repo here, so I just updated the project configuration to run the script from the repo instead of having it copied into the execute shell configuration

PR #373, opened 2020-04-28
AB-436: Fix issues. datasets' class descriptions accept empty strings

tzamalisp:
Based on the [AB-436](https://tickets.metabrainz.org/browse/AB-436?jql=project%20%3D%20AB%20AND%20status%20%3D%20Open%20AND%20labels%20%3D%20good-first-bug) jira, the following issues are now fixed:

- When uploading a dataset via a CSV file, if the class name description is empty during the submission, the corresponding value that is stored into the DB is not anymore a null value but an empty string `''`.

- The `defaultdict(lambda: {"description": None, "recordings": []})` had to change to `defaultdict(lambda: {"description": '', "recordings": []})` class dictionary declaration inside the _"webserver/views/datasets.py", line 338, `_parse_dataset_csv()`_ function, in order to change the `key/value` class description of the class dictionary declaration.

- The same submission logic is fixed and followed for both the CSV importer and the dataset editor. This logic has been fixed too in case of updating the dataset.

- Additionally, inside the _"db/dataset.py"_ the `create_from_dict()` and `update()` functions are changed in the lines where a for loop was searching inside the class dictionary for the description key, and if it was missing the class description was set to `None` (i.e. `null`). Now, these values are set to `''`.

- The dataset_class table schema is also changed so that the "description" column can't anymore accept null values. The relevant constraint is now updated inside the _"admin/sql/create_tables.sql"_ file for that table. Finally, a new file _"admin/updates/20200428-alter-dataset_class-schema-description-not-null.sql"_ is created that automatically updates the existing stored null values with the _''_ and alters the dataset_class table schema to the new constraint.

- In _"admin/sql/create_tables.sql"_, a constraint `NOT NULL` is set on the table `dataset` for the `description` column.

-  In _"admin/updates/20200428-alter-dataset_class-schema-description-not-null.sql"_, two extra queries have been created that set the null values to an empty string of the `dataset`'s table `description` column and the `dataset` table schema is altered to not accept NULL values. 

- In _"db/test/test_dataset.py"_, in `_create_dataset()`, the  argument declaration is set from `desc=None` to `desc=""`.

Someone can test the CSV importer or the dataset editor if it works successfully by uploading or updating a dataset and submitting an empty description in the class description field (CSV or editor) or uploading a CSV dataset with only the _"<MBID>,<class>"_ fields in each row (i.e. without a _"description:<class>,<description text>"_ row inside the CSV file). Afterward, the submission of null values can be checked by opening a psql prompt and running, for example, the following query:

`SELECT name, description FROM dataset_class WHERE description IS NULL;`

0 rows should be returned back after running the query which means that no null values are now stored into the database in case of an empty class description.

p.s. As I 'm new to the group, I may not follow the PR guidelines explicitly but please, do not hesitate to ask me for further information if you have any questions.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -89,7 +89,7 @@ CREATE TABLE dataset (
 CREATE TABLE dataset_class (
   id          SERIAL,
   name        VARCHAR NOT NULL,
-  description TEXT,
+  description TEXT    NOT NULL, -- NULL values are not accepted
   dataset     UUID    NOT NULL -- FK to dataset
 );
 
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -79,7 +79,7 @@ ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id
 CREATE TABLE dataset (
   id          UUID,
   name        VARCHAR NOT NULL,
-  description TEXT,
+  description TEXT NOT NULL,
   author      INT NOT NULL, -- FK to user
   public      BOOLEAN NOT NULL,
   created     TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -89,7 +89,7 @@ CREATE TABLE dataset (
 CREATE TABLE dataset_class (
   id          SERIAL,
   name        VARCHAR NOT NULL,
-  description TEXT    NOT NULL, -- NULL values are not accepted
+  description TEXT    NOT NULL,
   dataset     UUID    NOT NULL -- FK to dataset
 );
 
--- a/db/dataset.py
+++ b/db/dataset.py
@@ -44,7 +44,9 @@ def create_from_dict(dictionary, author_id):
 
         for cls in dictionary["classes"]:
             if "description" not in cls:
-                cls["description"] = None
+                # cls["description"] = None
+                cls["description"] = ''
+                
             result = connection.execute("""INSERT INTO dataset_class (name, description, dataset)
                               VALUES (%s, %s, %s) RETURNING id""",
                                         (cls["name"], cls["description"], dataset_id))
@@ -124,7 +126,9 @@ def update(dataset_id, dictionary, author_id):
 
         for cls in dictionary["classes"]:
             if "description" not in cls:
-                cls["description"] = None
+                # cls["description"] = None
+                cls["description"] = ''
+
             result = connection.execute("""INSERT INTO dataset_class (name, description, dataset)
                               VALUES (%s, %s, %s) RETURNING id""",
                                         (cls["name"], cls["description"], dataset_id))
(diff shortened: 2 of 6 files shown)


alastair:
@brainzbot add to whitelist

admin/sql/create_tables.sql
```diff
@@ -89,7 +89,7 @@ CREATE TABLE dataset (
 CREATE TABLE dataset_class (
   id          SERIAL,
   name        VARCHAR NOT NULL,
-  description TEXT,
+  description TEXT    NOT NULL, -- NULL values are not accepted
```
alastair, line 92: we should do the same with the `dataset` table above too. The comment here isn't necessary, we can already see from the NOT NULL keyword that nulls aren't accepted.
The update script should also update the dataset table

db/dataset.py
```diff
@@ -44,7 +44,9 @@ def create_from_dict(dictionary, author_id):
 
         for cls in dictionary["classes"]:
             if "description" not in cls:
-                cls["description"] = None
+                # cls["description"] = None
```
alastair, line 47: We can just delete this, we don't need to keep it. I would also update the check to ensure that someone doesn't intentionally pass in a null value: `if "description" not in cls or cls["description"] is None:`
We should have a test in `db/test/test_dataset` to ensure that, if we pass in a class with no description or a value of None for a description, the dataset is still saved. a variation of `test_create_from_dict` would be good

alastair:
Now that jenkins has run the tests, I see that there is already a test failure: https://ci.metabrainz.org/job/acousticbrainz-server/676/console
See the section at the bottom of README about how to run tests locally.

admin/sql/create_tables.sql
tzamalisp: The changes are done. The update script is also updated for making the similar changes to the dataset table

db/dataset.py
tzamalisp: The commented lines have now been deleted. I checked also many parts of various scripts that insert `None` value in the description, and everything has been updated. I hope I don't miss anything. As you guided me, I ran also the tests locally as it is proposed in the README and everything is fine now. 

Yes, a variation of the  `test_create_from_dict` is a good idea. However, you are right about the changes that have to be done in `db/test/test_dataset.py`. When I ran the tests with the updated table schema of the dataset table, I saw a conflict in `_create_dataset()` inside the test_dataset where the argument `desc` was set to `None`. I've changed the argument value to `""` and it worked.

alastair: right, I think there are 2 tests that we need here:
* If you upload a csv file with no description, the description value should be saved to the database as `""`, this is the change that you fixed
* If someone uses the [dataset create api](https://acousticbrainz.readthedocs.io/api.html#post--api-v1-datasets-) to submit a new dataset and provides no description _or a value of `None` for the description_ then the description should be created with the description changed to `""`. This is where we need a new test, `test_create_from_dict_none_description`, or similar.

tzamalisp: ok! I'll start checking it later today. Thanks

tzamalisp: the test case for the above conditions has been created as well as a minor fix in the _db/dataset.py_ that sets the description to `""` when it is inserted as `None` from the _create dataset API_.

db/test/test_dataset.py
```diff
@@ -50,6 +50,24 @@ def test_create_from_dict(self):
...
         self.assertEqual(len(ds["classes"][1]["recordings"]), 3)
 
+    def test_create_from_dict_none_description(self):
+        bad_dict = copy.deepcopy(self.test_data)
+        bad_dict["description"] = None
+        bad_dict["classes"][1]["description"] = None
+        id = dataset.create_from_dict(bad_dict, author_id=self.test_user_id)
+        ds = dataset.get(id)
+        self.assertIsNotNone(ds["classes"][1]["description"])
+        self.assertIsNotNone(ds["description"])
```
alastair, line 60: this test is perfect, thanks 👍

PR #374, opened 2020-05-01
AB-432: Allow badly-formatted MBIDs, but normalise them before use

alastair:
Postgres requires that uuids are all lower-case, and formatted with hyphens in the correct locations, but uuid.UUID accepts values missing hyphens or with them in non-standard locations. Use str(uuid.UUID(val)) to always normalise these values before passing to postgres.

### DIFF
--- a/webserver/views/api/v1/core.py
+++ b/webserver/views/api/v1/core.py
@@ -213,7 +213,7 @@ def _parse_bulk_params(params):
         elif len(parts) == 2:
             offset = parts[1]
         else:
-            raise webserver.views.api.exceptions.APIBadRequest("More than 1 : in '%s'" % recording)
+            raise webserver.views.api.exceptions.APIBadRequest("More than 1 colon (:) in '%s'" % recording)
 
         args = _validate_arguments(mbid, offset)
         ret.append(args)
@@ -262,7 +262,7 @@ def get_many_lowlevel():
 
        {"mbid1": {"offset1": {document},
                   "offset2": {document}},
-        "mbid2": {"offset1": {document}}
+        "mbid2": {"offset1": {document}},
         "mbid_mapping": {"MBID1": "mbid1"}
        }
 
@@ -271,9 +271,9 @@ def get_many_lowlevel():
     the offset will be 0.
 
     MBID keys are always returned in a normalised form (all lower-case, separated in groups of 8-4-4-4-12 characters),
-    even if the provided recording MBIDs are not given in this form. In the case that a MBID is not given in this
-    normalised form, the value `mbid_mapping` in the response will be a dictionary mapping user-provided MBIDs to
-    this form.
+    even if the provided recording MBIDs are not given in this form. In the case that a requested MBID is not given
+    in this normalised form, the value `mbid_mapping` in the response will be a dictionary mapping user-provided MBIDs
+    to this form.
 
     If the list of MBIDs in the query string has a recording which is not
     present in the database, then it is silently ignored and will not appear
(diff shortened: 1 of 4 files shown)


alastair:
Got to decide on line limits with pep8... Will open a ticket for it

webserver/views/api/v1/core.py
```diff
@@ -274,13 +312,18 @@ def get_many_highlevel():
 
        {"mbid1": {"offset1": {document},
                   "offset2": {document}},
-        "mbid2": {"offset1": {document}}
+        "mbid2": {"offset1": {document}},
+        "mbid_mapping": {"MBID1": "mbid1"}
```
alastair, line 316: This is the core change, now we'll return this `"mbid_mapping"` field with bulk queries. I decided to _always_ include the key, but only include items here which weren't in the normalised format. So if a valid, lower-case mbid is passed in, nothing will show here. Also, if the mbid doesn't exist in the database, it'll appear here, but won't be in the rest of the dictionary. I think this is a good trade-off, but could be convinced to always put all parameters in this mapping if we think it's needed.

webserver/views/api/v1/core.py
```diff
@@ -236,12 +263,17 @@ def get_many_lowlevel():
...
        }
 
     MBIDs and offset keys are returned as strings (as per JSON encoding rules).
     If an offset is not specified in the request for an MBID or is not a valid integer >=0,
     the offset will be 0.
-    MBID keys are always lower-case, even if the provided recording MBIDs are upper-case or mixed case.
+
+    MBID keys are always returned in a normalised form (all lower-case, separated in groups of 8-4-4-4-12 characters),
+    even if the provided recording MBIDs are not given in this form. In the case that a MBID is not given in this
```
paramsingh, line 274: Could clarify that we're talking about MBID given to us in the request (and not the response) in the second sentence.
```suggestion
    even if the provided recording MBIDs are not given in this form. In the case that a MBID in the request is not in this
```

webserver/views/api/v1/core.py
paramsingh: I'm happy to keep it as it is now.

webserver/views/api/v1/core.py
```diff
@@ -274,13 +312,18 @@ def get_many_highlevel():
...
+        "mbid_mapping": {"MBID1": "mbid1"}
        }
 
     MBIDs and offset keys are returned as strings (as per JSON encoding rules).
     If an offset is not specified in the request for an mbid or is not a valid integer >=0,
     the offset will be 0.
-    MBID keys are always lower-case, even if the provided recording MBIDs are upper-case or mixed case.
+
+    MBID keys are always returned in a normalised form (all lower-case, separated in groups of 8-4-4-4-12 characters),
+    even if the provided recording MBIDs are not given in this form. In the case that a MBID is not given in this
```
paramsingh, line 324: Could clarify that we're talking about MBID given to us in the request (and not the response) in the second sentence.
```suggestion
    even if the provided recording MBIDs are not given in this form. In the case that a MBID in the request is not in this
```

webserver/views/api/v1/core.py
```diff
@@ -91,13 +92,15 @@ def get_high_level(mbid):
 
     :query n: *Optional.* Integer specifying an offset for a document.
     :query map_classes: *Optional.* If set to 'true', map class names to human-readable values
+    TODO: provide a link to what these mappings are
```
paramsingh, line 95: Should fix this TODO, or is the plan to add a link to readthedocs when you deploy the docs changes?

webserver/views/api/v1/core.py
```diff
@@ -160,46 +192,41 @@ def _parse_bulk_params(params):
...
-            try:
-                offset = int(parts[1])
-                # Don't allow negative offsets
-                offset = max(offset, 0)
-            except ValueError:
-                offset = 0
+            offset = parts[1]
         else:
-            offset = 0
+            raise webserver.views.api.exceptions.APIBadRequest("More than 1 : in '%s'" % recording)
```
paramsingh, line 216:
```suggestion
            raise webserver.views.api.exceptions.APIBadRequest("More than 1 colon(:) in '%s'" % recording)
```

webserver/views/api/v1/core.py
alastair: good idea, thanks

webserver/views/api/v1/core.py
alastair: I've opened a ticket at https://tickets.metabrainz.org/browse/AB-444 for this

PR #376, opened 2020-05-11
AB-67, AB-366: Use the flask logger instead of print for informational messages

alastair:
https://tickets.metabrainz.org/browse/AB-67
https://tickets.metabrainz.org/browse/AB-366

Use `flask.current_app.logger` instead of print

### DIFF
--- a/hl_extractor/hl_calc.py
+++ b/hl_extractor/hl_calc.py
@@ -11,6 +11,7 @@ from threading import Thread
 from time import sleep
 
 import yaml
+from flask import current_app
 
 import db
 import db.data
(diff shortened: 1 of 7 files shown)


hl_extractor/hl_calc.py
```diff
@@ -47,7 +48,7 @@ def _calculate(self):
             f.write(self.ll_data.encode("utf-8"))
             f.close()
         except IOError:
-            print("IO Error while writing temp file")
+            current_app.logger.error("IO Error while writing temp file")
```
alastair, line 51: In further testing, I found that this is a problem. access to `current_app` isn't thread-safe, even though the underlying logging object is. I'll revert the changes in this file, pending some changes that I'm making in #378

PR #378, opened 2020-05-11
AB-439: Use concurrent.futures to manage threading in hl extractor

alastair:
https://tickets.metabrainz.org/projects/AB/issues/AB-439
https://tickets.metabrainz.org/projects/AB/issues/AB-440

This is the first part of AB-439 and AB-440. I'm going to split it up into a few different PRs to make it easier to follow the evolution of the changes.

This specific change moves from manual thread pool management to using [concurrent.futures](https://docs.python.org/3/library/concurrent.futures.html). For now we use a python2 backport. This won't be needed in python 3. This makes the queue submission code much simpler and easier to extend.

There's one difference to the way we load the thread pool in this new version. We load all document to the pool, but when it starts to run out we don't fill it up again. e.g. if we have 10 threads, and we've processed 90 and are on the last 10 documents, and some threads finish processing. In this case, we'll have some threads lying around. I didn't improve this case because it will add additional complexity to the code, and we know that each thread takes about the same time to run (and only a few seconds at that), so we assume that in most cases the threads will all finish at about the same time.

This also addresses the second point listed at AB-440. If we ask for 100 unprocessed items but only get 15, it means that there aren't many pending items. We should sleep for 30s after this to not continually query the database for items to process.

I've tested this myself, and the results seem to be correct (the hl data for the associated ll data is correctly written to the db), but @paramsingh I'd appreciate you looking over the changes and seeing if you can find anything that looks wrong.

### DIFF
--- a/hl_extractor/hl_calc.py
+++ b/hl_extractor/hl_calc.py
@@ -1,21 +1,24 @@
 #!/usr/bin/env python
-from __future__ import print_function
-
 import hashlib
 import json
+import logging
 import os
+import shutil
 import subprocess
 import sys
 import tempfile
-from time import sleep
+import time
+import traceback
 
 import concurrent.futures
+import six
 import yaml
+from flask import current_app
 
 import db
 import db.data
 
-DEFAULT_NUM_THREADS = 1
+DEFAULT_NUM_THREADS = 2
 
 SLEEP_DURATION = 30  # number of seconds to wait between runs
 BASE_DIR = os.path.dirname(__file__)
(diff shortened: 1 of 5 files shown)


hl_extractor/hl_calc.py
```diff
@@ -131,85 +111,56 @@ def create_profile(in_file, out_file, sha1):
...
+    except ValueError:
+        print("error %s: Cannot parse result document" % mbid)
+        print(hl_data)
+        jdata = {}
 
+    db.data.write_high_level(mbid, rowid, jdata, build_sha1)
 
-def main(num_threads):
+
+def cf_main(num_threads):
```
paramsingh, line 134: We should need to change the name where this gets invoked right: https://github.com/metabrainz/acousticbrainz-server/blob/0cfc2a72/worker_manage.py#L17?

paramsingh, line 134: We should need to change the name where this gets invoked right: https://github.com/metabrainz/acousticbrainz-server/blob/0cfc2a72/worker_manage.py#L17?

hl_extractor/hl_calc.py
```diff
@@ -131,85 +111,56 @@ def create_profile(in_file, out_file, sha1):
...
+
+            for future in concurrent.futures.as_completed(future_to_llid):
+                rowid, mbid = future_to_llid[future]
+                try:
+                    hl_data = future.result()
+                except Exception as exc:
+                    print('%r generated an exception: %s' % (mbid, exc))
+                else:
+                    num_processed += 1
+                    save_hl(mbid, rowid, hl_data, build_sha1)
```
paramsingh, line 158: Should we do some exception handling here too?  In case the db times out or something?

hl_extractor/hl_calc.py
alastair: yes, this was carryover from some of my testing, I have some improvements + tests to add to this, and have fixed it.

hl_extractor/hl_calc.py
alastair: yeah, we could move it into the try block and then have a specific except block to catch these errors. Generally, we haven't been catching db errors, but it's probably a good idea

hl_extractor/hl_calc.py
alastair: I fixed this when merging #381

PR #381, opened 2020-06-25
AB-349 (II): Process multiple documents with the highlevel extractor at once

alastair:
This is the second part of AB-439. Once approved, we can merge it into #378 and then merge that.

The slowst part of the highlevel extractor is loading the model files. However, the the extractor can take multiple lowlevel files as input and will output multiple highlevel files. This means that it only loads the models once, and the runtime for 20 lowlevel files will be almost 20x faster than running the binary 20 times in a row. 

Add some tests for the main part of the hl extractor

Use the logger instead of print statements. Manually get the logger in the main method that is run in a thread, to prevent flask threading issues.

One final thing I want to do before merging these 2 items into master is to set up a test server with a handful of documents and run both HL extractors, ensuring that the resulting hl files are the same.

### DIFF
--- a/hl_extractor/hl_calc.py
+++ b/hl_extractor/hl_calc.py
@@ -1,21 +1,24 @@
 #!/usr/bin/env python
-from __future__ import print_function
-
 import hashlib
 import json
+import logging
 import os
+import shutil
 import subprocess
 import sys
 import tempfile
-from time import sleep
+import time
+import traceback
 
 import concurrent.futures
+import six
 import yaml
+from flask import current_app
 
 import db
 import db.data
 
-DEFAULT_NUM_THREADS = 1
+DEFAULT_NUM_THREADS = 2
 
 SLEEP_DURATION = 30  # number of seconds to wait between runs
 BASE_DIR = os.path.dirname(__file__)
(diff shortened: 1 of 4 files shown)


hl_extractor/hl_calc.py
```diff
@@ -114,53 +165,65 @@ def get_build_sha1(binary):
...
-def cf_main(num_threads):
-    print("High-level extractor daemon starting with %d threads" % num_threads)
-    build_sha1 = get_build_sha1(HIGH_LEVEL_EXTRACTOR_BINARY)
-    create_profile(PROFILE_CONF_TEMPLATE, PROFILE_CONF, build_sha1)
+def main(num_threads=DEFAULT_NUM_THREADS):
+    current_app.logger.info("High-level extractor daemon starting with {} threads".format(num_threads))
+    if num_threads * MAX_ITEMS_PER_PROCESS > DOCUMENTS_PER_QUERY:
+        current_app.logger.warn("Number of threads ({}) * items per thread ({}) = {} is greater than number of "
+                                "documents selected from database ({}), this means that available resources "
+                                "will not be fully utillised".format(
```
alastair, line 189: spelling

alastair:
I tested this by loading 10000 random documents into an AB database, generating hl documents with the old extractor code, and then with this code, dumping the results to json, and comparing the files. Both files were exactly the same.

PR #383, opened 2020-08-03
add sklearn ml tool

tzamalisp:
Add the new Machine Learning infrastructure to start the integration with the AB


models/sklearn/classification/classifierGRID.py
```diff
@@ -0,0 +1,153 @@
...
+                                               os.path.join(exports_dir, "results")).inspect_directory()
+            results_best_dict_name = "result_{}_{}_best_{}.json"\
+                .format(self.class_name, tr_process["preprocess"], gsvc.best_score_)
+
+            results_dict = dict()
+            results_dict["score"] = gsvc.best_score_
+            results_dict["params"] = gsvc.best_params_
+            results_dict["n_fold"] = tr_process['n_fold']
+            results_dict["preprocessing"] = tr_process["preprocess"]
+            with open(os.path.join(results_path, results_best_dict_name), 'w') as grid_best_json:
```
alastair, line 113: I'll have to look at this workflow in a bit more detail - it would be better to not have this method do too many things - e.g. it seems that it performs the training process and also saves this file. I would prefer that we split it up into more methods, for example one to do a specific training process, and then another one that loops through the settings, calls the training process, and then saves the results

models/sklearn/classification/classifierGRID.py
```diff
@@ -0,0 +1,153 @@
...
+                json.dump(results_dict, grid_best_json, indent=4)
+
+            # export parameters that the
```
alastair, line 116: unfinished comment

models/sklearn/classification/classifierGRID.py
```diff
@@ -0,0 +1,153 @@
...
+
+            print(colored("Next train process..", "yellow"))
+            process_counter += 1
+            print()
+            print()
+        print(colored("Finishing training processes..", "blue"))
+        print()
+
+    def export_best_classifier(self):
+        # gather best scores from the exported grid clf models
```
alastair, line 139: in gaia we output a results file that gives the accuracy and parameters of each item in the grid search. Do we do something similar here?

models/sklearn/classification/evaluation.py
```diff
@@ -0,0 +1,279 @@
...
+import requests
+from utils import load_yaml, FindCreateDirectory, TrainingProcesses
+from transformation.transform import Transform
+from transformation.utils_preprocessing import flatten_dict_full
+from classification.report_files_export import export_report
+from logging_tool import LoggerSetup
+
+
+def fold_evaluation(config, n_fold, X, y, class_name, tracks, process, exports_path, log_level):
+    # logger setup
```
alastair, line 23: this kind of comment isn't necessary in this code. We know that it's logger setup because that's the name of the class that you're calling!

models/sklearn/classification/evaluation.py
```diff
@@ -0,0 +1,279 @@
...
+                                  process=process,
+                                  train_class=class_name,
+                                  exports_path=exports_path,
+                                  log_level=log_level).post_processing()
+    logger.debug("features prepared shape: {}".format(features_prepared.shape))
+
+    accuracy_model = []
+    predictions_df_list = []
+    fold_number = 0
+    for train_index, test_index in inner_cv.split(features_prepared):
```
alastair, line 84: this is a very long method. It would be good to split it into separate methods for loading data, processing a single split, generating images, saving files, etc

models/sklearn/classification/report_files_export.py
```diff
@@ -0,0 +1,26 @@
...
+from termcolor import colored
+from utils import load_yaml, FindCreateDirectory, TrainingProcesses
+
+
+def export_report(config, name, report, filename, train_class, exports_path):
+    exports_dir = "{}_{}".format(config.get("exports_directory"), train_class)
+    reports_path = FindCreateDirectory(exports_path, os.path.join(exports_dir, "reports")).inspect_directory()
+    # take current date and convert to string
+    now = datetime.now()
+    datetime_str = now.strftime("%Y-%m-%d")
```
alastair, line 12: looks like this one isn't used?

models/sklearn/classification/report_files_export.py
```diff
@@ -0,0 +1,26 @@
...
+    datetime_str_verbose = now.strftime("%Y-%m-%d, %H:%M:%S")
```
alastair, line 13: you can use now.isoformat() here

models/sklearn/classification/report_files_export.py
```diff
@@ -0,0 +1,26 @@
...
+    print("Creating report file..")
+    with open(os.path.join(reports_path, "{}.txt".format(filename)), 'w+') as file:
```
alastair, line 15: `file` is a python reserved keyword, I typically use `fp` here

models/sklearn/classification/report_files_export.py
```diff
@@ -0,0 +1,26 @@
...
+        file.write("{}".format(name))
+        file.write('\n')
```
alastair, line 17: you can put multiple `\n`s in a single write command if you want. Don't mix " and '
In the line above, if you just have name, you can probably just use fp.write(name) unless there's a specific reason for doing the format

models/sklearn/classification/report_files_export.py
```diff
@@ -0,0 +1,26 @@
...
+        file.write('\n')
+        file.write(str(report))
+        file.write('\n')
+        file.write('\n')
+        file.write('\n')
+        file.write("Date of execution: {}".format(datetime_str_verbose))
```
alastair, line 23: put the date in the first line

models/sklearn/create_classification_project.py
```diff
@@ -0,0 +1,125 @@
...
+    parser.add_argument('-v', '--verbose',
+                        default=1,
+                        help="Controls the verbosity: the higher, the more messages.",
+                        type=int)
+    parser.add_argument('-p', '--path',
+                        dest='exports_path',
+                        help='Path where the project results will be stored. If empty, the results will be saved in '
+                             'app directory')
+
+    # parser.add_argument('-t', '--template',
```
alastair, line 116: delete this if you don't need it

models/sklearn/classification/report_files_export.py
tzamalisp: I just used the `"{}".format(name)` for safety, because someone in the future could add a number for `name`. However, I changed it to `str(name)`.

models/sklearn/utils.py
```diff
@@ -0,0 +1,166 @@
+import os
+from pprint import pprint
+
+def load_yaml(path_file):
+    """
+    Todo: add comments, docstring info, etc.
+    :return:
+    """
+    try:
+        import yaml
```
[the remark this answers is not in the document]
tzamalisp: I saw that gaia did it the same way too. I followed your instructions and I put it at the top.

models/sklearn/utils.py
```diff
@@ -0,0 +1,166 @@
...
+    def inspect_directory(self):
+        """
+
+        :return:
+        """
+        # find dynamically the current script directory
+        # path_app = os.path.join(os.path.abspath(os.getcwd()))
+        full_path = os.path.join(self.exports_path, self.directory)
+        # create path directories if not exist --> else return the path
+        if not os.path.exists(full_path):
```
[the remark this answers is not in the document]
tzamalisp: Done. Instead of `exists_ok`, the correct argument is `exist_ok`.

models/sklearn/classification/classification_task_manager.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+
+        :param yaml_file: The configuration file name
+        :param train_class: The class that will be trained
+        :param X: The already shuffled data that contain the features
+        :param y: The already shuffled data that contain the labels
+        """
+        self.config = config
+        self.train_class = train_class
+        self.X = X
```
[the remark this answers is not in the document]
tzamalisp: I could change X and y to features and labels accordingly. I preferred to name these variables like that because of the standard terminology by sklearn and the Machine Learning in general. What do you think, should I change it? Check an example here: https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html

models/sklearn/utils.py
```diff
@@ -0,0 +1,166 @@
...
+                                process_dict["C"] = [2 ** x for x in c]  # 2 ** c
+                                # gamma
+                                gamma = classifier["gamma"]
+                                process_dict["gamma"] = [2 ** x for x in gamma]  # 2 ** gamma
+                                # class weights
+                                balance_classes = classifier["balanceClasses"]
+                                process_dict["balanceClasses"] = [change_weights_values(i) for i in balance_classes]
+                                processes.append(process_dict)
+                                # n_fold
+                                process_dict["n_fold"] = nfold_number
```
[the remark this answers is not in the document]
tzamalisp: I forgot to change it after completing the method implementation. It is ok now. Thanks for mentioning that.

models/sklearn/classification/classifierGRID.py
tzamalisp: Here, the outcome of this function is the best model from all the best models that are exported in each preprocessing step (basic, normalize, etc.). Gaia has a similar output too. Furthermore, the best model from each preprocessing step is also extracted, as well as all the params combinations that are executed for that step.
* best model params and score: https://github.com/tzamalisp/acousticbrainz-server/blob/sklearn_integration/models/sklearn/classification/classifier_grid.py#L107-L114
* All the grid params combinations executed for each step: https://github.com/tzamalisp/acousticbrainz-server/blob/sklearn_integration/models/sklearn/classification/classifier_grid.py#L116-L120

models/sklearn/utils.py
```diff
@@ -0,0 +1,166 @@
...
+                                trainings_counted += 1
+            # increase evaluation counter by 1
+            evaluation_counter += 1
+
+        print("Trainings to be applied: {}".format(trainings_counted))
+
+        return processes
+
+
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
tzamalisp: Sorry, I forgot to delete that code part after completing the implementation. I could write a test, but some checks take place in the `ClassificationTaskManager` and the `load_yaml()` if the configurations are right or not.

models/sklearn/classification/evaluation.py
tzamalisp: Done. This method is splitted to sub-methods for each separate step that takes place there.

models/sklearn/classification/classification_task_manager.py
```diff
@@ -0,0 +1,136 @@
...
+        self.images_path = ""
+        self.reports_path = ""
+
+        self.logger = ""
+        self.setting_logger()
+        self.files_existence()
+        self.config_file_analysis()
+
+    def setting_logger(self):
+        self.logger = LoggerSetup(config=self.config,
```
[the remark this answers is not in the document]
tzamalisp: Done. Everything is now set to a single logger

models/sklearn/logging_tool.py
```diff
@@ -0,0 +1,107 @@
...
+        #  if handlers are already present and if so, clear them before adding new handlers. This is pretty convenient
+        #  when debugging and the code includes the logger initialization
+        if logger_object.hasHandlers():
+            logger_object.handlers.clear()
+
+        # Add handlers to the logger
+        logger_object.addHandler(c_handler)
+        logger_object.addHandler(f_handler)
+
+        if self.level is None:
```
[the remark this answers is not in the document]
tzamalisp: I implemented it that way as a convenience if someone wants to run the tool with different debugging level or to run the tool from the terminal by calling the script with its logging argument. However, I changed the logging levels that can be inserted, from integers (0, 1, 2, 3, 4) to strings (logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR, logging.CRITICAL)

PR #388, opened 2021-01-22
Prepare for Consul upgrade

amCap1712:
The rationale of the changes made here is explained in the PR for upgrading [ListenBrainz](https://github.com/metabrainz/listenbrainz-server/pull/1237) setup.


alastair:
@brainzbot add to whitelist

docker/dataset_eval/consul-template-dataset-eval.conf
```diff
@@ -0,0 +1,11 @@
+template {
+    source = "/code/consul_config.py.ctmpl"
+    destination = "/code/config.py"
+}
+exec {
+    command = "cd /code && exec chpst -uacousticbrainz:acousticbrainz python worker_manage.py dataset_evaluator"
```
alastair, line 6: this won't work, because consul-template doesn't run this in a subshell, it just exec's the command. We also need to make an update to handle this in Listenbrainz. My plan is that we should have a shellscript for running python commands, e.g. `start-ab-command.sh`

```
#!/bin/bash
cd /code
exec chpst -uacousticbrainz:acousticbrainz python "$@"
```
and set `command = "start-ab-command.sh python worker_manage.py dataset_evaluator"`

PR #392, opened 2021-02-06
Use a loop to iterate dataset tables during dumps

amCap1712:
The change is self-explanatory. Iterating over the keys of the `dict` does not guarantee iteration in any particular order. If we want to iterate in insertion order, we can use an `OrderedDict` instead of a simple `dict`.


alastair (changes requested):
I think the idea here was that the tables should be done in a specific order so that when importing the dump, the dataset table rows existed when importing the dataset_class table (for example).

However, I've just looked at the import code again and we do this import without FK constraints, so I think it's actually OK.

In any case, I think it makes sense to be explicit by using an OrderedDict when creating `_DATASET_TABLES`

PR #396, opened 2021-04-18
AB-407: Redirect legacy API endpoints to new endpoint with http redirect

r0han22:
Redirecting legacy GET API views to the /api/v1 views with HTTP 301.


alastair:
As we're planning on shutting down AB, I'm going to close this.

PR #399, opened 2021-05-10
Run tests in github actions

alastair:


alastair:
see also https://github.com/metabrainz/listenbrainz-server/pull/1452

alastair:
With the docker publish action and cache settings, this shouldn't cause long build times unless early layers of the image are invalidated

PR #404, opened 2021-06-29
Add missing consul config flags

amCap1712:
These flags were added in https://github.com/metabrainz/acousticbrainz-server/pull/257 but we forgot to add those to consul config.


alastair (approved):
404 PR not found!

